PR #18, opened 2020-04-21
Add ESLint Foundation

dhurlburtusa:
Added ESLint foundation with support for React and TypeScript.

Mostly using recommended configuration/rules.

This is just a foundation. We may need to make some adjustments.

Added an npm script to run ESLint. `npm run lint:js`. (Plan to have `lint:css` to handle CSS. Plan to have `lint` to run both.)

## How has this been tested?

Ran the script and made adjustments until there were no warnings or errors.


.gitignore
```diff
@@ -70,38 +42,11 @@ typings/
...
-# FuseBox cache
-.fusebox/
-
-# DynamoDB Local files
-.dynamodb/
-
-# TernJS port file
-.tern-port
-
-# IDEs
```
yanirs, line 106: Let's keep IDE-specific configs out of the repo. Some of us prefer WebStorm rather than VS Code. That is, add back `.idea/` and also `.vscode/` and remove the VS Code files.

package.json
```diff
@@ -1,30 +1,35 @@
 {
   "name": "abacus",
+  "version": "0.0.1",
```
yanirs, line 3: What's the reason for reordering the JSONs? I don't mind any order, as long as we keep it consistent moving forward so that diffs are easy to view.

yanirs:
@dhurlburtusa Can we please keep PRs focused and small? Specifically, the last three commits (https://github.com/Automattic/abacus/pull/18/commits/5156ad6699243aa1a3f9244f518a8e2d77103551 https://github.com/Automattic/abacus/pull/18/commits/1402f1c4111f443f49b90b3efae66a6fd1934484 https://github.com/Automattic/abacus/pull/18/commits/6d3c2d2a80484755a75f3dc74c362fd8eb6c0a7e) change the scope of the PR, so they'd be better as a separate PR (that can have this PR as the base branch).

package.json
dhurlburtusa: The name and version are usually paired/used together. It was a minor change so I went for it.

.gitignore
dhurlburtusa: Fixed.

package.json
yanirs: I was thinking more about the changes below, of moving the dependencies and reordering `tsconfig.json`. It's hard to see what changed this way.

.gitignore
yanirs: Thanks! :slightly_smiling_face:

package.json
dhurlburtusa: I moved the "scripts" in `package.json` to the bottom because it is the most common thing I look at in this file. Having it at the bottom makes it easier to find than trying to find it above the every growing list of dependencies.

For the `tsconfig.json`, to me it is easier to reason about what is being included and excluded when `include` comes before `exclude`.

.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+    'padded-blocks': ['off'],
+
+    // Off because we are using TypeScript which expects us to declare the props.
+    'react/prop-types': 'off',
+
+    '@typescript-eslint/explicit-function-return-type': 'off',
+
+    // The standard `no-unused-vars` rule is handling unused variables.
+    // Off because it incorrectly detects whether a variable is unused.
+    // '@typescript-eslint/no-unused-vars': 'off',
```
yanirs, line 47: Following [this doc](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md), this seems to work for me:

```suggestion
    // Disabling the base rule as it can report incorrect errors
    'no-unused-vars': 'off',
    '@typescript-eslint/no-unused-vars': ['error'],
```

I ran `npm run lint:js` and got no errors with this change.

package.json
yanirs: Sounds good. I don't really mind the order as long as we don't change it again in subsequent PRs. :slightly_smiling_face:

yanirs:
Oh, one more thing: Can you please add a description of the new command to the README before merging?

.eslintrc.js
dhurlburtusa: Nice find. I am really surprised that the base rule is not disabled by default since we are extending `plugin:@typescript-eslint/eslint-recommended` and `plugin:@typescript-eslint/recommended`.

@yanir I have started an ordering convention in `.eslintrc.js`. I don't think there is a rule we can use to help keep it in order or even if the team is in agreement. Basically, I have the ESLint rules grouped by whether it is a base, built-in rule, from `eslint-plugin-react` (the rule names begin with `react/`, or from `@typescript-eslint/eslint-plugin` (the rule names begin with `@typescript-eslint/`). Then within each group, they are alphabetized by rule name. This just makes it easier to find things. Once we pretty much finalize the rules, we won't be make many, if any, changes.

yanirs: Ordering sounds good! Can just move the `no-unused-vars` up with the comment.

I haven't dug too deep, but looks like they're working on a better solution: https://github.com/typescript-eslint/typescript-eslint/issues/1856

dhurlburtusa:
Fixed unused variable rule.
Added note about the `lint:js` command.

This is awesome! I am finally on a team that takes reviews seriously and not just another annoying chore that management wants (actually I've never been on a team where management cared out code reviews except for the client I freelance with to do their code reviews.)

yanirs:
> This is awesome! I am finally on a team that takes reviews seriously and not just another annoying chore that management wants (actually I've never been on a team where management cared out code reviews except for the client I freelance with to do their code reviews.)

Woohoo! I'm glad you're enjoying it. :slightly_smiling_face:

PR #20, opened 2020-04-22
Add Stylelint

dhurlburtusa:
Added StyleLint along with a couple commands to run it.

## How has this been tested?

Added a test CSS file to try out whether some basic issues are found. Tried in a couple different directories such as `node_modules` to ensure they did not need to be configured to be ignored.


stylelint-test.css
```diff
@@ -0,0 +1,7 @@
+/*
```
yanirs, line 1: Should we add minimal stylesheet instead under the path where we intend to add stylesheets? Looking at the feature store dashboard repo, @jsnmoon added:

* `/styles/app.scss`: Our own definitions
* `/styles/vendor.scss`: Imports of third-party definitions
* `/styles/main.scss`: Imports from `app.scss` and `vendor.scss`

For now, can just add `/styles/main.scss` as an empty file to keep StyleLint happy.

README.md
```diff
@@ -14,6 +14,22 @@ All the scripts for this project are initiated via npm scripts. Please see the `
...
+
+Runs a linter on all the styling code.
+
+```sh
+npm run lint:css
+```
+
+**lint:css:fix**
+
+Runs a linter on all the styling code and fixes and found issues that are fixable.
```
yanirs, line 27: `and found` seems redundant.

```suggestion
Runs a linter on all the styling code and fixes issues that are fixable. 
```

package.json
```diff
@@ -23,11 +23,26 @@
...
+    "lint-staged": "^10.1.7",
+    "stylelint": "^13.3.3",
+    "stylelint-config-standard": "^20.0.0",
     "typescript": "^3.8.3"
   },
+  "lint-staged": {
+    "linters": {
+      "*.{css,scss}": [
+        "stylelint",
+        "git add --force"
```
withinboredom, line 36: Would there be any objections to fixing on commit, then running the linter? Similar to prettier's fixing of staged data before commit? (I don't actually know if that's possible)

README.md
dhurlburtusa: Was suppose to be `any` instead of `and`. I'll get it fixed up.

stylelint-test.css
dhurlburtusa: Done.

README.md
dhurlburtusa: Fixed.

package.json
dhurlburtusa: I'd like to see how it goes. So, yes. Let's do it.

dhurlburtusa: Done.

PR #26, opened 2020-04-23
Fix Prettier setup

### LINKED ISSUE #25: Tweak Prettier Setup
Run prettier during pre-commit hook.
Fix some a few issues in files that prettier doesn't know how to handle without a special parser.
Fix a formatting issue that snuck in before pre-commit hook was in place.

dhurlburtusa:
Mainly fixes line-endings.

Developers should configure IDE to use LF for new files because Prettier does not run on all files.

Resolves #25 

Adds a new config: `jsxSingleQuotes: true`. Meant to add this initially but forgot. Are you guy's okay with single quotes for JSX?

A little over a year ago I resisted them but I ended up switching because I preferred single quotes everywhere.

## How has this been tested?

Ran `npm run format` to run prettier.
Manually checked files prettier doesn't natively handle.


components/Layout.test.tsx
```diff
@@ -0,0 +1,12 @@
+import { render } from '@testing-library/react'
```
yanirs, line 1: I'm guessing that this wasn't supposed to be included in this PR. Can you please remove it?

PR #45, opened 2020-04-29
Add puppeteer

### LINKED ISSUE #31: Add Test Framework for E2E Tests
A continuation of #17 but for E2E tests.

Possibly useful links:
https://extendomattic.wordpress.com/2019/02/20/woocommerce-extensions-e2e-tests-part-1-selecting-the-framework/

We've decided to go with Puppeteer instead of Cypress. See discussion in https://betterexperiments.wordpress.com/2020/04/23/end-to-end-testing-framework/.

dhurlburtusa:
Add puppeteer for our E2E testing.

Resolves #31.

## How has this been tested?

Ran `npm run build && npm run type-check && npm run format:check && npm run lint && npm run test:unit -- --coverage && npm run test:e2e` locally and everything passed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,10 +11,10 @@ jobs:
       tag: '10.18.1'
     steps:
       - checkout
-      - puppeteer/install
       - node/with-cache:
           steps:
             - run: npm install
+            - puppeteer/install
             - run:
                 name: Format Checking
                 command: npm run format:check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,8 +37,8 @@ jobs:
                 name: Unit Testing
                 command: npm run test:unit -- --coverage
             - run:
-                name: End-to-End Testing
-                command: npm run test:e2e
+                name: End-to-End Smoke Tests
+                command: npm run test:e2e:smoke
 
 workflows:
   build-and-verify:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  node: circleci/node@2.0.3
+  node: circleci/node@1.1.6
   puppeteer: threetreeslight/puppeteer@0.1.2
 
 jobs:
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -20,7 +35,10 @@ jobs:
                 command: npm run lint
             - run:
                 name: Unit Testing
-                command: npm run test:unit -- --coverage
+                command: npm run test:unit -- --ci --coverage
+            - run:
+                name: End-to-End Smoke Tests
```
yanirs, line 40: What's the rationale behind running only the smoke tests on CI?

.circleci/config.yml
```diff
@@ -7,11 +7,26 @@ jobs:
...
           steps:
-            - run: npm install
+            - run:
+                name: Install Package Dependencies
+                command: npm install
+            # Copied from https://circleci.com/orbs/registry/orb/threetreeslight/puppeteer.
+            # Not using this Orb directly to prevent re-downloading Chromium which was already
+            # done in the previous step.
+            - run:
+                name: Install Headless Chrome Dependencies
```
yanirs, line 22: Not for this PR, but I think we will need to add caching of apt packages to speed up the build.

yanirs:
Can you please add a note to the readme?

My main question is about running only smoke tests on CircleCI. It doesn't matter right now given that we don't have non-smoke tests, but ideally we should run the full test suite (unless it's impossible for some reason).

.circleci/config.yml
dhurlburtusa: @yanirs I wasn't sure if we wanted to run the full e2e test suite. I'll change to run all e2e tests.

.circleci/config.yml
dhurlburtusa: @yanirs Would you please clarify what you mean by "Not for this PR"?
I'll investigate how the apt packages can be cached.

yanirs: Sorry, I meant that it's not a blocker for this PR or too important at this point, just something to think about / address in the future.

dhurlburtusa: Added #48.

dhurlburtusa:
@yanirs Okay, the CI build runs the whole E2E test suite. Do we still need something added to the readme? If not, I think this PR is ready.

yanirs:
> @yanirs Okay, the CI build runs the whole E2E test suite. Do we still need something added to the readme? If not, I think this PR is ready.

Thanks! Yeah, please add a note under the `Testing` section before merging, along the lines of `test:unit` (which should probably be updated to exclude the reference to E2E testing).

PR #50, opened 2020-05-01
Fix Pre-Commit Hook

dhurlburtusa:
It didn't seem like the pre-commit hook was actually running although I was using the same configuration on other projects I use it with.

I found out that we can see the calls to the lint-staged commands by looking at the Git console in VS Code.

## How has this been tested?

After committing, I can see in the Git console that the calls to the lint-staged commands were being made.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,11 @@
     "ts-jest": "^25.4.0",
     "typescript": "^3.8.3"
   },
+  "husky": {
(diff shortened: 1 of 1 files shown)


yanirs:
I was going to ask you about the pre-commit hook not working but you beat me to it. :slightly_smiling_face: Thanks for fixing it!

PR #58, opened 2020-05-05
Add an App-Level Error Boundary

dhurlburtusa:
Add an app-level error boundary to help ensure the app doesn't unexpectedly fail without an error for the end user.

Prerequisite for #7.

## How has this been tested?

Added unit tests to test new RenderErrorBoundary component.


components/RenderErrorBoundary.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    fireEvent.click(getByText('Clear'))
+
+    expect(onClear).toHaveBeenCalledTimes(1)
+
+    expect(onError).toHaveBeenCalledTimes(2)
+    expect(onError).toHaveBeenNthCalledWith(2, { clear, error, info })
+
+    expect(mockRenderProp).toHaveBeenCalledTimes(4)
+    expect(mockRenderProp).toHaveBeenNthCalledWith(3, { renderError: null })
+    expect(mockRenderProp).toHaveBeenNthCalledWith(4, { renderError: { clear, error, info } })
```
withinboredom, line 127: It's a bit difficult to reason how you got these expects and it might be worth adding a comment above each of them to explain what it means when they fail.

```suggestion
    // called after attempt to render bad component, now with an error.
    expect(mockRenderProp).toHaveBeenNthCalledWith(4, { renderError: { clear, error, info } })
```

This way someone debugging or extending the error boundary doesn't need to dissect the entire component to understand why it failed. If this specific case failed, they can now realize their changes made it so that the error wasn't handled correctly.

components/RenderErrorBoundary.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    expect(container).toHaveTextContent('Oh no! Not again!')
+
+    fireEvent.click(getByText('Clear'))
+
+    expect(onClear).toHaveBeenCalledTimes(1)
+
+    expect(onError).toHaveBeenCalledTimes(2)
+    expect(onError).toHaveBeenNthCalledWith(2, { clear, error, info })
+
+    expect(mockRenderProp).toHaveBeenCalledTimes(4)
```
withinboredom, line 125: I'd personally remove these checks. Later assertions will fail naturally and include more information, or at least move it until after the next couple of lines assert that it wasn't called more than expected.

components/RenderErrorBoundary.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react'
+
+interface RenderError {
+  clear: () => void
+  error: Error
+  info: React.ErrorInfo
+}
+
+type ChildrenParam0 = {
```
withinboredom, line 9: Can we just use

```suggestion
type ChildrenParam = {
```

pages/_app.tsx
```diff
@@ -29,9 +33,22 @@ const App = React.memo(function App(props: AppProps) {
...
   return (
-    <div className='app'>
-      <Route {...routeProps} />
-    </div>
+    <RenderErrorBoundary onError={onAppRenderError}>
+      {({ renderError }) => (
+        <>
+          {renderError ? (
+            <>
+              <p>An error occurred. If error persists, please contact the Experiments Platform team.</p>
```
withinboredom, line 41: Not relevant to this PR, but @yanirs, what do you think about getting "experiment-platform@automattic.com":

```suggestion
              <p>An error occurred. If error persists, please contact the <a href='mailto:experiment-platform@automattic.com'>Experiments Platform team</a>.</p>
```

Or we could create a public p2/forum (eventually)...

components/RenderErrorBoundary.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+  expect(mockRenderProp).toHaveBeenCalledWith({ renderError: null })
+  expect(container).toHaveTextContent('A child.')
+})
+
+test('RenderErrorBoundary called without `onClear` or `onError` passes render prop an argument object with a `renderError` property with an error, info, and a clear function when render prop errors', () => {
+  try {
+    // Temporarily turn off the error console.
+    jest.spyOn(console, 'error')
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    ;(console.error as jest.Mock).mockImplementation(() => {})
```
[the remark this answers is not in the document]
dhurlburtusa: I do and don't like having semi-colons off. I like because it's a tab bit easier on the eyes. I don't because of things like this. I'm good either way. The plus is we have Prettier in place to quickly fix if we want.

components/RenderErrorBoundary.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+import RenderErrorBoundary from './RenderErrorBoundary'
+
+const obj = {}
+
+function BadComponent() {
+  // Need a component that will fail when React attempts to render but also need
+  // something that TypeScript will compile. The type assertion allows TS to
+  // compile successfully but because we essentially lied to TS, it will fail as
+  // runtime.
+  return <>{(obj as { nonExistent: { boom: 'shaka laka' } }).nonExistent.boom}</>
```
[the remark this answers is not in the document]
dhurlburtusa: Thanks.
I didn't explore whether where is a more simple way to achieve the same effect. We can simplify if we think it is worth investing the time.

pages/_app.tsx
yanirs: Can leave the email out for now. :smile: 

If we want a link, let's link to the P2: https://betterexperiments.wordpress.com/.

Can add make it a public P2 if and when we go public. :slightly_smiling_face:

components/RenderErrorBoundary.test.tsx
dhurlburtusa: Added some comments.

components/RenderErrorBoundary.tsx
dhurlburtusa: The 0 is meant to align with the 0 in arg0.

PR #59, opened 2020-05-06
Add an Error Boundary for Pages

dhurlburtusa:
Add an error boundary around pages to more easily recover from errors on a page by page basis instead of the whole app.

For now, simply added the error boundary to the Layout component as the expectation is that this layout will be the top-level component of all pages.

Prerequisite for #7.

## How has this been tested?

Added more components and unit tests.


components/Layout.test.tsx
```diff
@@ -1,16 +1,34 @@
...
   const { container } = render(<Layout title='Some Title'>A child.</Layout>)
 
-  const divElmt = container.firstChild
-  if (divElmt) {
-    expect(divElmt.nodeName).toBe('DIV')
-  } else {
-    expect(divElmt).not.toBe(null)
-  }
+  expect(container).toHaveTextContent('A child.')
   // TODO: Make expectations more thorough.
```
withinboredom, line 11: I think this is better than before and this todo can be removed. A better test might be a jest snapshot.

components/BadComponent.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+const obj = {}
+
+export default function BadComponent() {
```
yanirs, line 5: Maybe we should have a file or directory with test helpers? It's a bit odd to have test-only components in the main `components/` directory.

dhurlburtusa: Moved to `helpers` directory.

PR #62, opened 2020-05-06
Add Initial API Parts

### LINKED ISSUE #3: Generate API client from Swagger doc
This looks useful: https://github.com/api-platform/client-generator

dhurlburtusa:
Add initial API parts.

Resolves #3.

## How has this been tested?

The plan is to add e2e tests once the list component is added. That should be the next PR. For now, just trying to keep the PR to a reasonable size.


api/ExperimentsApi.ts
```diff
@@ -0,0 +1,50 @@
+import { Experiment } from '@/models/index'
+import { getExperimentsAuthInfo } from '@/utils/auth'
+
+import UnauthorizedError from './UnauthorizedError'
+
+const DEVELOPMENT_API_URL_ROOT = 'https://virtserver.swaggerhub.com/yanir/experiments/0.1.0'
+const PRODUCTION_API_URL_ROOT = 'https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0'
+
+function resolveApiUrlRoot(host: string) {
```
dhurlburtusa, line 9: This and it associated constants can be moved to a separate module once another API module is added.

api/ExperimentsApi.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Finds all the available experiments.
+ *
+ * Note: Be sure to handle any errors that may be thrown.
+ *
+ * @throws UnauthorizedError
+ */
+async function findAll(): Promise<Experiment[]> {
+  const apiUrlRoot = resolveApiUrlRoot(window.location.host)
```
dhurlburtusa, line 22: @aaronyan @withinboredom @yanirs 
TODO: Discuss the following use of `window.location.host`. I am not a fan of mixing browser API code here. Consider injecting in API URL root with Next.js build. Next.js let's us know when it is building for production or otherwise. This info can be used to chose the URL root which we can access via env vars. For example, `process.env.EXPERIMENTS_REST_API_URL_ROOT`.

api/ExperimentsApi.ts
```diff
@@ -0,0 +1,50 @@
...
+  return fetch(fetchUrl, {
+    method: 'GET',
+    headers,
+  })
+    .then((response) => response.json())
+    .then((result) => {
+      const { experiments } = result
+
+      return experiments
+    })
```
yanirs, line 43: This can be simplified:

```suggestion
    .then((result) => result.experiments)
```

api/ExperimentsApi.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Note: Be sure to handle any errors that may be thrown.
+ *
+ * @throws UnauthorizedError
+ */
+async function findAll(): Promise<Experiment[]> {
+  const apiUrlRoot = resolveApiUrlRoot(window.location.host)
+
+  const headers: HeadersInit = new Headers()
+  const experimentsApiAuthInfo = getExperimentsAuthInfo()
+  if (experimentsApiAuthInfo) {
```
yanirs, line 26: We may end up making requests without a token. In addition, this code can be simplified to:

```typescript
  const accessToken = getExperimentsAuthInfo()?.accessToken
  if (!accessToken) {
      throw new UnauthorizedError()
  }
  const headers = new Headers({'Authorization': `Bearer ${accessToken}`})
```

api/ExperimentsApi.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ * Finds all the available experiments.
+ *
+ * Note: Be sure to handle any errors that may be thrown.
+ *
+ * @throws UnauthorizedError
+ */
+async function findAll(): Promise<Experiment[]> {
```
yanirs, line 21: Shall we keep the names aligned with http verbs? That is, `getAll()` or `getList()` rather than `findAll()`?

models/Experiment.ts
```diff
@@ -0,0 +1,96 @@
+import { Event, MetricAssignment, Platform, SegmentAssignment, Status, Variation } from './index'
+
+export interface Experiment {
+  /**
+   * Unique experiment ID.
+   */
+  readonly experimentId: number
```
yanirs, line 7: Was the conversion to camelCase done by the generator?

api/ExperimentsApi.ts
yanirs: Also, what about error handling?

api/ExperimentsApi.ts
dhurlburtusa: @yanirs I see your point about passing anything to `resolveApiUrlRoot`.

To clarify, the suggestion is to add build time env vars for the API URL root. There would be no hardcoding of the URL and there would be no need for `resolveApiUrlRoot`. The choice of the API URL root is done at build time instead of runtime. This will also keep `https://virtserver.swaggerhub.com/yanir/experiments/0.1.0 ` out of the production code (although there may be other ways to do that if a different solution is used).

api/ExperimentsApi.ts
dhurlburtusa: @yanirs In many libraries I've worked with that asynchronously fetch data, it's been more common to use `find*`. It implies that it is asynchronous. To me, `get` implies that everything is already available in memory and it just needs to be returned, maybe with some calculation or transformation done before returning.

Originally, in the trial code, we were passing in criteria to filter by status. The criteria was removed but I left the name the same.

The intention of `ExperimentsApi` is to act as a data layer. The client code doesn't need to know how the data is being fetched. For all it knows, it can be coming from GraphQL, REST, a local JSON file, etc.

Please let me know if you still think it would be better to rename with `get*`.

api/ExperimentsApi.ts
```diff
@@ -0,0 +1,50 @@
...
+  const apiUrlRoot = resolveApiUrlRoot(window.location.host)
+
+  const headers: HeadersInit = new Headers()
+  const experimentsApiAuthInfo = getExperimentsAuthInfo()
+  if (experimentsApiAuthInfo) {
+    if (!experimentsApiAuthInfo.accessToken) {
+      throw new UnauthorizedError()
+    }
+    headers.append('Authorization', `Bearer ${experimentsApiAuthInfo.accessToken}`)
```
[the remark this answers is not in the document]
dhurlburtusa: The code now only sends if the API URL root is the production version.

api/ExperimentsApi.ts
dhurlburtusa: @yanirs Since most errors should be presented to the user, I was planning to handle it when `findAll` was called.

models/Experiment.ts
dhurlburtusa: @yanirs Yes. I think this is the way to go since at the point the data is in memory and the JavaScript context/environment (that is, it is no longer in the data transport context). The API will need to transform snake_case to camelCase.

yanirs: Sounds good. Do you know of a library that does that for us?

api/ExperimentsApi.ts
yanirs: I'm happy with `findAll()` :slightly_smiling_face:

api/ExperimentsApi.ts
yanirs: Hmm... Not sure at this stage, as it depends on the actual error. Let's see how it goes when we have more real examples to work with. :slightly_smiling_face:

models/Experiment.ts
dhurlburtusa: Not off the top of my head put I'll search for something.

api/ExperimentsApi.ts
yanirs: I think I need to see what it looks like, including the solution to not leaking the production auth token to the mock server. I agree it'd be nice to have these settings as configurable constants rather than hardcoded strings with extra functions. Maybe explore this direction in a separate PR after this PR is merged?

api/ExperimentsApi.ts
dhurlburtusa: After addressing the feedback about sending the Authorization header only for production API, I ended up referencing the constant more than once.

Although when I wrote this function I was aware that the constants were only being used once, I find the individual `const`s much more self documenting.

pages/index.tsx
```diff
@@ -1,18 +1,40 @@
...
 
+import { Experiment } from '@/models/index'
+
 const debug = debugFactory('abacus:pages/index.tsx')
 
 const IndexPage = function IndexPage() {
   debug('IndexPage#render')
+  const [experiments, setExperiments] = useState<Experiment[] | null>(null)
+
+  useEffect(() => {
```
dhurlburtusa, line 16: Note: Doing the fetch here is just for testing that the call actually works. It will likely be moved else where depending on the decision from a P2 I am about to write regarding fetching, storing, caching, paginating, updating the experiments data.

components/Layout.test.tsx
```diff
@@ -71,3 +71,25 @@ test('renders RenderErrorView when has bad children', () => {
     ;(console.error as jest.Mock).mockRestore()
   }
 })
+
+test('renders RenderErrorView when has bad children', () => {
```
yanirs, line 75: Looks like this was added twice by mistake.

pages/index.tsx
yanirs: Sounds good. This file will undergo a rewrite anyway. :slightly_smiling_face:

components/Layout.test.tsx
dhurlburtusa: I think it is gone from this PR. I don't think I was correctly merging in the various branches.

@yanirs How do you keep track of all these branches and which ones need to be merged, which ones haven't been merged yet? Do you a visual tool to figure all this out?

I've been using VCS for over 15 years now. I started with CVS, something from Microsoft (I forget its name) when I worked with my first programming/development employer, we then moved to Subversion and later Mercurial. I started using Git about 6 years ago. Because I had used so many other VCSs, I guess I never took the time to learn more than the basics of using Git. I might need to spend some more time learning the different branching strategies available. I've read most of https://www.atlassian.com/git/tutorials. I think I might have to re-read https://medium.com/@kurtisnusbaum/large-diffs-are-hurting-your-ability-to-ship-e0b2b41e8acf and https://unhashable.com/stacked-pull-requests-keeping-github-diffs-small/.

When you were learning Git, was there any documentation/tutorials that stood out for you?

dhurlburtusa:
I thing the unnecessary commit was due to not being fully sync'd with `master` and the PR 59 branch or something. I think it is resolved now. I'll add an issue about testing the ExperimentApi. My original plan was to test it via e2e or integration tests.

components/Layout.test.tsx
yanirs: > When you were learning Git, was there any documentation/tutorials that stood out for you?

image

But seriously, I don't know Git very well, just enough to get by. :slightly_smiling_face: 

> How do you keep track of all these branches and which ones need to be merged, which ones haven't been merged yet? Do you a visual tool to figure all this out?

I use the command line tool for most work (which is the 5-10 commands alluded to by the parody cover), and for comparisons between branches and files I use `git difftool`. In my case I configured it to use [meld](https://meldmerge.org/), but I see [you can use VS Code](https://code.visualstudio.com/docs/editor/versioncontrol#_vs-code-as-git-diff-tool). Before pushing, I usually compare to the base branch, e.g.:

    $ git checkout master
    $ git pull origin master
    $ git checkout add/api-parts
    $ git difftool master

I also run such visual comparisons after merging, e.g.:

    $ git checkout add/api-parts
    $ git merge master
    ... resolve conflicts
    $ git log
    ... get the commit hash for the pre-merge version, e.g., abc123
    $ git difftool abc123

I'm sure there are better ways to leverage advanced Git commands, but these flows work for me. :slightly_smiling_face:

PR #63, opened 2020-05-07
Skip auth in non-production environments

### LINKED ISSUE #55: Re-enable e2e tests on CircleCI
With #44, we disabled e2e tests on CircleCI as it's still unclear how to run them with the authentication redirect to WP.com. Let's figure this out and re-enable them.

One possible solution is to store a fake auth token so that the redirect to WP.com doesn't happen. We don't need a real token since we should run the e2e tests with the mock API. Thoughts? @dhurlburtusa @withinboredom

### LINKED ISSUE #56: Figure out how to securely make Vercel work with our authorisation solution
#44 breaks our Vercel deployment due to the addition of WP.com OAuth with specific allowed hosts. Let's discuss internally and figure out how we can support feature branch deployment and authentication.

yanirs:
Fix #55 and resolve #56 by injecting a fake auth token in non-production environments. This also includes re-enabling e2e tests on CircleCI.

## How has this been tested?

* Locally: `npm run verify`
* Manual check that there is no redirect to authenticate when running locally
* Manual check that the Vercel deployment works
* Manual check that the CircleCI build passes

### DIFF
--- a/pages/auth.tsx
+++ b/pages/auth.tsx
@@ -37,7 +37,7 @@ const AuthPage = function AuthPage() {
           }
           saveExperimentsAuthInfo(experimentsAuthInfo)
 
-          window.location.assign(window.location.origin)
+          window.location.replace(window.location.origin)
         } else if (authInfo.error === 'access_denied') {
           setError('Please log into WordPress.com and authorize Abacus - Testing to have access.')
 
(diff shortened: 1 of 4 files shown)


pages/auth.tsx
```diff
@@ -37,7 +37,7 @@ const AuthPage = function AuthPage() {
           }
           saveExperimentsAuthInfo(experimentsAuthInfo)
 
-          window.location.assign(window.location.origin)
+          window.location.replace(window.location.origin)
```
yanirs, line 40: Not entirely related to this PR, but I think it's cleaner to not have the URL with the token in the browser's history.

dhurlburtusa: I agree. I missed changing this from when it was a popup.

PR #66, opened 2020-05-08
Add Experiments List

### LINKED ISSUE #7: Add experiment list page
This should probably be the default homepage

dhurlburtusa:
Add a list of experiments on home page.

- [x ] Add ExperimentsTable component.
- [x] Decide where to load data.
  + [x] Each time home page is loaded? (Store in page component's state.)
  + [x] Initially when home page is loaded, then periodic updates while page is loaded or next time home page is loaded if a certain amount of time transpires. (Store in app's state.)
- [x] Decide whether page of table should be remember across page navigations.

Resolves #7.

## How has this been tested?

Stories and unit tests have been added for all the new components. We still have 100% coverage.


tsconfig.json
```diff
@@ -11,8 +11,6 @@
     "moduleResolution": "node",
     "noEmit": true,
     "noFallthroughCasesInSwitch": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
```
dhurlburtusa, line 15: Our linting rules still detect unused locals and parameters. Removing these just prevents the TypeScript compiler from barfing while in development. Basically, this is for better DX.

components/ExperimentsTable.test.tsx
```diff
@@ -0,0 +1,129 @@
+import { fireEvent, getAllByText, getByText, render } from '@testing-library/react'
```
dhurlburtusa, line 1: I'm still pretty new to using `@testing-library/react` plus TypeScript. Still trying to get a good feel for how to best write unit tests.

tsconfig.json
yanirs: These are important rules because they catch bugs. I can see how it'd be a pain in development, though. How about [we set up more lenient rules in development](https://stackoverflow.com/questions/46927323/environment-specific-tsconfig-json)?

components/ExperimentsTable.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+import ExperimentsTable from './ExperimentsTable'
+
+export default { title: 'ExperimentsTable' }
+
+const experiments: Experiment[] = []
+
+export const withNoExperiments = () => <ExperimentsTable experiments={experiments} />
+
+const endDatetime = new Date(Date.now() + 14 * 24 * 60 * 60 * 1000)
```
yanirs, line 17: If we're adding `date-fns`, let's use its helpers for date arithmetic (here and elsewhere).

components/ExperimentsTable.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { Platform } from '@/models/Platform'
+import { Status } from '@/models/Status'
+
+import ExperimentsTable, { PER_PAGE_DEFAULT } from './ExperimentsTable'
+
+test('with no experiments, renders an empty table', () => {
+  const experiments: Experiment[] = []
+  const { getByText } = render(<ExperimentsTable experiments={experiments} />)
+
+  expect(getByText('ID')).toBeInTheDocument()
```
yanirs, line 14: This doesn't really test that the table is empty. Given that the other tests don't check the headings, maybe just check that `tbody` doesn't have children?

yanirs:
Looks good! I think this will be good to go with some lightweight documentation of the new components and the addition of the actual `ExperimentsTable` to the homepage.

> Decide where to load data: Each time home page is loaded? (Store in page component's state.)
> Initially when home page is loaded, then periodic updates while page is loaded or next time home page is loaded if a certain amount of time transpires. (Store in app's state.)

Let's go with each time the home page is loaded for simplicity.

> Decide whether page of table should be remember across page navigations.

Do you mean that we'd add the page number to the URL, i.e., that changing pages affects history? This makes sense, though I wouldn't mind deferring it to a separate issue.

tsconfig.json
dhurlburtusa: @yanirs I agree that they are important rules. The TSLint rules catch the same unused locals and unused parameters. Do you think the TSLint rules are sufficient or should we invest some time to create environment specific tsconfigs?

yanirs: Ah, does this TSLint fail on unused locals and parameters? If so, I agree we don't need them as compilation blockers in `tsconfig.json`.

dhurlburtusa: Just to double check, I temporarily added some unused params and locals, then I ran TSLint. It fails with errors. So, as long as the `@typescript-eslint/no-unused-vars` rule is not loosened to the point where it doesn't error anymore, we should be good.

components/ExperimentsTable.test.tsx
dhurlburtusa: We check for an empty `tbody` now.

components/ExperimentsTable.stories.tsx
dhurlburtusa: Done.

dhurlburtusa:
@yanirs 
> Do you mean that we'd add the page number to the URL, i.e., that changing pages affects history? This makes sense, though I wouldn't mind deferring it to a separate issue.
It could be added to the URL too to keep track of which page of the table to display.
Okay. let's defer to a separate issue. For now, the table still handles pagination when there are more than 25 but which page to display is handled in `ExperimentTable`'s state.

components/Layout.test.tsx
```diff
@@ -20,21 +20,25 @@ test('renders layout with declared title and children', () => {
...
-          |
-        </span>
-        <a
-          href="/metrics"
-        >
-          Metrics
-        </a>
-      </nav>
+      <div
+        class="ui container"
```
yanirs, line 24: Not a major concern, but all these new lines make it a bit hard to read. Is there something we can easily do to make it more readable?

components/ExperimentsTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const endIdx = beginIdx + perPage
+  const experimentsToRender = experiments.slice(beginIdx, endIdx)
+  return (
+    <Table>
+      <Table.Header>
+        <Table.Row>
+          <Table.HeaderCell>ID</Table.HeaderCell>
+          <Table.HeaderCell>Name</Table.HeaderCell>
+          <Table.HeaderCell>Started</Table.HeaderCell>
+          <Table.HeaderCell>Ends</Table.HeaderCell>
```
yanirs, line 55: These can be in the past or future. Let's stick to start/end.

```suggestion
          <Table.HeaderCell>Start</Table.HeaderCell>
          <Table.HeaderCell>End</Table.HeaderCell>
```

components/ExperimentsTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+          <Table.HeaderCell>Status</Table.HeaderCell>
+          <Table.HeaderCell>Platform</Table.HeaderCell>
+          <Table.HeaderCell>Owner</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+      <Table.Body>
+        {experimentsToRender.map((e) => (
```
yanirs, line 62: `e` is a bit too ambiguous. Change to `experiment`?

components/ExperimentsTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+interface Props {
+  experiments: Experiment[]
+  // The zero-based page of data to display. Defaults to 0.
+  initialPage?: number
+  // The number of rows/items to display per page. Defaults to PER_PAGE_DEFAULT.
+  perPage?: number
+}
+
+const ExperimentsTable = (props: Props) => {
```
yanirs, line 23: Can you please document this?

components/ErrorsBox.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+interface Props {
+  errors: Error[]
+}
+
+const ErrorsBox = (props: Props) => {
```
yanirs, line 7: Can you please document this?

api/ExperimentsApi.ts
```diff
@@ -1,15 +1,37 @@
...
+}
+
 const DEVELOPMENT_API_URL_ROOT = 'https://virtserver.swaggerhub.com/yanir/experiments/0.1.0'
 const PRODUCTION_API_URL_ROOT = 'https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0'
 
 function resolveApiUrlRoot() {
   return window.location.host === 'experiments.a8c.com' ? PRODUCTION_API_URL_ROOT : DEVELOPMENT_API_URL_ROOT
 }
 
+function resolvePlatform(input: string): Platform {
```
yanirs, line 23: This isn't much of a resolution. :slightly_smiling_face: It'd also be more appropriate to add `Platform.fromString()` rather than add it here.

api/ExperimentsApi.ts
```diff
@@ -32,7 +54,31 @@ async function findAll(): Promise<Experiment[]> {
       apiUrlRoot === PRODUCTION_API_URL_ROOT ? new Headers({ Authorization: `Bearer ${accessToken}` }) : undefined,
   })
     .then((response) => response.json())
-    .then((result) => result.experiments)
+    .then((result) => {
+      const rawExperiments: RawExperiment[] = result.experiments
```
yanirs, line 58: What do we need `RawExperiment` for? Can't we just skip this in favour of `result.experiments.map( ... )` below.

api/ExperimentsApi.ts
yanirs: Also, this _does_ need a test.

api/ExperimentsApi.ts
```diff
@@ -32,7 +54,31 @@ async function findAll(): Promise<Experiment[]> {
...
+
+      // TODO: Centralize this transformation logic.
+      // TODO: Try to make transformation less manual. For example, schema-driven.
+      const experiments: Experiment[] = rawExperiments.map((re) => {
```
yanirs, line 62: As discussed on Slack, let's split `Experiment` to `ExperimentBare` and `ExperimentFull` so it roughly matches the API's behaviour. Then there would be no need for dummy values.

api/ExperimentsApi.ts
```diff
@@ -32,7 +54,31 @@ async function findAll(): Promise<Experiment[]> {
...
+        const e = {
```
yanirs, line 63: Please let's not:
* Add unnecessary variables
* Give variables names like `e` or `re`

api/ExperimentsApi.ts
```diff
@@ -32,7 +54,31 @@ async function findAll(): Promise<Experiment[]> {
...
+          platform: resolvePlatform(re.platform),
+          ownerLogin: re.owner_login,
+          segmentAssignments: [], // TODO: Get from API. Currently not available, but required by Experiment interface.
+          startDatetime: parseISO(re.start_datetime),
+          status: Status.Staging, // TODO: Get from API. Currently not available, but required by Experiment interface.
+          variations: [], // TODO: Get from API. Currently not available, but required by Experiment interface.
+        }
+        return e
+      })
+      return experiments
```
yanirs, line 80: Again, there's no need to have `experiments` as a variable.

api/ExperimentsApi.ts
```diff
@@ -32,7 +54,31 @@ async function findAll(): Promise<Experiment[]> {
       apiUrlRoot === PRODUCTION_API_URL_ROOT ? new Headers({ Authorization: `Bearer ${accessToken}` }) : undefined,
   })
     .then((response) => response.json())
-    .then((result) => result.experiments)
+    .then((result) => {
+      const rawExperiments: RawExperiment[] = result.experiments
+
+      // TODO: Centralize this transformation logic.
```
yanirs, line 60: Let's add a tested `fromJson()` method to the models.

api/ExperimentsApi.ts
```diff
@@ -32,7 +54,31 @@ async function findAll(): Promise<Experiment[]> {
...
+          experimentId: re.experiment_id,
+          existingUsersAllowed: true, // TODO: Get from API. Currently not available, but required by Experiment interface.
+          metricAssignments: [], // TODO: Get from API. Currently not available, but required by Experiment interface.
+          name: re.name,
+          p2Url: 'https://betterexperiments.a8c.com/2020/05/13/example', // TODO: Get from API. Currently not available, but required by Experiment interface.
+          platform: resolvePlatform(re.platform),
+          ownerLogin: re.owner_login,
+          segmentAssignments: [], // TODO: Get from API. Currently not available, but required by Experiment interface.
+          startDatetime: parseISO(re.start_datetime),
+          status: Status.Staging, // TODO: Get from API. Currently not available, but required by Experiment interface.
```
yanirs, line 75: The mock returns interesting statuses now. :slightly_smiling_face:

components/Layout.test.tsx
dhurlburtusa: @yanir That's just the way Jest likes to do snapshots. They said they do that to make diffs easier to read. I can't find where I read that. I did a quick search to see if Jest allows choosing different snapshot formatting. I didn't come up with anything.

api/ExperimentsApi.ts
dhurlburtusa: @yanirs I agree. I read up a more on TypeScript enums. They are not are useful/powerful and the "enums" I wrote for the trial. I might bring in what I wrote from the trial instead of using TS enums.
I am going to first experiment a little more with the TS enums.

api/ExperimentsApi.ts
dhurlburtusa: @yanirs TS was complaining because it didn't have enough type information. So, it looks like we still need `RawExperiment` but we can eliminate the `rawExperiments` variable.

api/ExperimentsApi.ts
dhurlburtusa: 👍

api/ExperimentsApi.ts
dhurlburtusa: @yanirs

What should the input shape match? The raw API data?

At the moment, the "models" are just TypeScript interfaces. Are you thinking we should change them to classes with a static `fromJson`?

Or we could do something like the following using a module level function:
```
function fromJson(rawExperimentBare: RawExperimentBare): ExperimentBare {
  return { /* insert transformation here */ }
}
```
Thoughts?

Instead of saying "raw", maybe be more specific. For example, `ApiExperimentBare`, `RestApiExperimentBare`, `RestExperimentBare`. Other?

api/ExperimentsApi.ts
yanirs: From local testing, something like this would work:

```typescript
interface ApiJson {
  [key: string]: any
}
```

```typescript
      return result.experiments.map((re: ApiJson) => {
        return {
          experimentId: re.experiment_id,
          name: re.name,
          startDatetime: parseISO(re.start_datetime),
          endDatetime: parseISO(re.end_datetime),
          status: re.status,
          platform: resolvePlatform(re.platform),
          ownerLogin: re.owner_login,
        }
      })
```

It'd still need the type conversion to `ExperimentBare`.

api/ExperimentsApi.ts
yanirs: See https://github.com/Automattic/abacus/pull/66#discussion_r425468053. I think we can avoid raw types with a generic JSON type:

```typescript
interface ApiJson {
  [key: string]: any
}
```

> At the moment, the "models" are just TypeScript interfaces. Are you thinking we should change them to classes with a static `fromJson`?

Yeah, I think that classes would be more useful, as we will probably want to add other methods.

api/ExperimentsApi.ts
yanirs: Hmm... Might be better to stick to TS enums. See https://stackoverflow.com/a/41548441 for conversion from string and https://stackoverflow.com/a/28151986 for a way of adding enum "methods". Note that this is just from a cursory look and quick search, so there may be issues I'm not seeing.

api/ExperimentsApi.ts
dhurlburtusa: @yanirs Are you okay with naming it something that doesn't use JSON in the name when it is not JSON. JSON is a string data format. That is, a sequence of characters. That is, once the JSON is parsed, it is a JavaScript object, array, or primitive.

How about `ApiData`?

```ts
interface ApiData {
  [key: string]: any
}
```

yanirs: Makes sense. :slightly_smiling_face:

jest-puppeteer.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   server: {
     command: 'echo "Building app..." && npm run build && echo "Starting app..." && npm run start',
     debug: true, // Allows us to see the output of the above commands.
-    launchTimeout: 30 * 1000,
+    launchTimeout: 60 * 1000,
```
dhurlburtusa, line 11: I kept getting timeouts waiting for the server to launch. So, gave it some more time.

models/Platform.ts
```diff
@@ -6,7 +8,13 @@
...
  *   usually written in PHP. Email and landing pages experiments are likely run here.
  */
-export enum Platform {
+enum Platform {
   Calypso = 'calypso',
   Wpcom = 'wpcom',
 }
+
+function toPlatform(input: string) {
+  return Platform[capitalize(input) as keyof typeof Platform] || Platform.Calypso
```
yanirs, line 17: Defaulting to `Calypso` here and to `Staging` in `toStatus()` may mask bugs. Let's throw exceptions for bad strings.

models/ExperimentBare.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  platform: Platform
+
+  /**
+   * The login name of the experiment owner.
+   */
+  ownerLogin: string
+}
+
+export function fromApiDataToBare(apiData: ApiData) {
```
yanirs, line 43: I played around with it a bit locally and I think it'll be clearer and more self-documenting to turn `ExperimentBare` into a class with the following constructor:

```typescript
  constructor(apiData: ApiData) {
    this.experimentId = apiData.experiment_id;
    this.name = apiData.name;
    this.startDatetime = parseISO(apiData.start_datetime);
    this.endDatetime = parseISO(apiData.end_datetime);
    this.status = toStatus(apiData.status);
    this.platform = toPlatform(apiData.platform);
    this.ownerLogin = apiData.owner_login;
  }
```

Then in the API we can do:

```typescript
result.experiments.map((apiData: ApiData) => new ExperimentBare(apiData))
```

models/Platform.ts
yanirs: Actually, given that the enum is defined using strings, it turns out this function is unnecessary. Can simply use `input as Platform`.

models/ExperimentBare.ts
yanirs: Update: Let's get rid of the `toStatus()` and `toPlatform()` functions, as `<str> as <enum>` seems to work:

```typescript
  constructor(apiData: ApiData) {
    this.experimentId = apiData.experiment_id
    this.name = apiData.name
    this.startDatetime = parseISO(apiData.start_datetime)
    this.endDatetime = parseISO(apiData.end_datetime)
    this.status = apiData.status as Status
    this.platform = apiData.platform as Platform
    this.ownerLogin = apiData.owner_login
  }
```

models/Platform.ts
dhurlburtusa: Yeah, I meant to ask you about the defaulting of the Platform and Status. I was only doing that because we have `status` and `platform` defined as not being nullable in the experiments interfaces. So, TS was complaining about assigning `Status | undefined` to `Status` (similarly for `Platform`).

> Actually, given that the enum is defined using strings, it turns out this function is unnecessary. Can simply use input as Platform.

I tried this. It compiles, but it results in 'Calypso' instead of 'calypso'.

yanirs: > I tried this. It compiles, but it results in 'Calypso' instead of 'calypso'.

Are you sure? I ran it inlined with the constructor from https://github.com/Automattic/abacus/pull/66#discussion_r425508873 and I'm seeing the lowercase version in the table.

dhurlburtusa: I am not sure what I saw, but the updated unit tests are passing with what I we are expecting. I am curious what TS is doing behind the scenes.

models/ExperimentBare.ts
dhurlburtusa: Works great! Thanks for the help!

models/ExperimentFull.ts
```diff
@@ -0,0 +1,111 @@
...
+    super(apiData)
+    this.conclusionUrl = apiData.conclusion_url || null
+    this.deployedVariationId = apiData.deployed_variation_id || null
+    this.description = apiData.description
+    this.endReason = apiData.end_reason || null
+    this.existingUsersAllowed = apiData.existing_users_allowed
+    this.exposureEvents = Array.isArray(apiData.exposure_events)
+      ? apiData.exposure_events.map((exposureEvent: ApiData) => ({
+          event: exposureEvent.event,
+          props: exposureEvent.props,
```
dhurlburtusa, line 84: @yanirs Should we make classes now for `Event`, `MetricAssignment`, `SegmentAssignments`, and `Variations` or save it for a future PR?

yanirs: Can keep it for a future PR. There's already plenty of stuff in this PR. :slightly_smiling_face:

PR #70, opened 2020-05-16
Add Experiments API Tests

### LINKED ISSUE #65: Add Tests for ExperimentsApi.ts
There are many ways to test the `api/ExperimentsApi.ts` since it makes actual fetch requests.

Unit tests can be created using mocks. The `fetch` function itself could be mocked. With this being mocked, what benefit are we really getting out of it? It would bypass the authorization code. It would test that we can mock a response that is parsable as JSON, the returned shape is correct, and we can correctly mimic the environment it will run in (browser with known `window.location.host` and `localStorage`).

To me, an integration test would be more beneficial than a unit test. Then the actual call using `fetch` is tested along with the authorization (assuming it is connected to an endpoint that using the authorization header).

An e2e test would be able to test the real thing in a real environment.

A lot of times with testing API calls, more work is involved creating mocks and setting up the appropriate environment. Sometimes I wonder if this is worth the effort.

dhurlburtusa:
Added Experiments API integration tests.

Resolves #65.

## How has this been tested?

`npm run verify` (which now includes integration tests. Unit test's coverage still at 100%.

Integration tests were set to run in CircleCI. Will check how they went after creating this PR.


api/ExperimentsApi.ts
```diff
@@ -20,17 +20,20 @@ function resolveApiUrlRoot() {
  */
 async function findAll(): Promise<ExperimentBare[]> {
   const apiUrlRoot = resolveApiUrlRoot()
-
-  const accessToken = getExperimentsAuthInfo()?.accessToken
-  if (!accessToken) {
-    throw new UnauthorizedError()
+  const isProduction = apiUrlRoot === PRODUCTION_API_URL_ROOT
```
yanirs, line 23: I see what you did here, but it took a while to understand. I'll push a small refactor to make it a bit more explicit.

dhurlburtusa: 👍

PR #72, opened 2020-05-17
Add Metrics API

dhurlburtusa:
Add metrics API so we can pull down the list of available metrics to use with experiment creation.

## How has this been tested?

Added unit/integration tests for new modules/classes.


api/MetricsApi.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+  }
+
+  const fetchUrl = `${apiUrlRoot}/metrics`
+  return fetch(fetchUrl, {
+    method: 'GET',
+    headers: isProduction ? new Headers({ Authorization: `Bearer ${accessToken}` }) : undefined,
+  })
+    .then((response) => response.json())
+    .then((result) => result.metrics.map((apiData: ApiData) => new MetricBare(apiData)))
```
yanirs, line 33: There's too much duplication between this `ExperimentsApi.findAll()`. If we abstract the common parts we'll get something like:

```suggestion
  return fetchApi('GET', '/metrics')
    .then((result) => result.metrics.map((apiData: ApiData) => new MetricBare(apiData)))
```

With `fetchApi` defined as:

```typescript
async function fetchApi(method: string, path: string, body: BodyInit | null = null) {
  const apiUrlRoot = resolveApiUrlRoot()

  let headers
  if (apiUrlRoot === PRODUCTION_API_URL_ROOT) {
    const accessToken = getExperimentsAuthInfo()?.accessToken
    if (!accessToken) {
      throw new UnauthorizedError()
    }
    headers = new Headers({ Authorization: `Bearer ${accessToken}` })
  }

  return fetch(`${apiUrlRoot}${path}`, {
    method: method,
    headers: headers,
    body: body
  }).then((response) => response.json())
}
```

api/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Makes a request to the Experiment Platform's API with any necessary
+ * authorization information, parses the response as JSON, and returns the parsed
+ * response.
+ *
+ * Note: Be sure to handle any errors that may be thrown.
+ *
+ * @throws UnauthorizedError
+ */
+async function fetchApi(method: string, path: string, body: BodyInit | null = null) {
```
withinboredom, line 17: It may be a good idea to use const enums instead of a string, or even just a whitelist possible set of strings to get some of that intellisense magic:

```suggestion
export const enum AvailableEnpoints { // or a better name?
  Experiments = '/experiments'
}

async function fetchApi(method: string, path: AvailableEndpoints, body: BodyInit | null = null) {
```

Or

```suggestion
type ExperimentEndpoints = "/experiments" | "/metrics"

async function fetchApi(method: string, path: ExperimentEndpoints, body: BodyInit | null = null) {
```

api/utils.ts
```diff
@@ -0,0 +1,36 @@
+import { getExperimentsAuthInfo } from '@/utils/auth'
+
+import UnauthorizedError from './UnauthorizedError'
+
+const DEVELOPMENT_API_URL_ROOT = 'https://virtserver.swaggerhub.com/yanir/experiments/0.1.0'
+const PRODUCTION_API_URL_ROOT = 'https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0'
```
withinboredom, line 6: Can we get this from the environment?

api/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Note: Be sure to handle any errors that may be thrown.
+ *
+ * @throws UnauthorizedError
+ */
+async function fetchApi(method: string, path: string, body: BodyInit | null = null) {
+  const apiUrlRoot = window.location.host === 'experiments.a8c.com' ? PRODUCTION_API_URL_ROOT : DEVELOPMENT_API_URL_ROOT
+
+  let headers
+  if (apiUrlRoot === PRODUCTION_API_URL_ROOT) {
```
withinboredom, line 21: Can we turn on Authentication from the environment instead of checking magic strings? It's easily outside of the scope of this PR, but just curious if we can head in that direction?

api/ExperimentsApi.ts
```diff
@@ -19,24 +11,8 @@ function resolveApiUrlRoot() {
...
-
-  const fetchUrl = `${apiUrlRoot}/experiments`
-  return fetch(fetchUrl, {
-    method: 'GET',
-    headers: headers,
-  })
-    .then((response) => response.json())
-    .then((result) => result.experiments.map((apiData: ApiData) => new ExperimentBare(apiData)))
+  const result = await fetchApi('GET', '/experiments')
+  return result.experiments.map((apiData: ApiData) => new ExperimentBare(apiData))
```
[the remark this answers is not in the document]
dhurlburtusa: I see how it will minimize lines. I, personally, find some things easier to understand with multiple lines, especially when using an async call. I find that many single line versions take longer to read and grok.
Also, even though I have a 32" monitor (and I am running native resolution), I usually have two or three editor tabs next to each other. Longer lines sometimes means I have to scroll horizontally. :(

api/utils.ts
dhurlburtusa: I like.

api/utils.ts
dhurlburtusa: I suggested the same thing in another PR. It think it would be better not to expose the development API URL root to production.
Plus, it will give the developer the ability to pick a different URL for some other purpose.

api/utils.ts
dhurlburtusa: I think it would be good to use the environment too. Then we are not checking magic strings, the production-only code does not exist in development but only exists in production.

api/ExperimentsApi.ts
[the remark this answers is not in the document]
withinboredom: I'm used to seeing it in some C# code I maintain :D so I'm biased. Good point about horizontal scrolling too.

api/utils.ts
withinboredom: When we get there, it might be worth making a `npm run switch-env production|staging|dev` that can enable certain environments. Might be worth doing something similar to Calypso (which you may not be familiar with) where we load a configuration json file depending on the environment set. 

So, we could have 3 configuration files that are checked in and a `.env` file to choose which environment. `npm run switch-env` could probably handle all that. Happy to make a PR for what I'm thinking of later this week.

dhurlburtusa: I think I have a good understanding of what you mean. I think it is a good idea to have an easy way to switch between environments. At the moment, however, I don't think we'll need it given the current requirements.

api/utils.ts
yanirs: This won't work well once we have dynamic strings. I think it overcomplicates things for this PR.

api/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+      throw new UnauthorizedError()
+    }
+    headers = new Headers({ Authorization: `Bearer ${accessToken}` })
+  }
+
+  return fetch(`${apiUrlRoot}${path}`, {
+    method,
+    headers,
+    body,
+  }).then((response) => response.json())
```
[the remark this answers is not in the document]
yanirs: Can also collapse it to one line if prettier allows. :slightly_smiling_face:

api/utils.ts
yanirs: Let's keep it to another PR. The environment-based setup just overcomplicates things right now for little gain.

api/utils.ts
yanirs: Also, keep in mind that this function is a helper that's internal to the API client. Outside the client, we will call specific functions. Let's not overdo type safety unnecessarily. :slightly_smiling_face:

api/utils.ts
yanirs: Maybe open a low priority issue for that? We need to sort out testing with authentication in different environments, and we don't even have production working yet, so let's keep the hardcoded approach for now. :slightly_smiling_face:

api/ExperimentsApi.ts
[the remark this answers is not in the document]
yanirs: I disagree with the horizontal scrolling argument. We have a line length limit, so minimising line length below the limit we agreed on is pointless.

That said, doing too much in a single statement can make the code hard to read. However, the definition of too much depends on what you're used to. For example, when I was new to Python I found list comprehensions very confusing, but then I got used to them.

In this case, I'm leaning towards @withinboredom's suggestion (I actually suggested the same on Slack to @dhurlburtusa yesterday), as I'm not a fan of too many temporary/one-off variables unless they increase readability. With a name like `result`, I'm not sure that readability is increased, but I don't feel too strongly about keeping it.

PR #73, opened 2020-05-17
Add  Segments API

dhurlburtusa:
Add segments API so we can pull down the list of available segments to use with experiment creation.

## How has this been tested?

Added unit/integration tests for new modules/classes.


api/SegmentsApi.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Finds all the available segments.
+ *
+ * Note: Be sure to handle any errors that may be thrown.
+ *
+ * @throws UnauthorizedError
+ */
+async function findAll(): Promise<Segment[]> {
+  const apiUrlRoot = resolveApiUrlRoot()
```
yanirs, line 16: The same duplication comment from #72 applies here.

PR #74, opened 2020-05-17
Add Metrics List

### LINKED ISSUE #12: Add metric list page
For the initial iteration, it should be very similar to the experiment list (currently the root page). So: use Material Table (similarly to the way it's used in the `ExperimentsTable` component) to show the results we get from the `/metrics` endpoint (`MetricBare` objects returned by `MetricsApi.findAll()`).

dhurlburtusa:
Resolves #12.

## How has this been tested?

Added unit test for new MetricsTable component.

Will need to added e2e tests once we can get past authentication.


components/MetricsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+  // The zero-based page of data to display. Defaults to 0.
+  initialPage?: number
+  // The number of rows/items to display per page. Defaults to PER_PAGE_DEFAULT.
+  perPage?: number
+}
+
+/**
+ * Renders a table of "bare" experiment information.
+ */
+const MetricsTable = (props: Props) => {
```
yanirs, line 25: This highlights an issue that I realised after the `ExperimentsTable` PR: Why are we using our own table components? There's a lot of duplication between between this and `ExperimentsTable`, and neither one does anything too fancy. I'm sure we'll want to do things like sorting, searching, and filtering, which would increase duplication and maintenance overheads. I think we should spend some time researching an off-the-shelf solution, replace `ExperimentsTable` with it, and then update this PR to use it.

withinboredom: Heh, off-the-shelf components are probably designed for server-side paging and searching and not for in-memory pages and searches. I could be wrong (I didn't research it).

I do agree that we should probably consolidate logic though, whether that's an off-the-shelf component or our own. It may be outside the scope of this PR though, as I think it's fair to say that it needs some research. We should aim to have it figured out by the time we add the next table though :)

components/MetricsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+  debug('MetricsTable#render')
+  const { metrics, initialPage = 0, perPage = PER_PAGE_DEFAULT } = props
+  const [page, setPage] = useState(initialPage)
+
+  const pageCount = Math.ceil(metrics.length / perPage)
+  const pageNums = Array.from(Array(pageCount).keys()).map((num) => num + 1)
```
withinboredom, line 31: This will likely not be an issue here, but it's worth point out that we probably don't want to calculate this every render. It might be better as part of state and `useEffect` to manage it.

```ts
function calculatePageData() {
  const pageCount = Math.ceil(metrics.length / perPage)
  const pageNums = Array.from(Array(pageCount).keys()).map((num) => num + 1)
  return [pageCount, pageNums]
}

const [[pageCount, pageNums], setPageData] = useState(calculatePageData) // note the lack of parens, so it's only calc'd once

useEffect(() => {
  setPageData(calculatePageData)
}, [metrics, perPage])
```

components/MetricsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  function handlePageStep(event: MouseEvent<HTMLAnchorElement>) {
+    debug('MetricsTable#handlePageStep')
+    event.preventDefault()
+    event.stopPropagation()
+    setPage(page + toInt(event.currentTarget.dataset.step))
+  }
```
withinboredom, line 38: This could probably be a higher order function so you don't have to rely on the rendered state:

```suggestion
  function handlePageStep(nextPage) => (event: MouseEvent<HTMLAnchorElement>) => {
    debug('MetricsTable#handlePageStep')
    event.preventDefault()
    event.stopPropagation()
    setPage(nextPage)
  }
```

components/MetricsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  function handleSetPage(event: MouseEvent<HTMLAnchorElement>) {
+    debug('MetricsTable#handleSetPage')
+    event.preventDefault()
+    event.stopPropagation()
+    setPage(toInt(event.currentTarget.dataset.page))
+  }
```
withinboredom, line 45: If you use the higher order function, this can be eliminated.

components/MetricsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+      {pageCount > 1 && (
+        <Table.Footer>
+          <Table.Row>
+            <Table.HeaderCell colSpan='3'>
+              <Menu floated='right' pagination>
+                <Menu.Item
+                  data-step={-1}
+                  disabled={page <= 0}
+                  icon='chevron left'
+                  onClick={handlePageStep}
```
withinboredom, line 79: If you use the higher order function you can just:

```suggestion
                  onClick={handlePageStep(page - 1)}
```

pages/metrics.tsx
```diff
@@ -0,0 +1,38 @@
...
+const MetricsPage = function MetricsPage() {
+  debug('MetricsPage#render')
+  const [error, setError] = useState<Error | null>(null)
+  const [metrics, setMetrics] = useState<MetricBare[] | null>(null)
+
+  useEffect(() => {
+    MetricsApi.findAll()
+      .then((metrics) => setMetrics(metrics))
+      .catch(setError)
+  }, [])
```
withinboredom, line 24: Can be simplified using `async`:

```suggestion
  useEffect(async () => {
    try {
      setMetrics(await MetricsApi.findAll())
    } catch (error) {
      setError(error)
    }
  }, [])
```

pages/metrics.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <Layout title='Metrics'>
+      <Container>
+        <img src='/img/logo.png' width='100' />
+        <h1>Metrics</h1>
+        {error && <ErrorsBox errors={[error]} />}
+        {metrics && <>{metrics.length === 0 ? <p>No metrics yet.</p> : <MetricsTable metrics={metrics} />}</>}
```
withinboredom, line 32: From being bit from this before, check that `metrics` is actually `null`. An empty array is falsy and a valid value.

withinboredom: Thanks for the slack ping @dhurlburtusa, indeed `[] && true === true` :TIL:!

components/MetricsTable.tsx
dhurlburtusa: Before doing research on some OTS solution, let's decide what features/requirements we are looking for in a table. Pagination is already implemented but can be changed to allow the user to select how many to display per page. I think adding filtering and sorting might be a good idea. However, I think they have lower priority than finishing the Experiment creation form first.

The one I know about is `react-virtualized`. I've seen it used in the project I do code reviews for another company. It seems overly convoluted/complicated (this may be partly due to no hands-on experience with it). So, I decided to KIS with similar code to what was used in the trial.

Yes, I definitely agree that there is a lot of duplication. I knew that when I created the PR.

Do you think it is better to start a discussion before creating a PR or to create the PR to have the discussion? I meant to put a comment in the PR to discuss this duplication.

pages/metrics.tsx
yanirs: How is this simpler? :slightly_smiling_face: 

The suggestion is longer and more syntax-heavy &ndash; I find it less readable.

components/MetricsTable.tsx
yanirs: > Heh, off-the-shelf components are probably designed for server-side paging and searching and not for in-memory pages and searches. I could be wrong (I didn't research it).

@withinboredom let's not make unsubstantiated claims. It seems highly unlikely that there are no existing solutions for rendering tables with client-side pagination. Here's [one example](https://www.npmjs.com/package/react-data-table-component) from a quick search.

> I do agree that we should probably consolidate logic though, whether that's an off-the-shelf component or our own. It may be outside the scope of this PR though, as I think it's fair to say that it needs some research. We should aim to have it figured out by the time we add the next table though :)

Strongly disagree about it being out of scope. Adding this amount of duplication and reinventing the wheel is unacceptable. Just a reminder of our top priority: **We're building for maintenance**. We already have an internal UI that covers this functionality, so there's no reason for another quick-and-dirty solution.

> Before doing research on some OTS solution, let's decide what features/requirements we are looking for in a table. Pagination is already implemented but can be changed to allow the user to select how many to display per page. I think adding filtering and sorting might be a good idea. However, I think they have lower priority than finishing the Experiment creation form first.

This entire PR is a lower priority than experiment creation, which raises the question why we're even discussing it now.  In addition, there may be some context missing, but views for experiments &ndash; including a polished list &ndash; are a higher priority than managing metrics and segments. The experiment pages block the initial launch with a small set of experimenters, while metric/segment management don't

I agree on clarifying requirements as part of researching an OTS solution &ndash; opened #77 to do that. Let's close this PR for now until #77 is addressed.

> Yes, I definitely agree that there is a lot of duplication. I knew that when I created the PR.
>
> Do you think it is better to start a discussion before creating a PR or to create the PR to have the discussion? I meant to put a comment in the PR to discuss this duplication.

As a rule of thumb, duplicating hundreds of lines in code that we're supposed to maintain isn't something that I'm likely to agree to. :slightly_smiling_face: So discussing prior to implementation is strongly encouraged. Citing our coding principles again: _"for non-trivial changes, it is often best to get some alignment within the team before jumping to write code. In fact, some pre-coding discussions may save time on PR reviews."_

yanirs:
Following https://github.com/Automattic/abacus/pull/74#discussion_r427027410, I'm closing this until #77 is addressed.

PR #78, opened 2020-05-19
Add an isSnakeCase Util

dhurlburtusa:
Add an isSnakeCase util which we'll need for validation.

## How has this been tested?

A unit test was added. Coverage is still at 100%.


utils/isSnakeCase.ts
```diff
@@ -0,0 +1,18 @@
...
+const SNAKE_CASE_RE = /^[a-z0-9]+(:?_[a-z0-9]+)*$/
+
+/**
+ * Determines whether the value given is in snake_case.
+ */
+function isSnakeCase(value: unknown): boolean {
+  let matches = false
+  if (typeOf(value) === 'string') {
+    const input = value as string
+    matches = !input.includes('__')
```
yanirs, line 12: I think that double underscores still count as snake case.

utils/isSnakeCase.ts
```diff
@@ -0,0 +1,18 @@
+import { typeOf } from 'qc-type_of'
+
+const SNAKE_CASE_RE = /^[a-z0-9]+(:?_[a-z0-9]+)*$/
+
+/**
+ * Determines whether the value given is in snake_case.
+ */
+function isSnakeCase(value: unknown): boolean {
```
yanirs, line 8: Why would we call this function with non-strings? It'd be good to let the Typescript compiler handle the type check rather than use `typeOf`.

yanirs:
> Add an isSnakeCase util which we'll need for validation.

Can you please push a WIP PR/branch that shows how this will be used? As noted inline, I'm not sure about the need for a type check. I'm also unsure about the need for this at all, which is hard to judge without extra context &ndash; I'm wondering whether we can use [the `pattern` attribute for form validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern).

utils/isSnakeCase.ts
dhurlburtusa: Okay. If sequential underscores are acceptable, then I'll just use a regexp and the pattern attribute. I'll use the same regexp as in the swagger file.

utils/isSnakeCase.ts
dhurlburtusa: The thought was to make this a general purpose utility that handles any data type. If the data type is not a string, then it can't be in snake_case.

dhurlburtusa:
Just going to use regexp and the pattern attribute.

PR #82, opened 2020-05-25
Add Experiment Creation Foundation

dhurlburtusa:
Adds create function to Experiments API which is common code needed to prepare for implementing the experiments creation form with different libraries.

## How has this been tested?

Add integration test for `ExperimentsAPI.create`.


models/ExperimentFull.ts
```diff
@@ -108,4 +111,45 @@ export class ExperimentFull extends ExperimentBare {
       variationId: variation.variation_id,
     }))
   }
+
+  // eslint-disable-next-line @typescript-eslint/naming-convention
```
yanirs, line 115: Is this to avoid calling the method `toJson()`? I think that the non-capitalised version is preferable. Otherwise, things get weird with multiple acronyms, e.g., `toAPIJSON()` is less readable than `toApiJson()`.

models/ExperimentBare.ts
```diff
@@ -40,8 +46,12 @@ export class ExperimentBare {
   constructor(apiData: ApiData) {
     this.experimentId = apiData.experiment_id
     this.name = apiData.name
-    this.startDatetime = parseISO(apiData.start_datetime)
-    this.endDatetime = parseISO(apiData.end_datetime)
+    this.startDatetime =
```
yanirs, line 49: Why is this needed? My thinking with the `apiData` constructor was that we'd have a separate constructor to use from JS code. Also, we probably don't ever need to create an `ExperimentBare` from the client, so the new constructor can be in `ExperimentFull`.

models/ExperimentBare.test.ts
```diff
@@ -26,6 +26,26 @@ describe('models/ExperimentBare.ts module', () => {
           status: Status.Staging,
         })
       })
+      it('called with valid API data using `Date` instances should create a new `ExperimentBare` instance', () => {
```
yanirs, line 29: See comment below, but valid API data _shouldn't_ have `Date` instances, so this is a bit misleading.

api/ExperimentsApi.ts
```diff
@@ -1,8 +1,17 @@
...
 import { ApiData } from './ApiData'
 import { fetchApi } from './utils'
 
+/**
+ * Attempts to create a new experiment.
+ *
+ * Note: Be sure to handle any errors that may be thrown.
+ */
+async function create(experiment: ExperimentFull) {
+  return await fetchApi('POST', '/experiments', JSON.stringify(experiment))
```
yanirs, line 12: I think it'd be better to have the call to `JSON.stringify()` in `fetchApi()`.

models/ExperimentFull.ts
yanirs: Actually, as this doesn't return a JSON, maybe name it `toApiData` and make the return type explicit?

models/ExperimentFull.ts
```diff
@@ -77,28 +80,28 @@ export class ExperimentFull extends ExperimentBare {
     this.deployedVariationId = apiData.deployed_variation_id || null
     this.description = apiData.description
     this.endReason = apiData.end_reason || null
-    this.existingUsersAllowed = apiData.existing_users_allowed
+    this.existingUsersAllowed = toBool(apiData.existing_users_allowed)
```
yanirs, line 83: Doesn't the API return a boolean correctly?

package.json
```diff
@@ -20,6 +20,7 @@
     "next": "^9.3.5",
     "qc-to_bool": "^1.1.2",
     "qc-to_int": "^1.0.1",
+    "qc-type_of": "^1.0.0",
```
yanirs, line 23: Where is this used?

models/ExperimentFull.ts
yanirs: Ah, I see, you're overriding the default `toJSON()`. Never mind. :slightly_smiling_face:

models/ExperimentFull.test.ts
```diff
@@ -198,5 +198,171 @@ describe('models/ExperimentFull.ts module', () => {
         })
       })
     })
+
+    describe('toJSON', () => {
```
yanirs, line 202: To make things consistent, let's either use `JSON.stringify()` in all tests or `toJSON()` in all tests and make the description match what we're doing.

models/ExperimentFull.ts
```diff
@@ -108,4 +111,45 @@ export class ExperimentFull extends ExperimentBare {
...
+      deployed_variation_id: this.deployedVariationId,
+      end_reason: this.endReason,
+      existing_users_allowed: this.existingUsersAllowed,
+      p2_url: this.p2Url,
+      metric_assignments: this.metricAssignments.map((metricAssignment: ApiData) => ({
+        attribution_window_seconds: metricAssignment.attributionWindowSeconds,
+        change_expected: metricAssignment.changeExpected,
+        experiment_id: metricAssignment.experimentId,
+        is_primary: metricAssignment.isPrimary,
+        metric_assignment_id: metricAssignment.metricAssignmentId,
```
yanirs, line 136: When passed into the API, we often don't have `metric_assignment_id`, `segment_assignment_id`, and `variation_id`. They should be omitted, as they are assigned by the server.

models/ExperimentBare.ts
```diff
@@ -40,8 +46,12 @@ export class ExperimentBare {
...
+      apiData.start_datetime instanceof Date
+        ? apiData.start_datetime
+        : parseISO(stripTimeAndTzInfo(apiData.start_datetime))
```
yanirs, line 52: I think we discussed it on Slack, but we shouldn't strip the time component as it's not going to be round dates for experiments that started/stopped.

models/ExperimentBare.ts
yanirs: Actually, TS support for multiple constructors is too weird/incomplete, so we should probably do something like this: https://stackoverflow.com/a/38688403

models/ExperimentBare.test.ts
dhurlburtusa: @yanirs True. I think changing this constructor to work with the JS naming convention (i.e, camelCase) and the correct data types would be good. Then creating separate static factory functions to handle creating an Experiment instance from apiData would be good.

Thoughts?

api/ExperimentsApi.ts
dhurlburtusa: Okay. I can see that. I'll move it to `fetchApi`.

package.json
dhurlburtusa: It might be a left over from using it before I committed the code. I'll double check.

models/ExperimentFull.ts
dhurlburtusa: @yanirs The API does. It was a result of trying to use this constructor with the form data. I think I'll add a separate static factory function for form data or pre-transform the form data before passing it to the new constructor (camelCase, correct data types).

Thoughts on which choice may be better?

models/ExperimentFull.ts
dhurlburtusa: > Actually, as this doesn't return a JSON, ...

Yeah, I am really disappointed in the choice of the function name we are required to use in order to serialize objects as JSON. I wish `JSON.stringify` would have picked a different name. So, instead, I try to think about the function as a helper function that ultimately leads to JSON.

models/ExperimentFull.ts
dhurlburtusa: I wasn't 100% sure if we were going to need them for the experiment details page since we don't have a design for it yet. I was imagining that we would need the IDs when an a12n deletes one of them. So, I decided to just include them. I'll go ahead and remove them. If we do need them, then we can add them back.

models/ExperimentFull.test.ts
dhurlburtusa: I think it is better to always use `JSON.stringify` in the tests to ensure we are generating what the API is expecting.

I thought there are a reason why I also made the call directly to `ExperimentFull#toJSON`. I don't recall what that reason was. If I don't remember, I'll remove it. If I do and it is still important to make the direct call, then I'll document why it is necessary.

api/ExperimentsApi.ts
yanirs: Maybe we can make `fetchApi()` accept `any` as the type of its third argument?

models/ExperimentFull.test.ts
yanirs: Sounds good. :slightly_smiling_face: It's just a bit confusing that the behaviour isn't consistent across the three test cases.

models/ExperimentFull.ts
yanirs: We do need them when reading from the API, but not when writing. That's why they're marked as `readOnly` in the API docs. Does that make sense?

models/ExperimentBare.ts
dhurlburtusa: Thanks for pointing that out. This makes it work with the data coming from the form but breaks the data coming from the API. I'll pull it out and use it with the form data.

models/ExperimentBare.test.ts
yanirs: I think we can work with the solution proposed in https://stackoverflow.com/a/38688403, but I'll play with it a bit to make sure. I'm planning on pushing a PR with the models for the analysis page soon, so we can see what it looks like and discuss based on that. I'll work on it now. :slightly_smiling_face:

models/ExperimentFull.ts
dhurlburtusa: Yes, that makes sense. I'll remove the data that doesn't need to be sent to the server for creation.

Since we are currently using this specifically for serializing the data for the experiment creation API, we can't use it as a general serialization mechanism. I am starting to think it might be better to have specific functions for serialization for specific purposes. We may need one for when "patching" different fields.

@yanirs What do you think about changing this function to use JSON.stringify internally and rename it something like `toCreateApiBody`? Since this is currently speculation, we probably should follow YAGNI at this point.

models/ExperimentBare.test.ts
yanirs: Opened #84.

models/ExperimentFull.ts
yanirs: Might be YAGNI, but it is a good point about the generality of the conversion to JSON. Maybe we should combine this with the ideas of moving `JSON.stringify()` into `fetchApi()` and defining `fromApiData()` as a static method? Then it becomes natural to add `toApiData()`. Something like:

```typescript
// New interface to be implemented by models that can be written to the API.
interface ApiInput {
  toApiData(): ApiData
}

// Model example -- we will populate this with real values, as in the toJSON example
class Model implements ApiInput {
  toApiData(): ApiData {
    return {
      field: 'value'
    }
  }
}

// New signature and body conversion for fetchApi
async function fetchApi(method: string, path: string, body: ApiInput | null = null) {
  // auth, etc.

  return (
    await fetch(`${apiUrlRoot}${path}`, {
      method,
      headers,
      body: body === null ? null : JSON.stringify(body),
    })
  ).json()
}
```

What do you think?

dhurlburtusa: I think that will work (except I think you meant the following (see below)) but at this point I am not really seeing what it is buying us. Mainly due to only having one model being serialized at this point. I might be able to see the benefit once other models are being serialized. What do you see the benefit of moving JSON.stringify into `fetchApi`?
```
      body: body === null ? null : JSON.stringify(body.toApiData()),
```
Basically, it looks like we are moving part of the responsibility of serialization of the models into the `fetchApi` function but we need to introduce an interface to do so - which is not necessarily a bad thing. It is nice to have the interface to enforce consistency between the models.

We'll need to allow updates to experiments when we implement the details page. I took a look at the API schema. A `PatchRequest` exists that is an array of the field changes. Would this become a separate model that implements `ApiInput` too? Are there any other API requests that include a body we should consider? Looks like there will be POST and PUT requests for Metrics. Segments will just have a POST.

I'm inclined to move forward with the current implementation until we have more example to see a pattern. Then we can change implementation details later. Because fetchApi is effectively private, its implementation can be changed as necessary later too.

yanirs: > I think that will work (except I think you meant the following)

Yep, that's what I meant. Good catch!

> What do you see the benefit of moving JSON.stringify into `fetchApi`?

With the current implementation, it's assumed that any non-null `body` is a JSON string, but this isn't enforced on the type level, as `BodyInit` can be many other things. If we specify the interface and stringify the JSON in `fetchApi()`, it all becomes more explicit, clearer, and DRYer.

> I took a look at the API schema. A `PatchRequest` exists that is an array of the field changes. Would this become a separate model that implements `ApiInput` too?

PATCH isn't implemented yet, but yes, a `PatchRequest` would likely be a separate class that implements `ApiInput`.

> Are there any other API requests that include a body we should consider? Looks like there will be POST and PUT requests for Metrics. Segments will just have a POST.

Yes, and PUT for Experiments. In all cases we will be stringifying a model.

> The potential issue is if the same model instance needs to be serialized in different ways. How does `fetchApi` know which one to use?

I don't see a concrete case of this being needed now. As you noted, using `toJSON()` is somewhat misleading/confusing because it changes the field names and omits some fields.

> I think we will need to be able to do the following with the `ExperimentsApi` module:

Not really. I think it'll be more like:

```typescript
async function create(experiment: ExperimentFull) {
  return await fetchApi('POST', '/experiments', experiment)
}

async function put(experiment: ExperimentFull) {
  return await fetchApi('PUT', '/experiments/' + experiment.experimentId, experiment)
}

async function patch(experimentBefore: ExperimentFull, experimentAfter: ExperimentFull) {
  // some code to create patchRequest based on the difference between the experiments
  return await fetchApi('PATCH', '/experiments/' + experimentBefore.experimentId, patchRequest)
}
```

Let's worry about PATCH later, though. I don't see us implementing it anytime soon.

api/ExperimentsApi.ts
withinboredom: > accept any as the type of its third argument?

Generally try to avoid `any`, but you can use `|`, so `experiment: ExperimentFull | ExperimentBare` and then `if(experiment is ExperimentBare) doBareThings();` Even better if you can have a common base type and then use discriminators to cast to the appropriate type.

models/ExperimentFull.test.ts
```diff
@@ -198,5 +198,171 @@ describe('models/ExperimentFull.ts module', () => {
...
+      it('called on new experiment with minimal fields.', () => {
+        const experiment = new ExperimentFull({
```
withinboredom, line 204: It's probably a good idea to break these experiment objects into a factory method so if and when we change the schema we only need to change it in a couple of places instead of dozens.

models/ExperimentFull.ts
withinboredom: Can't we just return a bool from the form data?

dhurlburtusa: > Can't we just return a bool from the form data?
@withinboredom Yep. I plan to move some of the transformation to the form before an ExperimentFull instance is created.

models/ExperimentFull.test.ts
dhurlburtusa: Got rid of `toJSON`. Basically renamed to `toApiData`.

models/ExperimentFull.test.ts
dhurlburtusa: @withinboredom Would you elaborate and/or clarify? When you say "experiment objects", are you talking about the object literal passed to the constructor? Or are you talking about the experiment instance? Or something else?

models/ExperimentBare.ts
```diff
@@ -4,11 +4,33 @@ import { ApiData } from '@/api/ApiData'
...
+      platform: apiData.platform as Platform,
+      ownerLogin: apiData.owner_login,
+    })
+  }
+
   /**
    * Unique experiment ID.
    */
-  readonly experimentId: number
+  readonly experimentId?: number
```
dhurlburtusa, line 33: After changing the ExperimentFull constructor to accept datatype correct values, TypeScript wanted the experiment ID to be a number. We won't have this information when we are initially creating an experiment.

Had to do something similar for `segment_assignment_id`, `metric_assignment_id`, `variation_id`, etc.

Do this this is right approach?

models/ExperimentBare.ts
```diff
@@ -4,11 +4,33 @@ import { ApiData } from '@/api/ApiData'
...
 
+export interface ExperimentBareData {
+  experimentId?: number
+  name: string
+  startDatetime: Date
+  endDatetime: Date
+  status: Status
+  platform: Platform
+  ownerLogin: string
+}
```
dhurlburtusa, line 15: Not sure if there is a better to handle the data coming into the constructor. This duplicates the properties of the class.

models/ExperimentFull.test.ts
```diff
@@ -53,7 +53,7 @@ describe('models/ExperimentFull.ts module', () => {
           conclusionUrl: null,
           deployedVariationId: null,
           description: 'An example experiment.',
-          endDatetime: new Date(2020, 1, 29),
+          endDatetime: new Date(Date.UTC(2020, 1, 29)),
```
dhurlburtusa, line 56: Note: Creating a date using Date.UTC so that we get the same results for the different timezones we are in.

models/ExperimentFull.ts
```diff
@@ -71,41 +132,54 @@ export class ExperimentFull extends ExperimentBare {
...
+    this.variations = data.variations.slice()
+  }
+
+  toApiData() {
+    return {
+      experiment_id: this.experimentId,
+      name: this.name,
+      description: this.description,
+      start_datetime: formatIsoUtcOffset(this.startDatetime),
+      end_datetime: formatIsoUtcOffset(this.endDatetime),
```
dhurlburtusa, line 155: Note: Using this special function so that we can get consistent results regardless of which timezone we test the code. Otherwise a function such as `date-fns/formatISO` will have a `-06:00` offset for me at this time of year.

utils/date.ts
```diff
@@ -0,0 +1,13 @@
+import format from 'date-fns/format'
+
+/**
+ * Formats the date in ISO-8601 format with a UTC timezone value plus a `+00:00`
+ * offset.
+ *
+ * @param date
+ */
+function formatIsoUtcOffset(date: Date) {
+  return format(new Date(date.getTime() + date.getTimezoneOffset() * 60 * 1000), "yyyy-MM-dd'T'HH:mm:ss.SSS'+00:00'")
```
dhurlburtusa, line 10: Note: I can get the same effect by using `date-fns-timezone/format`. But this library either doesn't tree-shake properly or it depends on a lot of the timezone information it comes with. I thought it would be better to write a simple function to meet our needs instead of adding an extra 34KB (not that this is very much).

models/ExperimentBare.ts
yanirs: Yes, there is. I'm not sure why we're not following the pattern from #84.

models/ExperimentFull.test.ts
```diff
@@ -198,5 +198,130 @@ describe('models/ExperimentFull.ts module', () => {
...
+          [
+            {
+              attributionWindowSeconds: MetricAssignmentAttributionWindowSecondsEnum.OneWeek,
+              changeExpected: true,
+              isPrimary: true,
+              metricId: 4,
+              minDifference: 0.05,
+            },
+          ],
+        )
```
dhurlburtusa, line 278: ¡Ay, caramba!
Sorry Uncle Bob. I know it is recommended to use 3 or fewer parameters. But this is TypeScript and you weren't thinking about TypeScript but other languages. Maybe your best practice still applies?

models/MetricAssignment.ts
```diff
@@ -5,12 +5,12 @@ export interface MetricAssignment {
   /**
    * Globally-unique assignment ID.
    */
-  readonly metricAssignmentId: number
+  readonly metricAssignmentId?: number
```
dhurlburtusa, line 8: Note: Checked to see if we can make this required. Nope, unless we want to declare them explicitly each time.

models/ExperimentBare.test.ts
```diff
@@ -7,22 +7,45 @@ describe('models/ExperimentBare.ts module', () => {
   describe('ExperimentBare', () => {
     describe('constructor', () => {
```
yanirs, line 8: We don't need to test the constructor since it's auto-generated. See #84: I added `// istanbul ignore next (skip coverage for auto-generated constructor)` comments before the constructor so it doesn't count towards coverage.

models/ExperimentBare.ts
```diff
@@ -5,45 +5,25 @@ import { ApiData } from '@/api/ApiData'
...
-
-  constructor(apiData: ApiData) {
-    this.experimentId = apiData.experiment_id
-    this.name = apiData.name
-    this.startDatetime = parseISO(apiData.start_datetime)
-    this.endDatetime = parseISO(apiData.end_datetime)
-    this.status = apiData.status as Status
-    this.platform = apiData.platform as Platform
-    this.ownerLogin = apiData.owner_login
+  static fromApiData(apiData: ApiData) {
```
yanirs, line 8: Can you please add docs as in #84?

models/ExperimentFull.test.ts
yanirs: Yeah, let's not initialise like by calling the constructor directly, as it's error prone. Better change to:

```
const experiment: ExperimentFull = {
  experimentId: null,
  name: 'my_experiment',
  ...
}
```

TypeScript will complain if we get any types wrong or forget fields.

models/ExperimentFull.ts
```diff
@@ -1,111 +1,119 @@
...
   Event,
   MetricAssignment,
   MetricAssignmentAttributionWindowSecondsEnum,
+  Platform,
   SegmentAssignment,
+  Status,
   Variation,
 } from './index'
 
 export class ExperimentFull extends ExperimentBare {
```
dhurlburtusa, line 15: Note: TypeScript didn't require that we add `extends ApiDataSource`. @yanirs Do you think we should add it to be explicit?

utils/date.ts
yanirs: It'd be good to also add this note to the docstring.

models/ExperimentFull.test.ts
yanirs: I don't think it's worth breaking up right now, as the tests aren't too repetitive yet.

models/ExperimentFull.ts
```diff
@@ -7,9 +9,46 @@ import {
...
   Variation,
+  Status,
+  Platform,
 } from './index'
 
-export class ExperimentFull extends ExperimentBare {
+/**
+ * An experiment with full data.
+ */
+export class ExperimentFull extends DataTransferObject<ExperimentFull> {
```
yanirs, line 19: Changed to inherit from `DataTransferObject<ExperimentFull>` since the null constraints are different and I'm not sure it will work with inheritance from `ExperimentBare`. Might need to play with it more.

dhurlburtusa: I like the DTO abstract class. Good find.

api/ExperimentsApi.ts
dhurlburtusa: @withinboredom We decided to create an interface the model classes will implement to provide the data the API expects.

models/ExperimentFull.ts
```diff
@@ -71,41 +116,88 @@ export class ExperimentFull extends ExperimentBare {
    */
   deployedVariationId?: number | null
```
dhurlburtusa, line 117: @yanirs Should this and the other properties be declared readonly too?

dhurlburtusa:
@yanirs There are two remaining comments to be resolved. They have questions for you. Would you please take a look? Thanks.

__tests__/api/ExperimentsApi.test.ts
```diff
@@ -1,11 +1,40 @@
...
+        }),
+      )
+      // We expect that the response will return the new experiment with its newly
+      // assigned ID. These integration tests test against the "development" API which
+      // only returns mock data. So, instead of trying to keep in sync with the actual
+      // mock values, we'll test that it is valid to create an instance from the data.
+      expect(apiData).toBeDefined()
+      expect(() => {
+        ExperimentFull.fromApiData(apiData)
+      }).not.toThrowError()
```
dhurlburtusa, line 35: @withinboredom @yanirs This is what I came up with for testing the create experiment functionality. Do you think this is sufficient? Ideas for improvement?

models/ExperimentFull.ts
yanirs: Yeah, worth adding. :slightly_smiling_face: 

I think that TypeScript figures it out based on it having the `toApiData()` function.

models/ExperimentFull.ts
yanirs: Yeah, I missed those ones.

__tests__/api/ExperimentsApi.test.ts
yanirs: Yeah, I think it's good enough for now, but shouldn't the return value from `create()` be an `ExperimentFull` instance rather than `apiData`?

models/ExperimentFull.ts
dhurlburtusa: Right, and although it is not necessary, I think being explicit in this case makes it more clear of the intention.

PR #83, opened 2020-05-25
Add Experiment Creation Form with React Hook Form

dhurlburtusa:
A give-up implementation of the Experiment creation form with minimal UI layout for easier comparison.

## How has this been tested?

This is not meant to be merged. It is to be compared with another WIP PR that will be committed today or tomorrow.

Also, note that this is not intended to be a polished product. There may be some extra stuff laying around that is meant to spark a conversation or just because I thought it would be better to get feedback before spending too much time to implementing something we won't need.


pages/new.tsx
```diff
@@ -0,0 +1,24 @@
+import debugFactory from 'debug'
```
yanirs, line 1: Let's put this under `/experiments/new`.

package.json
```diff
@@ -14,16 +14,22 @@
     "node": ">=10.18.1"
   },
   "dependencies": {
+    "classnames": "^2.2.6",
     "date-fns": "^2.13.0",
     "debug": "^4.1.1",
     "isomorphic-unfetch": "^3.0.0",
+    "json-stringify-safe": "^5.0.1",
```
dhurlburtusa, line 21: Don't appear to be used in this PR.

package.json
```diff
@@ -39,14 +45,19 @@
     "@storybook/react": "^5.3.18",
     "@testing-library/jest-dom": "^5.5.0",
     "@testing-library/react": "^10.0.3",
+    "@types/classnames": "^2.2.10",
     "@types/debug": "^4.1.5",
     "@types/expect-puppeteer": "^4.4.1",
     "@types/jest": "^25.2.1",
     "@types/jest-environment-puppeteer": "^4.3.1",
+    "@types/json-stringify-safe": "^5.0.0",
```
dhurlburtusa, line 53: Don't appear to be used in this PR.

pages/new.tsx
dhurlburtusa: Cool. That would be my preference.

@yanirs Do we still want the experiment list at `/` or should we move it to `/experiments`? If so, what do we want at `/`? 

In an interface like Abacus, it is not uncommon to have a dashboard at `/` after the user authenticates. I don't think there is a need at the moment for a dashboard but maybe in the future.

yanirs: I think we can keep the experiment list under `/` for now, but I agree that having a dashboard there with the experiment list under `/experiments` makes more sense once we have more exciting stuff to show. :slightly_smiling_face: Can do that as part of the polishing iterations.

components/ExperimentForm.tsx
```diff
@@ -0,0 +1,492 @@
...
+import React from 'react'
+import ReactDatePicker from 'react-datepicker'
+import { Controller, useFieldArray, useForm } from 'react-hook-form'
+import ReactSelect from 'react-select'
+
+import ExperimentsApi from '@/api/ExperimentsApi'
+
+import { ExperimentFull, Platform } from '@/models/index'
+
+const SNAKE_CASE_RE = /^[a-z][a-z0-9_]*[a-z0-9]$/
```
withinboredom, line 16: Regex's are starting to become scattered all over the code base without much (if any) documentation which requires parsing regex's to understand what they do, but no idea what they should do. Can we start a `regex.ts` file that we can then import these from.

dhurlburtusa: I'll try to remember to centralize our regexp in a module when doing the React Final Form code. Please remind me if I forget.

dhurlburtusa:
We have decided to say goodbye to React Hook Form and use React Final Form.

Issue #11, opened 2020-04-14
Add experiment analysis page


yanirs:

yanirs:
Closing in favour of #95, #96, #97, and #98.

PR #84, opened 2020-05-27
Add analysis models

yanirs:
As part of addressing #11, add models to reflect what we get from the API.

This implementation diverges from the other models following https://github.com/Automattic/abacus/pull/82#discussion_r430088747 &ndash; see inline comments for thoughts. In general, we need a nice uniform interface for our models, and I think the approach here is cleaner than what we have so far.

## How has this been tested?

`npm run verify`, which includes tests for the new functionality.


models/Analysis.ts
```diff
@@ -0,0 +1,87 @@
+import parseISO from 'date-fns/fp/parseISO'
+
+import { ApiData } from '@/api/ApiData'
+
+// TODO: add docs
```
yanirs, line 5: I removed the auto-generated docs when shuffling things around, but I think we should have something &ndash; see `MetricEstimate` for an example.

models/Analysis.ts
```diff
@@ -0,0 +1,87 @@
...
+class Recommendation {
+  constructor(
```
yanirs, line 7: A few changes from the other models:

* Declare the fields as constructor arguments to take advantage of TypeScript auto-assignment. See [this post](https://www.stevefenton.co.uk/2013/04/stop-manually-assigning-typescript-constructor-parameters/), among others.
* Make all the fields `readonly`, as we never want them assigned outside of the constructor.
* Add `fromApiData()` as a static method, which allows direct initialisation using data we get from the API.

The main downside I see with this approach is that it gets messy with classes that have many arguments (probably only `ExperimentFull` at this stage). However, we can work around it in such cases by adding a `fromFormData()` method, which feels a bit redundant... Not sure. :slightly_smiling_face:

yanirs: > The main downside I see with this approach is that it gets messy with classes that have many arguments (probably only `ExperimentFull` at this stage). However, we can work around it in such cases by adding a `fromFormData()` method, which feels a bit redundant... Not sure. :slightly_smiling_face:

Actually, it's not needed, as the following seems to work:

```typescript
const rec: Recommendation = {
  endExperiment: true,
  chosenVariationId: 2,
  reason: RecommendationReason.CiGreaterThanRope,
  warnings: []
}
```

I really need to learn TypeScript properly. :slightly_smiling_face:

yanirs: Replaced the `new this()` calls with type-checked initialisers with named arguments: https://github.com/Automattic/abacus/pull/84/commits/3a73db17ecc123759e50379738853529b4cf4e9d

yanirs:
@dhurlburtusa @withinboredom this is now ready for review.

models/Analysis.ts
```diff
@@ -0,0 +1,136 @@
...
+   * Create an instance from raw API data (parsed JSON).
+   *
+   * @param apiData Raw API data.
+   */
+  static fromApiData(apiData: ApiData): Recommendation {
+    return {
+      endExperiment: apiData.end_experiment,
+      chosenVariationId: apiData.chosen_variation_id,
+      reason: apiData.reason as RecommendationReason,
+      warnings: apiData.warnings.map((warning: string) => warning as RecommendationWarning),
```
withinboredom, line 35: If it's possible for `apiData.warnings` to be null:

```suggestion
      warnings: apiData.warnings?.map((warning: string) => warning as RecommendationWarning),
```

yanirs: It isn't. :slightly_smiling_face:

models/Analysis.test.ts
```diff
@@ -0,0 +1,130 @@
+import { Analysis, AnalysisStrategy, RecommendationReason, RecommendationWarning } from './Analysis'
+
+describe('models/Analysis.ts module', () => {
+  describe('Analysis', () => {
+    describe('fromApiData', () => {
+      it('with null estimates and recommendation should create a valid instance', () => {
+        const analysis = Analysis.fromApiData({
```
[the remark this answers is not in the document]
yanirs: I see what you mean, but I intentionally changed most of the fields between the three test cases so there aren't any reasonable defaults. Can consider adding it later if we do find that it makes things cleaner.

yanirs:
@withinboredom Thanks for the review. Following a Slack discussion with @dhurlburtusa and the suggestion from https://github.com/microsoft/TypeScript/issues/3895#issuecomment-488355369, I added `DataTransferObject` as the base class, which makes it possible to instantiate the classes even when they have non-static methods. I'll merge now to unblock #82. If you see any issues with this approach we can address them in a new PR.

PR #85, opened 2020-05-27
WIP: Add Experiment Creation Form with React Final Form

dhurlburtusa:
An initial implementation of the the experiment creation form with minimal UI layout for easier comparison.

Note: This is not intended to be a polished product. Not all the validation is complete. Console.logs are still present. That will be cleaned up in the final product. Please focus on the differences between the different implementations.

Note: There are some differences between this implementation and the React Hook Form implementation besides the obvious library difference. In this implementation, separate field components have been created. The beginning of a validation/message "library" was added. This was not done with the RHF implementation because I don't really want to spend any more time with that library/framework.

## How has this been tested?

This is just the start. Several unit tests (for messages.ts, valiators.ts, etc) need to be added. Postponing that until implementation is approved to use.


components/ExperimentForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+  existing_users_allowed: boolean
+  metric_assignments: object[]
+  name: string
+  owner_login: { label: string; value: string }
+  p2_url: string
+  platform: string
+  segment_assignments: object[]
+  start_datetime: Date
+  status: string
+  variations: { name: string; allocatedPercentage: number }[]
```
dhurlburtusa, line 63:
```suggestion
  variations: { name: string; allocated_percentage: number; is_default: boolean }[]
```

components/ReactSelectField.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const debug = debugFactory('project:components/ReactSelectField.tsx')
+
+interface Props {
+  input: {
+    name: string
+    [attrName: string]: string
+  }
+  label?: ReactNode
+  options: { label: string; value: string }[]
```
dhurlburtusa, line 17: This goes along with the input component. Move under `input`.

dhurlburtusa:
> I like this one much better, but ultimately, which one did you find easiest to work with should be the one we go with.

React Final form is 1000% better than React Hook Forms.

jessie-ross:
I may slide my 2c in here:

I do recommend something that works with a Schema lib like [Yup](https://github.com/jquense/yup) or [JOI](https://hapi.dev/module/joi/).

They provide run-time schema validation and coercion which means:
- We can validate the items coming in from the server.
   - Currently we have no validation
- We can type coerce the values coming from the server, i.e. dates
- We can provide user-level validation for forms
- We can perform integration tests

The overall benefits would be:
- Reduced code duplication across server validation, coercion, user validation, and integration testing.
- We get run-time tests at the boundaries of our system ensuring the data coming in is as we expect.

yanirs:
Closing as this PR is now stale. @jessie-ross can still take code from this branch for their work on the experiment creation form.

PR #86, opened 2020-05-27
Add Analysis API client and initial experiment page

yanirs:
As part of addressing #11, fetch analyses from the API and display the stringified JSON on the experiment page.

## How has this been tested?

* Ran `npm run verify`, which includes integration tests for `AnalysesApi`
* Navigated to https://abacus-git-add-analysis-client.a8c.now.sh/experiments/1 and verified that the mock analyses are displayed as a JSON string
* Navigated to https://abacus-git-add-analysis-client.a8c.now.sh/experiments/not-a-number and verified that an error message is displayed


pages/experiments/[id].tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useRouter } from 'next/router'
+import React, { useEffect, useState } from 'react'
+import Container from 'semantic-ui-react/dist/commonjs/elements/Container'
+
+import AnalysesApi from '@/api/AnalysesApi'
+
+import ErrorsBox from '@/components/ErrorsBox'
+import Layout from '@/components/Layout'
+
+import { Analysis } from '@/models'
```
Comment (withinboredom) [inline pages/experiments/[id].tsx:11]: Just curious, but why all the whitespace between imports? It seems random, but maybe I'm missing something?

pages/experiments/[id].tsx
```diff
@@ -0,0 +1,37 @@
...
+  const experimentId = Number(useRouter().query.id)
+  debug(`ExperimentPage#render ${experimentId}`)
+
+  const [error, setError] = useState<Error | null>(null)
+  const [analyses, setAnalyses] = useState<Analysis[] | null>(null)
+
+  useEffect(() => {
+    AnalysesApi.findByExperimentId(experimentId)
+      .then((analyses) => setAnalyses(analyses))
+      .catch(setError)
```
Comment (withinboredom) [inline pages/experiments/[id].tsx:25]: Could be simplified to also handle when the id changes and the error needs to be reset also needs to handle when `experimentId` is `null` or `undefined` and maybe if `NaN` since you attempt to cast to `Number` above. Can easily just go to `/path/to/analsyis/not-a-number` to cause it.

```suggestion
  useEffect(async () => {
    if(experimentId == null) return;
    
    let analysis = error = null;
    try {
      analysis = await AnalysisApi.findExperimentId(experimentId)
    } catch (er) {
      error = er
    } finally {
      setAnalysis(analysis)
      setError(error);
    }
```

or, at least

```suggestion
  useEffect(() => {
    if(experimentId == null) return;
    AnalysesApi.findByExperimentId(experimentId)
      .then((analyses) => {
        setAnalyses(analyses)
        setError(null)
      })
      .catch((er) => {
        setError(er)
        setAnalysis(null)
      })
```

pages/experiments/[id].tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import ErrorsBox from '@/components/ErrorsBox'
+import Layout from '@/components/Layout'
+
+import { Analysis } from '@/models'
+
+const debug = debugFactory('abacus:pages/index.tsx')
+
+export default function ExperimentPage() {
+  const experimentId = Number(useRouter().query.id)
```
Comment (withinboredom) [inline pages/experiments/[id].tsx:16]: I wouldn't cast to a number here, it hides a few things. `Number(null) === 0`, `Number('hi') === NaN`, `Number(012) === 10`, instead do some sanity checks and cast in your effect where you can set an error there.

```suggestion
```

Comment (dhurlburtusa) [reply pages/experiments/[id].tsx:16]: Can use the `qc-to_int` package. It doesn't have the issues `parseInt` or `Number` have. If it can't be parsed and an integer, then the default is returned. This is so that NaN, which in most cases isn't helpful, isn't returned unless that is what you want the default to be.

```
import { toInt } from 'qc-to_int'
```

```suggestion
  const experimentId = toInt(useRouter().query.id, null)
```

It also exports a `toIntOrNull` function.

pages/experiments/[id].tsx
```diff
@@ -0,0 +1,37 @@
...
+  const [analyses, setAnalyses] = useState<Analysis[] | null>(null)
+
+  useEffect(() => {
+    AnalysesApi.findByExperimentId(experimentId)
+      .then((analyses) => setAnalyses(analyses))
+      .catch(setError)
+  }, [experimentId])
+
+  return (
+    <Layout title='Experiment: insert_name_here'>
```
Comment (dhurlburtusa) [inline pages/experiments/[id].tsx:29]: If we need to display the experiment information, then we'll need to fetch it too.

__tests__/api/AnalysesApi.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    it('should return a set of analyses with the expected shape', async () => {
+      const analyses = await AnalysesApi.findByExperimentId(123)
+      expect(analyses).toBeDefined()
+      expect(Array.isArray(analyses)).toBe(true)
+      expect(analyses.length).toBeGreaterThan(0)
+      analyses.forEach((analysis) => {
+        expect(typeof analysis.metricAssignmentId).toBe('number')
+        expect(Object.values(AnalysisStrategy).includes(analysis.analysisStrategy)).toBe(true)
+        expect(typeof analysis.participantStats['total']).toBe('number')
+        expect(typeof analysis.participantStats['not_final']).toBe('number')
```
dhurlburtusa, line 15: These don't require bracket notation. There is an ESLint we can add to require/fix not use the more verbose bracket notation when not needed. (I seems like you are switching language contexts. Not Python here but JavaScript :-))
```suggestion
        expect(typeof analysis.participantStats.total).toBe('number')
        expect(typeof analysis.participantStats.not_final).toBe('number')
```

models/Analysis.test.ts
```diff
@@ -0,0 +1,130 @@
...
+            total: 123,
+            not_final: 50,
+            variation_1: 100,
+            variation_2: 23,
+          },
+          metric_estimates: null,
+          recommendation: null,
+          analysis_datetime: '2020-05-04',
+        })
+        expect(analysis).toEqual({
```
dhurlburtusa, line 20: If we do change to extending DTO, then we should test that we are getting an instance of Analysis. Just need to wrap with new Analysis( ... ).

Same comment for the other tests below.

models/Analysis.ts
```diff
@@ -0,0 +1,136 @@
+import parseISO from 'date-fns/fp/parseISO'
+
+import { ApiData } from '@/api/ApiData'
+
+/**
+ * An analysis recommendation.
+ */
+class Recommendation {
```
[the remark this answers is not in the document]
yanirs: That's an old diff. :slightly_smiling_face: It's correct on `master`. I'll fix it now.

pages/experiments/[id].tsx
Comment (yanirs) [reply pages/experiments/[id].tsx:29]: Yes, we will. Future PR. :slightly_smiling_face:
The idea here is just to show that it works.

__tests__/api/AnalysesApi.test.ts
yanirs: Yeah, I used the quoted version because `participantStats` is more like an open-ended mapping. This makes it more explicit, but I don't mind using the dot notation if you like it better. :slightly_smiling_face:

api/AnalysesApi.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Finds all the available analyses for the given experimentId.
+ *
+ * Note: Be sure to handle any errors that may be thrown.
+ *
+ * @throws UnauthorizedError
+ */
+async function findByExperimentId(experimentId: number): Promise<Analysis[]> {
+  const apiResults = await fetchApi('GET', '/analyses/' + experimentId)
+  return apiResults.analyses.map((apiData: ApiData) => Analysis.fromApiData(apiData))
```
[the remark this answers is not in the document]
yanirs: Yeah, but I was also balancing doing too much in one line/statement. I'll change it for consistency, though.

pages/experiments/[id].tsx
Comment (yanirs) [reply pages/experiments/[id].tsx:11]: Yep, though I think we can group them as third-party imports at the top (lines 1-4) and our own imports below without blank lines. Not fussed.

pages/experiments/[id].tsx
```diff
@@ -0,0 +1,37 @@
...
+import Container from 'semantic-ui-react/dist/commonjs/elements/Container'
+
+import AnalysesApi from '@/api/AnalysesApi'
+
+import ErrorsBox from '@/components/ErrorsBox'
+import Layout from '@/components/Layout'
+
+import { Analysis } from '@/models'
+
+const debug = debugFactory('abacus:pages/index.tsx')
```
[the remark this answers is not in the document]
Comment (yanirs) [reply pages/experiments/[id].tsx:13]: Good catch!

models/Analysis.ts
```diff
@@ -35,12 +35,12 @@ class Recommendation extends DataTransferObject<Recommendation> {
    * @param apiData Raw API data.
    */
   static fromApiData(apiData: ApiData): Recommendation {
-    return new Recommendation({
+    return {
```
yanirs, line 38: For some reason the DTO constructor doesn't work as expected, but dropping the constructor calls fixes it. I tried debugging it, and I suspect it's something to do with the way the app is bundled, as it works fine in tests. Removing the constructor calls for now as this is an unrelated bug that we need to solve in general. I'll open an issue.

yanirs: Added #88

PR #87, opened 2020-05-28
WIP: Add Experiment List with Ant D Table

dhurlburtusa:
This the first of two PRs for comparing different UI libraries' table components.

I've only added basic functionality for now. I plan to add some more but I need to sync up some branches first for some other PRs.

## How has this been tested?

The current unit test was updated.

### DIFF
--- a/components/ExperimentsTable.tsx
+++ b/components/ExperimentsTable.tsx
@@ -16,6 +16,8 @@ interface Props {
   experiments: ExperimentBare[]
 }
 
+// TODO: Finish tests to run through the following function if this route is chosen.
+/* istanbul ignore next */
 const createSortByComparable = (
   name: 'endDatetime' | 'name' | 'ownerLogin' | 'platform' | 'startDatetime' | 'status',
 ) => {
--- a/components/ExperimentsTable.tsx
+++ b/components/ExperimentsTable.tsx
@@ -1,23 +1,19 @@
+import { Table } from 'antd'
 import { format } from 'date-fns'
 import debugFactory from 'debug'
-import { toInt } from 'qc-to_int'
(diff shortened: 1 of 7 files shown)


components/ExperimentsTable.tsx
```diff
@@ -1,118 +1,91 @@
...
+  return (a: ExperimentBare, b: ExperimentBare) => {
+    if (a[name] < b[name]) {
+      return -1
+    }
+    if (a[name] > b[name]) {
+      return 1
+    }
+    return 0
+  }
+}
```
dhurlburtusa, line 31: Not sure if there is better way to do this with TypeScript (still learning). All the data types in ExperimentBare work with `<` and `>`.  Not sure if we want to sort in a case-insensitive manner. Won't make a difference with `name` (snake_case), `platform` (enum), `status`, (enum). May make a difference for `ownerLogin` if can contain upper and lower case.

components/ExperimentsTable.tsx
```diff
@@ -1,118 +1,91 @@
...
+
+const sortByEndDatetime = createSortByComparable('endDatetime')
+const sortByName = createSortByComparable('name')
+const sortByOwnerLogin = createSortByComparable('ownerLogin')
+const sortByPlatform = createSortByComparable('platform')
+const sortByStartDatetime = createSortByComparable('startDatetime')
+const sortByStatus = createSortByComparable('status')
```
dhurlburtusa, line 38: We can also create comparitor functions using some npm packages. I know of `natural-orderby`. We can search for others if desired.

dhurlburtusa:
We've decided to move forward with Material UI and Material Table. So, closing and saying goodbye to Ant D.

PR #90, opened 2020-05-31
Add a Basic Material UI Table Component

### LINKED ISSUE #77: Research replacing ExperimentTable with an existing component
As discussed in https://github.com/Automattic/abacus/pull/74#discussion_r426334593, we should clarify the medium-term requirements around our tables and consider using a third-party solution rather than reinventing the wheel. Clarifying the requirements can happen on P2.

This issue is a blocker to work on any new tables.

dhurlburtusa:
This the first of two PRs for comparing different UI libraries' table components.

I've only added basic configuration for now. The default configuration comes with a search field and sorting.

Resolves #77.

## How has this been tested?

Added unit tests for the new component. The tests could be a little more thorough for the search and pagination. But since a finalized design is not done, I'd prefer to wait we know what should be tested. Opening an issue as a reminder.


components/ExperimentsTable.tsx
```diff
@@ -1,116 +1,43 @@
...
-                  onClick={handlePageStep}
-                  role='button'
-                />
-              </Menu>
-            </Table.HeaderCell>
-          </Table.Row>
-        </Table.Footer>
-      )}
-    </Table>
+    <MaterialTable
```
aaronyan, line 29: Wow, that's clean!

components/ExperimentsTable.tsx
```diff
@@ -1,118 +1,42 @@
 import { format } from 'date-fns'
 import debugFactory from 'debug'
-import { toInt } from 'qc-to_int'
-import React, { MouseEvent, useState } from 'react'
-
-import Menu from 'semantic-ui-react/dist/commonjs/collections/Menu'
```
yanirs, line 6: Do we still need the `semantic-ui-react` dependencies?

components/ExperimentsTable.tsx
```diff
@@ -1,118 +1,42 @@
...
+import MaterialTable from 'material-table'
+import React from 'react'
 
 import { ExperimentBare } from '@/models/index'
 
-const debug = debugFactory('abacus:components/ExperimentsTable.tsx')
+/**
+ * Renders the date in ISO-8601 date-only format.
+ */
+const isoDateRenderer = (date: Date) => <span className='whitespace-no-wrap'>{format(date, 'yyyy-MM-dd')}</span>
```
yanirs, line 11: Maybe add a `title` attribute with the full UTC time since it's not necessarily a round UTC date?

components/ExperimentsTable.tsx
```diff
@@ -1,118 +1,42 @@
...
+      columns={[
+        { title: 'Name', field: 'name' },
+        { title: 'Start', field: 'startDatetime', render: (experiment) => isoDateRenderer(experiment.endDatetime) },
+        { title: 'End', field: 'endDatetime', render: (experiment) => isoDateRenderer(experiment.endDatetime) },
+        { title: 'Status', field: 'status' },
+        { title: 'Platform', field: 'platform' },
+        { title: 'Owner', field: 'ownerLogin' },
+      ]}
+      data={experiments}
+      options={{ showEmptyDataSourceMessage: false, showTitle: false }}
```
yanirs, line 37: Should we make the default page size higher than five for now?

components/ExperimentsTable.tsx
dhurlburtusa: @yanirs We are still using the Container component in the page layout and a couple pages. I can replace them with the MUI equivalent. This PR or a new one?

components/ExperimentsTable.tsx
dhurlburtusa: @yanirs How about 25 initial render? User can then select 25, 50, and 100 from drop down.

components/ExperimentsTable.tsx
yanirs: New one is better, as it's a bit out of scope for this one.

components/ExperimentsTable.tsx
yanirs: Sounds good. :slightly_smiling_face:

PR #91, opened 2020-06-02
Add Initial, Basic, Ugly Experiment Details

dhurlburtusa:
Add Initial, Basic, Ugly Experiment Details.

Adds a new API method to fetch experiment data.

## How has this been tested?

An integration test was added for the new API method. Because the mock API is not returning a status, had to temporarily comment out that expectation in the test.

### DIFF
--- a/pages/experiments/[id].tsx
+++ b/pages/experiments/[id].tsx
@@ -1,41 +1,103 @@
 import debugFactory from 'debug'
+import format from 'date-fns/format'
 import { useRouter } from 'next/router'
 import React, { useEffect, useState } from 'react'
 import Container from 'semantic-ui-react/dist/commonjs/elements/Container'
 import { toIntOrNull } from 'qc-to_int'
 
 import AnalysesApi from '@/api/AnalysesApi'
+import ExperimentsApi from '@/api/ExperimentsApi'
 import ErrorsBox from '@/components/ErrorsBox'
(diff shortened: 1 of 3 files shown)


pages/experiments/[id].tsx
```diff
@@ -1,41 +1,103 @@
...
-import { Analysis } from '@/models'
+import { Analysis, ExperimentFull } from '@/models'
 
 const debug = debugFactory('abacus:pages/experiments/[id].tsx')
 
+function ExperimentDetails(props: { experiment: ExperimentFull }) {
+  const { experiment } = props
+  return (
+    <div>
+      <p>TODO: Determine actual layout.</p>
```
Comment (dhurlburtusa) [inline pages/experiments/[id].tsx:20]: I hope that it is okay to leave in these known to dos here in the code as we iterate over the changes. I've done it in the WIP PRs but that seems to be acceptable there. Please let me know if it would be better to leave these out.

Note: Some of the to dos are definitely for the future but some could be determined and addressed in this PR.

Comment (yanirs) [reply pages/experiments/[id].tsx:20]: Yeah, that's fine. For TODOs that are left over after this PR, can you please add issues so we don't forget about them?

For these specific TODOs:

- Actual layout: We'll worry about it later when we have the design.
- Date format: I think it's worth including time and giving it in UTC with ISO8601 format. Can consider localising it later.
- P2 link text: I think `P2` is fine for this iteration.

Issue #96, opened 2020-06-04
Add analysis warnings and reduce table content


yanirs:
Quoting an internal post:

> It won’t be too hard to compress the table by showing a single analysis strategy. As discussed in the analysis strategy post, the MITT variants are mostly there as a check that dropping crossovers and spammers doesn’t affect the analysis results. Therefore, we can show either the `itt_pure` results or the `mitt_no_spammers_no_crossovers` for each metric assignment if this doesn’t affect the conclusions. If there is a difference in the recommendations between strategies, we’ll simply show a warning for the metric assignment saying that it’s been flagged for manual analysis. In the early stages of Abacus, we’ll take it from there. 🙂
>
> Similarly, `pp_naive` would only be shown if the experiment has exposure events and the recommendation for `pp_naive` matches that of `mitt_no_spammers_no_crossovers` (or `itt_pure`). We can default to showing only `pp_naive` in such cases, as it matches the experimenters’ design more closely than the ITT variants. We can be liberal with our warnings and manual analysis flagging until we gain a better understanding of how everything works in the wild. 🙂
>
> Other warnings that are worth showing:
>
> * If the observed variation split is very different from the requested split.
> * If a metric assignment hasn’t been analysed (this is the case for one of the A/A experiment metrics, since we don’t support negative data yet).
> * The warnings returned from the Python ETL.

This issue depends on #95.

yanirs:
More warnings to consider (will update as I go):
* if the total and variation counts for a metric assignment don't match the latest for the primary metric (might be fine due to different analysis datetimes)

PR #99, opened 2020-06-04
Add analysis summary tables

### LINKED ISSUE #95: Add analysis summary tables
As a first step away from the analysis JSON dump, we should show two tables:

* Summary of the latest analyses (for the last analysis datetime)
* Summary of the latest participant counts by analysis strategy

yanirs:
Resolve #95 by adding analysis summary tables.

## How has this been tested?
* Manual verification with Storybook
* Local manual verification with data from a real experiment
* New tests, which mostly rely on snapshots


components/AnalysisSummary.tsx
```diff
@@ -0,0 +1,81 @@
+import { Analysis, AnalysisStrategy, ExperimentFull } from '@/models'
+import _ from 'lodash'
+import { formatIsoUtcOffset } from '@/utils/date'
+import React from 'react'
+import { Paper, Table, TableBody, TableCell, TableContainer, TableHead, TableRow } from '@material-ui/core'
+
```
dhurlburtusa, line 6: I'd like to have clean, easy to manage imports. The https://www.npmjs.com/package/eslint-plugin-simple-import-sort ESLint plugin does a great job at doing that and it can auto-fix too. I think the default sort order the plugin enforces will be suitable and it is what I've been following in this project. Should take about 5 minutes for me to add plus some time to test and clean up.

components/AnalysisSummary.tsx
```diff
@@ -0,0 +1,81 @@
+import { Analysis, AnalysisStrategy, ExperimentFull } from '@/models'
+import _ from 'lodash'
```
dhurlburtusa, line 2: Because we only use a small fraction of lodash, it might be better to import just what we need.

```suggestion
import filter from 'lodash/filter'
import groupBy from 'lodash/groupBy'
import orderBy from 'lodash/orderBy'
import sortBy from 'lodash/sortBy'
```

There is a babel plugin we can add to do this for us.

There is also Webpack's tree-shacking. I am not sure if Next.js has it on my default.

We can run `npm run analyze` to see what is being pulled into the bundle. If it is there, then we should address it at some point. Could wait until a later optimization phase.

Okay, just finished looking the output of `npm run analyze`. We'll should do something with MUI too. Whole library being bundled.

components/AnalysisSummary.tsx
```diff
@@ -0,0 +1,81 @@
...
+  [AnalysisStrategy.MittNoSpammersNoCrossovers]: 'Without crossovers and spammers',
+  [AnalysisStrategy.PpNaive]: 'Exposed without crossovers and spammers',
+}
+
+export default function AnalysisSummary(props: { analyses: Analysis[]; experiment: ExperimentFull }) {
+  const { analyses, experiment } = props
+  if (analyses.length === 0) {
+    return <h2>No analyses yet.</h2>
+  }
+  const sortedAnalyses = _.orderBy(analyses, ['analysisDatetime'], ['desc'])
```
dhurlburtusa, line 21: We can memoize some of these functions.

components/AnalysisSummary.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const metricAssignmentIdToSortedAnalyses = _.groupBy(sortedAnalyses, 'metricAssignmentId')
+  const primaryMetricAssignmentId = _.filter(experiment.metricAssignments, ['isPrimary', true])[0]
+    .metricAssignmentId as number
+  const latestPrimaryMetricAnalyses = _.sortBy(
+    _.filter(metricAssignmentIdToSortedAnalyses[primaryMetricAssignmentId], [
+      'analysisDatetime',
+      metricAssignmentIdToSortedAnalyses[primaryMetricAssignmentId][0].analysisDatetime,
+    ]),
+    'analysisStrategy',
+  )
```
dhurlburtusa, line 33: I think it would be cleaner if we can move these calculations to a function since half the variable are just intermediaries that aren't used in the JSX below.

`latestAnalysisDatetime`, `latestPrimaryMetricAnalyses`, `sortedVariations` are used in JSX.
`sortedAnalyses`, `metricAssignmentIdToSortedAnalyses`, `primaryMetricAssignmentId` are not used in JSX.

Then the new function would be a good candidate for memoization.

components/AnalysisSummary.tsx
```diff
@@ -0,0 +1,81 @@
...
+                <TableCell key={variation.variationId}>
+                  <code>{variation.name}</code>
+                </TableCell>
+              ))}
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {latestPrimaryMetricAnalyses.map((analysis: Analysis) => (
+              <TableRow key={analysis.analysisStrategy}>
+                <TableCell>{strategyToTitle[analysis.analysisStrategy]}</TableCell>
```
dhurlburtusa, line 65: I haven't seen it run yet, but sometimes it is better to only show the first of a series of repeated values.

components/AnalysisSummary.tsx
```diff
@@ -0,0 +1,81 @@
...
+  // - add not final under metric assignment summary -- confusing under counts since it's metric-specific
+  // - handle edge cases
+  // - add some light tests
+  return (
+    <>
+      <h2>Analysis summary for {formatIsoUtcOffset(latestAnalysisDatetime)}</h2>
+      <p>Found {analyses.length} analysis objects in total.</p>
+
+      <h3>Participant counts</h3>
+      <TableContainer component={Paper}>
```
dhurlburtusa, line 49: What was the reason behind using MUI core table instead of Material Table? Extra features by MT too much? Dynamic set of columns due to variations? Something else?

components/AnalysisSummary.tsx
dhurlburtusa, line 33: May want to checkout https://www.npmjs.com/package/sift if we are going to use a bunch of filtering. May just be overkill at this point.

components/AnalysisSummary.tsx
yanirs: Sounds good!

components/AnalysisSummary.tsx
yanirs: I started off with a plain `table` but I figured we can use the core MUI table to make it prettier. Since it's static we don't need the extra features of Material Table, but it may actually be cleaner now that you mention it. Building the table from the individual components is quite verbose. I'll look into using Material Table here.

components/AnalysisSummary.tsx
yanirs: What do you mean?

components/AnalysisSummary.tsx
yanirs: I went with `lodash` since I saw you added it in another PR. It should be good enough for now, especially as its other functions are useful.

yanirs: Yeah, I think we can consider it. The structure/variables aren't final, it's still WIP. :slightly_smiling_face:

components/AnalysisSummary.tsx
yanirs: What would be the advantage of adding memoisation now? Also, how do you see us adding it to a specific function? I'd rather avoid premature optimisations, so it can come later.

components/AnalysisSummary.tsx
dhurlburtusa: I may be making an assumption, but will there be multiple rows with the same title next to each other? If so, then it might be cleaner to only show the title in the first row until the title changes in subsequent rows.

components/AnalysisSummary.tsx
yanirs: > There is a babel plugin we can add to do this for us.

So we would import `_` and it'll break it up as part of the build? I'm happy to add specific imports, but using `_` is easier when developing, and I like the namespacing.

> We can run `npm run analyze` to see what is being pulled into the bundle. If it is there, then we should address it at some point. Could wait until a later optimization phase.
> 
> Okay, just finished looking the output of `npm run analyze`. We'll should do something with MUI too. Whole library being bundled.

Yeah, let's open a separate issue for that. It'd be good to do once things are more stable in terms of what we have on the pages.

components/AnalysisSummary.tsx
yanirs: Ah, I see. The row titles will be unique because the analyses are uniquely indexed by (`metricAssignmentId`, `analysisStrategy`, `analysisDatetime`), and `latestPrimaryMetricAnalyses` contains the analyses for a single (`metricAssignmentId`, `analysisDatetime`) combination.

components/AnalysisSummary.tsx
dhurlburtusa: The one I know of is https://www.npmjs.com/package/babel-plugin-transform-imports. Just took a peak. May not work with just the `_` namespace.
There are a lot of tools to address the over-bundling issue. It is definitely a pre-optimization. Opening an issue as a reminder.

models/ExperimentFull.ts
```diff
@@ -181,4 +181,11 @@ export class ExperimentFull implements ApiDataSource {
       })),
     }
   }
+
+  /**
+   * Return the primary metric assignment ID for this experiment if one exists.
+   */
+  getPrimaryMetricAssignmentId(): number | undefined {
+    return this.metricAssignments.filter((metricAssignment) => metricAssignment.isPrimary)[0].metricAssignmentId
```
dhurlburtusa, line 189:
```suggestion
    return this.metricAssignments.find((metricAssignment) => metricAssignment.isPrimary)?.metricAssignmentId
```

helpers/fixtures.ts
```diff
@@ -0,0 +1,167 @@
...
+  ExperimentFull,
+  MetricAssignment,
+  MetricBare,
+  Platform,
+  RecommendationReason,
+  Status,
+  Variation,
+} from '@/models'
+
+function createAnalysis(fieldOverrides: Partial<Analysis>) {
```
yanirs, line 14: @dhurlburtusa @withinboredom what do you think about this approach to creating our test and story fixtures? I think this is along the lines of what Rob suggested on a previous PR.

components/AnalysisSummary.tsx
withinboredom: FWIW, it's [not recommended to pull in the entirety of lodash](https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark) unless it really makes sense. That blog post recommends https://github.com/lodash/babel-plugin-lodash to address this so you can just use `_.map()` instead of `map()` but it goes into why you want to import one-by-one instead of everything.

The biggest downside for using a plugin like this is that it can greatly increase build times. IMHO, it's easier to just import one by one (and it's what is done in Calypso fwiw).

helpers/fixtures.ts
withinboredom: Looks :100:

components/AnalysisSummary.tsx
dhurlburtusa: Another solution I used in a previous project was to create a module where different utilities are imported and then default exported in an object. I did this specifically for `lodash` but it could be a general solution.

Below is basically what I did:

```
// utils.ts
import filter from 'lodash/filter'
import map from 'lodash/map'

export default {
  filter,
  map,
}
```

```
import _ from '@/utils'

_.map(...)
_.filter(...)
```
Or
```
import utils from '@/utils'

utils.map(...)
utils.filter(...)
```
So there is no confusion if we import/export from other libraries besides lodash.

BTW, the entire lodash library is a small fraction compared to the entire MUI library.

dhurlburtusa: > The biggest downside for using a plugin like this is that it can greatly increase build times.

I'd be curious how much some of these plugins increase build times. For plugins that simply do optimizations, they can be configured to only run for a production build to help keep development fast(er).

components/AnalysisSummary.tsx
dhurlburtusa: Most optimizations should be held off until the code can be profiled and there is previously committed code that can be compared against.

In some cases, it is obvious that repeated work is being done unnecessarily. (This is such a cases because the calculation is deterministic based on the input parameters.) Whether it is really making a negative impact would need to wait for profiling. In these cases, I weigh the difference in DX between adding the optimization. In this case, it is memoization which has minimal impact to DX. In cases where the optimization is more involved, I'd defer until after profiling.

In more cases than not (or naught?), there is a _perceived_ optimization opportunity that should be held off until more data is gathered (profiling, et al).

So, I agree that premature optimization should be avoided but I think there are some acceptable exceptions.

helpers/fixtures.ts
```diff
@@ -0,0 +1,167 @@
...
+      variation_1: { estimate: 0.12, bottom: 0, top: 10.0 },
+      variation_2: { estimate: -0.12, bottom: -1.123, top: 1.0 },
+    },
+    recommendation: {
+      endExperiment: true,
+      chosenVariationId: 2,
+      reason: RecommendationReason.CiInRope,
+      warnings: [],
+    },
+    analysisDatetime: new Date(2020, 4, 10),
```
dhurlburtusa, line 35: If these are being used for tests, we probably want to ensure this value is consistent across timezones.
```suggestion
    analysisDatetime: new Date(Date.UTC(2020, 4, 10)),
```

Note: Comment applies to other dates too.

components/AnalysisSummary.tsx
withinboredom: > I'd be curious how much some of these plugins increase build times.

I imagine it's largely related to the application size and it's use of the plugin. If it only has to change a few files, it's probably negligible. If it were used in something like Calypso... :shudder: Calypso already takes long enough to build :D

yanirs: > BTW, the entire lodash library is a small fraction compared to the entire MUI library.

:arrow_up: This.

I don't mind using individual imports to resolve this. The babel plugin hasn't been updated in ages. However, let's not get distracted by micro-optimisations and remember our likely user base for the foreseeable future: Repeat visitors on a good connection with populated cache. We can afford larger bundle sizes if they make life easier, especially at this point. :slightly_smiling_face:

components/AnalysisSummary.tsx
yanirs: @dhurlburtusa I'm not sure what you're suggesting in practice, though? :slightly_smiling_face: If you're proposing to wrap some potentially-expensive calls with `useMemo()` I'm happy to do it.

components/AnalysisSummary.tsx
dhurlburtusa: Agreed. We have low-priority #104 to remind us later to optimize bundle if it ever becomes a "real" issue.

components/AnalysisSummary.tsx
dhurlburtusa: This is one of those "acceptable exceptions" to optimize, at least in my eyes, since we know we are repeating the same calculation again each time something causes the component to re-render (assuming the analyses isn't what was changing). But these are very unlikely to show any improvement at the human timescale.

So, I'll leave it up to you whether to memoize some of these calculations.

Without a profiler and spending "gobs" of time, we can only guess the benefit.

components/AnalysisSummary.test.tsx
```diff
@@ -1,185 +1,699 @@
...
+          class="MuiTable-root"
+        >
+          <thead
+            class="MuiTableHead-root"
+          >
+            <tr
+              class="MuiTableRow-root MuiTableRow-head"
+            >
+              <th
+                class="MuiTableCell-root MuiTableCell-head"
```
yanirs, line 46: It's pretty annoying that these classes are applied to every single cell, but I figure that if the contents of the table change, the diff will be clear.

components/AnalysisSummary.test.tsx
```diff
@@ -1,185 +1,699 @@
...
+                280
+              </td>
+            </tr>
+          </tbody>
+        </table>
+      </div>
+    </div>
+  `)
+
+  expect(container.querySelector('.analysis-latest-results')).toMatchInlineSnapshot(`
```
yanirs, line 204: This table will definitely change with #96.

components/AnalysisSummary.test.tsx
```diff
@@ -1,185 +1,699 @@
...
+  `)
 })
 
 test('shows the analyses JSON in debug mode', () => {
   const { container } = render(
-    <AnalysisSummary analyses={analyses} experiment={experiment} metrics={[]} debugMode={true} />,
+    <AnalysisSummary analyses={analyses} experiment={experiment} metrics={metrics} debugMode={true} />,
   )
-  expect(container.querySelector('pre')).not.toBeNull()
+  expect(container.querySelector('pre')).toMatchSnapshot()
```
yanirs, line 698: Inlining the JSON felt like taking it was taking it too far. :slightly_smiling_face:

components/AnalysisSummary.tsx
```diff
@@ -1,6 +1,172 @@
...
+    <>
+      {Object.entries(metricAssignmentIdToLatestAnalyses).map(([metricAssignmentId, metricAnalyses]) => (
+        <div key={metricAssignmentId}>
+          <div>
+            <strong>Metric: </strong>
+            <code>{metricsById[metricAssignmentsById[metricAssignmentId].metricId].name}</code>
+          </div>
+          <div>
+            <strong>Attribution window: </strong>
+            {metricAssignmentsById[metricAssignmentId].attributionWindowSeconds / 3600} hours
```
yanirs, line 114: Will replace this with the component from #106 once it's ready.

components/AnalysisSummary.tsx
```diff
@@ -1,6 +1,172 @@
...
+          </div>
+          <div>
+            <strong>Last analyzed: </strong>
+            {format(metricAnalyses[0].analysisDatetime, 'yyyy-MM-dd')}
```
yanirs, line 118: Will replace with the datetime component from #106 as well.

components/AnalysisSummary.tsx
```diff
@@ -1,6 +1,172 @@
...
+        </TableBody>
+      </Table>
+    </TableContainer>
+  )
+}
+
+/**
+ * Render the latest results for the experiment for each metric assignment.
+ *
+ * Note: This is likely to change a lot as part of https://github.com/Automattic/abacus/issues/96.
```
yanirs, line 88: I need to play with it as part of #96, but this is likely to become a single table with the bottom line for each metric.

components/AnalysisSummary.tsx
```diff
@@ -1,6 +1,172 @@
...
+ * Render a table of participant counts based on the latest metric analyses for the given experiment.
+ */
+function ParticipantCounts({
+  experiment,
+  latestPrimaryMetricAnalyses,
+}: {
+  experiment: ExperimentFull
+  latestPrimaryMetricAnalyses: Analysis[]
+}) {
+  const sortedVariations = _.orderBy(experiment.variations, ['isPrimary', 'name'], ['desc', 'asc'])
```
yanirs, line 52: I don't think there's a point memoising calls like this one, where the arrays we work on are fairly small.

components/AnalysisSummary.tsx
```diff
@@ -1,6 +1,172 @@
...
+              </TableCell>
+            ))}
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {latestPrimaryMetricAnalyses.map((analysis) => (
+            <TableRow key={analysis.analysisStrategy}>
+              <TableCell>{AnalysisStrategyToHuman[analysis.analysisStrategy]}</TableCell>
+              <TableCell>{analysis.participantStats.total}</TableCell>
+              {sortedVariations.map((variation) => (
```
withinboredom, line 72: It would probably make it a bit more readable to deconstruct the variation id here and elsewhere:

```suggestion
              {sortedVariations.map(({variationId}) => (
```

components/AnalysisSummary.tsx
```diff
@@ -1,6 +1,172 @@
...
+}: {
+  experiment: ExperimentFull
+  metrics: MetricBare[]
+  metricAssignmentIdToLatestAnalyses: { [key: number]: Analysis[] }
+}) {
+  const metricsById = useMemo(() => _.zipObject(_.map(metrics, 'metricId'), metrics), [metrics])
+  const metricAssignmentsById = _.zipObject(
+    _.map(experiment.metricAssignments, 'metricAssignmentId') as number[],
+    experiment.metricAssignments,
+  )
```
withinboredom, line 103: Can we `useMemo()` here as well?

components/AnalysisSummary.tsx
```diff
@@ -1,6 +1,172 @@
...
+  return (
+    <>
+      {Object.entries(metricAssignmentIdToLatestAnalyses).map(([metricAssignmentId, metricAnalyses]) => (
```
withinboredom, line 106: I'd suggest sorting this after calling `Object.entries` as the order is not guaranteed to be deterministic which can cause some interesting artifacts during rendering. I'd also suggest using lodash here since it's already imported.

components/AnalysisSummary.tsx
```diff
@@ -1,6 +1,172 @@
...
+        <div key={metricAssignmentId}>
+          <div>
+            <strong>Metric: </strong>
+            <code>{metricsById[metricAssignmentsById[metricAssignmentId].metricId].name}</code>
```
withinboredom, line 110: These long chains are fairly hard to grok. It might be more readable to replace with a function: `getMetricName(metric, id)` or something.

components/AnalysisSummary.tsx
withinboredom: It might be worth putting a `{/* todo: comment */}` in the jsx so it isn't lost to github history.

components/AnalysisSummary.tsx
```diff
@@ -16,16 +182,45 @@ export default function AnalysisSummary({
...
+        <h3>Latest results by metric</h3>
+        <LatestResults
+          experiment={experiment}
+          metrics={metrics}
+          metricAssignmentIdToLatestAnalyses={metricAssignmentIdToLatestAnalyses}
+        />
+      </div>
 
       {debugMode ? <pre>{JSON.stringify(analyses, null, 2)}</pre> : ''}
     </>
```
withinboredom, line 226: This file is pretty massive. Can we split it up into smaller components?

components/AnalysisSummary.test.tsx
```diff
@@ -1,185 +1,699 @@
...
+test('renders the full tables with some analyses', () => {
+  const { container } = render(<AnalysisSummary analyses={analyses} experiment={experiment} metrics={metrics} />)
+
+  expect(container).toHaveTextContent(`Found ${analyses.length} analysis objects in total.`)
+  // In non-debug mode, we shouldn't have a <pre> element with the JSON.
+  expect(container.querySelector('pre')).toBeNull()
+
+  // Table snapshots are somewhat verbose, but they will allow us to see if there are any expected changes to the
+  // numbers in the tables.
+  expect(container.querySelector('.analysis-participant-counts')).toMatchInlineSnapshot(`
```
withinboredom, line 26: Can we just use `toMatchSnapshot()` and let `jest` handle the snapshotting?

yanirs: `jest` also handles inline snapshots: https://jestjs.io/docs/en/snapshot-testing#inline-snapshots

withinboredom: Yes, but it's very manual and tedious. When jest is handling it, I can see the diff in the test run and fix what I need to just the same as here, but it's much easier to update an automated snapshot than tediously getting whitespace correct in an inline snapshot.

components/AnalysisSummary.tsx
yanirs: See https://github.com/Automattic/abacus/pull/99/files#r437946375 &ndash; I'm not sure there's a point doing it for small arrays.

components/AnalysisSummary.tsx
yanirs: ~200 lines doesn't seem massive to me. :slightly_smiling_face: As noted, I don't think it's worth doing at this stage because #96 will significantly change the sub-components in this file, and will include some restructuring.

withinboredom: Sounds good. Massive might have been the wrong word, I just meant that this file seems to be responsible for a lot and could benefit from having it's responsibility reduced.

components/AnalysisSummary.test.tsx
yanirs: I'm not sure what you mean by "when jest is handling it". Jest can automatically update inline snapshots. That's how the snapshot ended up here in the first place. :slightly_smiling_face:

components/AnalysisSummary.tsx
yanirs: Yeah, definitely. I was thinking it'd also be good to unit test some of the analysis transformations, as they're not trivial. Planning on doing it in the next PR, when they're more stable. :slightly_smiling_face:

components/AnalysisSummary.test.tsx
withinboredom: Ohhh ... I somehow totally didn't see "except the snapshot values are written automatically back into the source code." :facepalm:

models/ExperimentFull.ts
```diff
@@ -185,7 +185,7 @@ export class ExperimentFull implements ApiDataSource {
   /**
    * Return the primary metric assignment ID for this experiment if one exists.
    */
-  getPrimaryMetricAssignmentId(): number | undefined {
-    return this.metricAssignments.find((metricAssignment) => metricAssignment.isPrimary)?.metricAssignmentId
+  getPrimaryMetricAssignmentId(): number | null {
+    return this.metricAssignments.find((metricAssignment) => metricAssignment.isPrimary)?.metricAssignmentId || null
```
dhurlburtusa, line 189:
```suggestion
    return this.metricAssignments.find((metricAssignment) => metricAssignment.isPrimary)?.metricAssignmentId ?? null
```
I don't know if `metricAssignmentId` could be `0`, but if it is then we don't want to convert it to `null`.

The `??` is the nullish_coalescing_operator. It's fairly new to JS (supported by TS). See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator for details.

yanirs: It can't be zero, but no harm in using `??`. Thanks! :slightly_smiling_face:

PR #103, opened 2020-06-04
Link to Experiment Details

dhurlburtusa:
Link to experiment details page on click of experiment table row.

## How has this been tested?

Will try to add an e2e test for this.


components/ExperimentsTable.tsx
```diff
@@ -27,6 +28,13 @@ interface Props {
 const ExperimentsTable = (props: Props) => {
   debug('ExperimentsTable#render')
   const { experiments } = props
+  const router = useRouter()
+
+  /* istanbul ignore next; to be handled by an e2e test */
+  const handleRowClick = (event?: React.MouseEvent, rowData?: ExperimentBare) => {
+    /* istanbul ignore next */
```
yanirs, line 35: Looks like this isn't needed since the previous comment covers the function.

I tested it locally so I'll push the change.

dhurlburtusa: Awesome.

e2e/navigation.test.tsx
```diff
@@ -0,0 +1,32 @@
+// Empty export forces file to be a module which TS requires with its current
```
dhurlburtusa, line 1: I am not sure what the best way to organize our E2E tests will be. I am thinking we'll be able to better determine that as we add more tests and begin to recognize a pattern.

e2e/navigation.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  beforeAll(async () => {
+    await page.goto('http://a8c-abacus-local:3000')
+  })
+
+  describe('from home page', () => {
+    // In non-production contexts, we should see the main page immediately.
+    it('should navigate to experiment details page on row click', async () => {
+      // Rendering of the table is dynamic. That is, it is not rendered until the
+      // experiment data loads, hence the reason for the wait.
+      await page.waitForSelector('.MuiTableHead-root')
```
dhurlburtusa, line 17: Would like to have a better, more robust way to select the table. If there are multiple tables or the order of the tables change, then the tests will be brittle with the current implementation. Unfortunately, Material Table doesn't make it very easy to add classes to the different parts of the table. Looks like we have to implement a custom table part component to do so.

e2e/navigation.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+      // Assert the experiment list is rendered.
+      const $tableHead = await page.$('.MuiTableHead-root')
```
dhurlburtusa, line 20: Technically, Material Table renders two MUI Tables. The first one with a table head and table body containing the data. The second table just has a table foot where the pagination controls live. So, the fact that the table exists is a good indication that the table is rendered. I'd still prefer something less brittle and more self-documenting.

Thoughts about improvements?

e2e/navigation.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+      expect($tableHead).toBeDefined()
+
+      // Click an experiment row.
+      const $tableRows = await page.$$('.MuiTableBody-root .MuiTableRow-root')
+      expect($tableRows.length).toBeGreaterThan(0)
+      await Promise.all([page.waitForNavigation(), $tableRows[0].click()])
```
dhurlburtusa, line 26: Even though Abacus is a SPA, a "navigation" is still "fired".

This pattern of `await Promise.all([page.waitForNavigation(), $element.click()])` is recommended in the docs.

e2e/navigation.test.tsx
yanirs: Would it not work if we just wait for the `$tableRows`, as done below?

dhurlburtusa: I am not sure what you mean. Are you saying we can remove this explicit check of table existence in favor of it being implied when the exist of table rows exists? If so, I thought about that. Sometimes I am on the fence about whether being explicit. When a check is explicit, it is not accidentally lost when other code is changed by someone not realizing the implied check. Because this explicit check in this case is not really helping, I guess removing it would be more simple.

yanirs: Yeah, that's what I meant. Up to you if you want to keep the explicit table existence check.

PR #105, opened 2020-06-05
Remove Semantic UI Component Library

dhurlburtusa:
## How has this been tested?

Updated some tests due to UI component changes. Ran `npm run verify`. All tests pass.


styles/main.scss
```diff
@@ -1,5 +1,3 @@
-@import '~semantic-ui-css/semantic.min.css';
```
yanirs, line 1: Do we need to import any CSS for material UI?

dhurlburtusa: @yanirs No, at least not for the default styling -- just font imports. MUI is using something called JSS (a type of CSS-in-JS). I haven't researched it yet. I haven't yet read up on how theming works in MUI. Given the fact that MUI is including classes like `MuiContainer-root`, I am guessing that some styling can be controlled the standard CSS way.

PR #106, opened 2020-06-06
WIP: Add Experiment Detail Cards

dhurlburtusa:


api/MetricsApi.ts
```diff
@@ -1,8 +1,48 @@
-import { MetricBare } from '@/models'
+import { MetricBare, MetricFull } from '@/models'
+import _ from 'lodash'
 
 import { ApiData } from './ApiData'
 import { fetchApi } from './utils'
 
+interface MetricCacheEntry {
```
dhurlburtusa, line 7: We need more than the data available in MetricBare when rendering the experiment details. Although I wanted to avoid the 1 + N request issue, the API provides no other way to fetch the full metric details. So, this simplistic caching mechanism using an simple expiration should help. If we need other caches, we can extract this out and make generic.

api/MetricsApi.ts
```diff
@@ -1,8 +1,48 @@
...
+      const age = Date.now() - cachedAt.getTime()
+      if (age < FIVE_MINUTES_IN_MS) {
+        metric = entry.metric
+      }
+    }
+
+    return metric
+  }
+
+  // invalidate(metricId?: number) {
```
dhurlburtusa, line 31: This will be used by the metrics update screen.

components/AttributionWindow.tsx
```diff
@@ -0,0 +1,28 @@
...
+  [AttributionWindowSeconds.OneWeek]: '1 week',
+  [AttributionWindowSeconds.SixHours]: '6 hours',
+  [AttributionWindowSeconds.ThreeDays]: '3 days',
+  [AttributionWindowSeconds.ThreeWeeks]: '3 weeks',
+  [AttributionWindowSeconds.TwelveHours]: '12 hours',
+  [AttributionWindowSeconds.TwentyFourHours]: '24 hours',
+  [AttributionWindowSeconds.TwoWeeks]: '2 weeks',
+}
+
+interface Props {
```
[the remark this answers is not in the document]
dhurlburtusa: Yes. I guess I missed this one.

components/MetricDetails.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { MetricFull } from '@/models'
+
+/**
+ * Renders the metric details.
+ */
+const MetricDetails = (props: { metric: MetricFull }) => {
+  const { metric } = props
+  return (
+    <div>
+      <table>
```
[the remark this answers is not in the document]
dhurlburtusa: There isn't/wasn't a mockup yet. Just wanted to get the data rendered. Will be using MUI subsequent passes over the code.

styles/main.scss
```diff
@@ -1,6 +1,5 @@
-.errors-box {
-  background: #f8d7da;
-  border: 1px solid #f5c6cb;
-  color: #721c24;
-  padding: 1rem;
-}
+@import 'partials/fonts';
```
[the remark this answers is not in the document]
dhurlburtusa: In my experience, I've found this helpful when working with other teams and when the combined files are more that about 300 lines long.

Otherwise, the CSS tends to become one giant bag of unorganized CSS rules with no regard for the cascade.

The first four partials should be fairly non-volatile. The theme partial should have the most activity.

api/MetricsApi.ts
dhurlburtusa: I think that's all we need to add to MetricBare to make that usable.

I am not sure about the dialog vs page. I just tossed the dialog in so I see that the full metric data is coming back. The MetricDetails component is designed to work either in a dialog or a page. So, easy change once that decision is made.

styles/main.scss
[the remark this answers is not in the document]
yanirs: Sorry, I should have been clearer. I'm not opposed to partials. I was wondering about the need for all the CSS _in_ the partials in this PR, especially given our use of MUI. I put the comment here because it's not specific to a file or rule. :slightly_smiling_face:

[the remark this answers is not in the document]
dhurlburtusa: I am guessing that we'll have 1000 lines or less of CSS because of MUI.
I expect that we'll need to add some more utility classes as we continue building.
There is some stuff in `_theme.scss` (h1, h2, etc, et al) that might be overriding what MUI brings. So, we can probably get rid of some of that.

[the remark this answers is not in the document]
yanirs: Maybe we can bring things in as we use them? I'm also unsure we need or want a lot of the stuff from `_utilities.scss`, especially with all the `!important`. :slightly_smiling_face:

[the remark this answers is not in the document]
dhurlburtusa: I've just been bringing in what I need. Some of these are not used in this PR. I copied this from the form code.

Please let me know if you are unfamiliar with why `!important` is used on utility classes.

I've been using these since before they started becoming populate in Bootstrap, Foundation, Tailwind, etc.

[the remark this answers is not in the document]
dhurlburtusa: I think we should only use the utility classes as one-offs and not to use to style the app. I believe the latter is the philosophy behind tailwind and some other CSS libraries.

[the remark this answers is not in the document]
yanirs: Yeah, that's what I was thinking. It's just hard to see the need for the utilities at this point, assuming our styles and MUI cover most of the cases. 

Another thought is about the names of the classes we do keep: `w-100`, `d-none`, and `mt-0` aren't that descriptive, while `text-center`, and `underline` are. I prefer the latter, so I'd rename `w-100` to `full-width` or `width-100`, `d-none` to `hidden` or `display-none`, and `mt-0` to `margin-top-0`. What do you think?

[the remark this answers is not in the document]
dhurlburtusa: I agree that some of the names are cryptic. I was just using the same ones used by Tailwind and Bootstrap 4 since, for some developers, they are well known.

At the top of the `_utilities.scss` file is the following comment to help.

```
/**
| Utility Class Declarations:
|
| Note: Most of these utility classes are copies from tailwindcss.com or
| Bootstrap 4. This is done so that we don't need to bring in the whole CSS
| library.
|
| **Adding New Utility Classes**
|
| First check Tailwind and Bootstrap 4. Copy the class declaration here following
| sort order on the main CSS property.
============================================================================= */
```
If we want to give them less cryptic names, we can. It could help or hinder. I can see it going either way.

A developer knowing we use some Tailwind or Bootstrap names may assume a few class names and later find out that we picked our own.

I don't know what the best solution is. It may be subjective. In [this library](https://github.com/dhurlburtusa/css-properties) I wrote long time ago, I decided to use the property name followed by the value and kabob case it. There are a few exceptions for commonly used classes.

pages/experiments/[id].tsx
```diff
@@ -1,56 +1,420 @@
...
+            </TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          <TableRow className='align-top'>
+            <TableCell component='th' scope='row' variant='head'>
+              Platform
+            </TableCell>
+            <TableCell>{experiment.platform}</TableCell>
+          </TableRow>
```
Comment (dhurlburtusa) [inline pages/experiments/[id].tsx:104]: I am finding the MUI table components to be quite verbose. I can import them in we shorter names but it won't enforce that the same name is used else where in the code base. I think it would be sloppy to import `TableRow` as `TableRow` in one place and `Tr` in another.

A possible solution would be to create a module, import the components, and export them with the names we prefer to use.

This will also address the other issue of the imports being bloated with `from '@material-ui/core/...'` repeated over and over.

Thoughts?

Note: The overall size of MUI isn't too big so I don't believe we have to worry about preventing Webpack from making good decisions for chunks.

styles/main.scss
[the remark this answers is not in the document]
yanirs: Yeah, these are good points. I guess it will ultimately depend on how the classes are used. If they're used a lot, we essentially get inline styling but with lower readability. I'm not too fussed for now, but would be interested to hear @withinboredom's opinion.

pages/experiments/[id].tsx
Comment (yanirs) [reply pages/experiments/[id].tsx:104]: For static tables specifically, I was thinking it may be worth adding a component that's similar to Material Table but without all the bells and whistles. We'll pass in tabular data and it'll render a table. The general idea would be to add higher-level abstractions where they make sense &ndash; not just tables. What do you think?

styles/main.scss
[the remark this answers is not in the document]
withinboredom: `w-100` et al. is basically the norm. It comes from css `width: 100`. It's also friendly to refactors:

`w-100` to `w-75` vs. `full-width` to `three-fourth-width` or something. The other thing to be conscientious of is that CSS is usually minified, but not obfusticated like Javascript (at least until recently-ish). Meaning the strings you choose are usually sent to production as-is, while your js variables are renamed to `a`, `b`, `c`, etc to shorten the code. It behooves you to use smaller names and the ones here are pretty standard across the industry.

TBH, I would just type `w-` and pick the width I want from code-completion/intellisense. I wouldn't even think of `full-width` or `margin-top-0`. But I've also been using Bootstrap and friends since forever, so I may be biased.

pages/experiments/[id].tsx
Comment (dhurlburtusa) [reply pages/experiments/[id].tsx:104]: I think that is a great idea. I wonder how well it will work with the inline editing. We'll see when we get there.

styles/main.scss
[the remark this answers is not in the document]
dhurlburtusa: @withinboredom Do you have to do anything special to get the code-completion? Does it work in the JSX `classNames` context? Do you have to be careful with false suggestions? Does it only work if you have previously typed `w-something` in the currently, or somewhere in the codebase? IDE specific? (So many questions!)

I'd like to get better code-completion in VS Code but I haven't researched what I can do to improve my DX.

dhurlburtusa:
@yanirs 

> To make this reviewable, it'd help a lot to use this as a reference PR and carve out parts of it to more focused standalone PRs.

Could you help me better understand the process for doing this?

The following is how I am imagining it:

1) Take a look at the (remaining) parts of this PR.
2) Figure out what can be extracted out into a PR that is manageable size with a single focus.
3) Commit that as a PR which may stand on its own or may be in a series of stacked PRs.
4) Repeat steps 1 through 3, consuming parts of this PR until all parts of this PR are accounted for.

yanirs:
> @yanirs
> 
> > To make this reviewable, it'd help a lot to use this as a reference PR and carve out parts of it to more focused standalone PRs.
> 
> Could you help me better understand the process for doing this?
> 
> The following is how I am imagining it:
> 
>     1. Take a look at the (remaining) parts of this PR.
>     2. Figure out what can be extracted out into a PR that is manageable size with a single focus.
>     3. Commit that as a PR which may stand on its own or may be in a series of stacked PRs.
>     4. Repeat steps 1 through 3, consuming parts of this PR until all parts of this PR are accounted for.

@dhurlburtusa Yeah, that's pretty much what I had in mind. There are now 30 files changed in this PR, which is a lot and may become even more with full tests. Can easily extract some of the components, API/model changes, cosmetic changes, and CSS adjustments to their own PRs. We don't want 30 PRs, but maybe 4-5 would work well? That's what I've been trying to do with #99.

I'm happy to have a Slack chat if you want to discuss concrete suggestions quickly.

styles/main.scss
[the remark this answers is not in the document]
withinboredom: I just double-checked and it works 100% fine on this branch. Here's my list of IntelliSense extensions, the first one being the css one:

VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=Zignd.html-css-class-completion
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=christian-kohler.npm-intellisense
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=christian-kohler.path-intellisense
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=VisualStudioExptTeam.vscodeintellicode

components/MetricDetails.test.tsx
```diff
@@ -0,0 +1,67 @@
+import { render } from '@testing-library/react'
```
dhurlburtusa, line 1: @dhurlburtusa Remove this from this PR. See if @jessie-rose can use it.

components/MetricDetails.tsx
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
```
dhurlburtusa, line 1: @dhurlburtusa Remove this from this PR. See if @jessie-ross can use it.

models/MetricFull.ts
```diff
@@ -24,7 +25,7 @@ export class MetricFull extends MetricBare {
   /**
    * Constructs a new metric.
    */
-  constructor(data: Readonly<MetricFull>) {
+  constructor(data: ExcludeMethods<MetricFull>) {
```
dhurlburtusa, line 28: @dhurlburtusa Fix incorrect merge.

styles/partials/_elements.scss
```diff
@@ -0,0 +1,103 @@
+/**
```
dhurlburtusa, line 1: @dhurlburtusa Should be able to remove all but the main SCSS files.

utils/currency.ts
```diff
@@ -0,0 +1,10 @@
+const usCurrencyDollarFormatter = new Intl.NumberFormat('us', { style: 'currency', currency: 'USD' })
```
dhurlburtusa, line 1: @dhurlburtusa This can be removed now. It has been moved to formatters.

utils/formatters.test.ts
```diff
@@ -1,6 +1,16 @@
-import { formatIsoUtcOffset, formatUsCurrencyDollar } from './formatters'
+import { formatBoolean, formatIsoUtcOffset, formatUsCurrencyDollar } from './formatters'
```
dhurlburtusa, line 1: @dhurlburtusa This is a good candidate for a small PR.

components/ErrorsBox.tsx
```diff
@@ -14,14 +14,10 @@ const useStyles = makeStyles((theme: Theme) =>
   }),
 )
 
-interface Props {
```
dhurlburtusa, line 17: This can be moved to a small "cleanup" PR.

components/ExperimentsTable.test.tsx
```diff
@@ -2,9 +2,7 @@ import { fireEvent, getAllByText, getByText, getByTitle, render } from '@testing
 import addToDate from 'date-fns/add'
 import React from 'react'
 
-import { ExperimentBare } from '@/models/ExperimentBare'
```
dhurlburtusa, line 5: This can be moved to a small "cleanup" PR.

components/RenderErrorView.tsx
```diff
@@ -2,11 +2,7 @@ import React from 'react'
 
 import type { RenderError } from '@/components/RenderErrorBoundary'
 
-type Props = {
```
dhurlburtusa, line 5: This can be moved to a small "cleanup" PR.

pages/experiments/[id].tsx
Comment (dhurlburtusa) [reply pages/experiments/[id].tsx:108]: Created a LabelValuePanel to address this. See #135.

.eslintrc.js
```diff
@@ -92,6 +92,7 @@ module.exports = {
         specialLink: ['hrefLeft', 'hrefRight'],
       },
     ],
+    'jsx-a11y/no-autofocus': 'off',
```
yanirs, line 95: Why is this turned off now?

components/ExperimentsTable.test.tsx
yanirs: In any case, it doesn't belong in this PR, which doesn't even touch the `ExperimentsTable`. Let's keep PRs focused.

models/Event.ts
```diff
@@ -12,7 +12,7 @@ export class Event {
   /**
    * Event properties that further constrain this event selection.
    */
-  public readonly props?: { [key: string]: string }
+  public readonly props?: { [key: string]: unknown }
```
yanirs, line 15: I thought we agreed not to change this? `props` _is_ a mapping from string to string.

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+  SegmentAssignment,
+  SegmentType,
+  Variation,
+} from '@/models'
+import { formatBoolean, formatUsCurrencyDollar } from '@/utils/formatters'
 
 const debug = debugFactory('abacus:pages/experiments/[id].tsx')
 
-function ExperimentDetails({ experiment }: { experiment: ExperimentFull }) {
+interface MetricAssignmentsRowData {
```
Comment (yanirs) [inline pages/experiments/[id].tsx:43]: Do we really need another interface and `toMetricAssignmentsRowData()` as a standalone function? It's only called once, and it'd be cleaner to pass the `experiment` and `metrics` to `MetricAssignmentsPanel` and inline it, without adding intermediate types.

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+}
+
+function toMetricAssignmentsRowData(metricAssignments: MetricAssignment[], metrics: MetricBare[]) {
+  debug('toMetricAssignmentsRowData', metricAssignments, metrics)
+  const metricAssignmentsRowData: MetricAssignmentsRowData[] = metricAssignments.map((metricAssignment) => ({
+    attributionWindowSeconds: metricAssignment.attributionWindowSeconds,
+    changeExpected: metricAssignment.changeExpected,
+    isPrimary: metricAssignment.isPrimary,
+    metric: metrics.find((metric) => metric.metricId === metricAssignment.metricId),
+    metricAssignmentId: metricAssignment.metricAssignmentId || -1, // Forces to a number to make TS happy.
```
Comment (yanirs) [inline pages/experiments/[id].tsx:59]: Use `as number`?

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+    minDifference: metricAssignment.minDifference,
+  }))
+
+  return metricAssignmentsRowData
+}
+
+function toSegmentsByType(segmentAssignments: SegmentAssignment[], segments: Segment[]) {
```
Comment (yanirs) [inline pages/experiments/[id].tsx:66]: Maybe move to a utility/helper file and test it?

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+      segments.push(segment)
+      segmentsByType[type] = segments
+    } else {
+      console.error(`Unable to lookup segment with ID ${segmentAssignment.segmentId}.`)
+    }
+  })
+  return segmentsByType
+}
+
+function LabelValuePanel(props: {
```
Comment (yanirs) [inline pages/experiments/[id].tsx:85]: As a more generic component, this can be pulled out and tested.

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+              <TableCell padding={padding}>{value}</TableCell>
+            </TableRow>
+          ))}
+        </TableBody>
+      </Table>
+    </Paper>
+  )
+}
+
+function AudiencePanel(props: { experiment: ExperimentFull; segments: Segment[] }) {
```
Comment (yanirs) [inline pages/experiments/[id].tsx:117]: The audience panel and its dependencies can be a standalone PR.

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+  )
+  const data = [
+    { label: 'Description for ending experiment', value: experiment.endReason },
+    { label: 'Conclusion URL', value: experiment.conclusionUrl },
+    { label: 'Deployed variation', value: deployedVariation?.name },
+  ]
+  return <LabelValuePanel data={data} title='Conclusions' />
+}
+
+function GeneralPanel(props: { experiment: ExperimentFull }) {
```
Comment (yanirs) [inline pages/experiments/[id].tsx:163]: `GeneralPanel` and `ConclusionsPanel` can be a standalone PR.

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+          <DatetimeText datetime={experiment.endDatetime} excludeTime />
+        </>
+      ),
+    },
+    { label: 'Owner', value: experiment.ownerLogin },
+  ]
+  return <LabelValuePanel data={data} title='General' />
+}
+
+function MetricAssignmentsPanel(props: { metricAssignmentsRowData: MetricAssignmentsRowData[] }) {
```
Comment (yanirs) [inline pages/experiments/[id].tsx:189]: Let's make `MetricAssignmentsPanel` a standalone PR as well.

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+    background: '#ff980014',
+    color: '#ff9800',
+  },
+  staging: {
+    background: '#82828214',
+    color: '#828282',
+  },
+})
+
+function ExperimentStatus({ experiment }: { experiment: ExperimentFull }) {
```
Comment (yanirs) [inline pages/experiments/[id].tsx:361]: This isn't the only place where we display and style experiment status. Let's make it a standalone PR, move the component to its own file, test it, and use it in `ExperimentsTable`.

utils/formatters.test.ts
yanirs: Let's move it then. :slightly_smiling_face:

pages/experiments/[id].tsx
```diff
@@ -1,66 +1,432 @@
...
+          <TableRow key={segment.segmentId}>
+            <TableCell>{segment.name}</TableCell>
+          </TableRow>
+        ))}
+      </TableBody>
+    </Table>
+  )
+}
+
+function VariationsTable(props: { variations: Variation[] }) {
```
Comment (yanirs) [inline pages/experiments/[id].tsx:309]: This may also potentially be a good standalone PR.

.eslintrc.js
dhurlburtusa: It was because I wanted to auto focus the primary button in a dialog for a better UX. But that dialog is going away. Next time I want to auto focus an element, I check with the team whether to disable this rule globally, locally, or not auto-focus.

models/Event.ts
dhurlburtusa: Reverting.

pages/experiments/[id].tsx
Comment (dhurlburtusa) [reply pages/experiments/[id].tsx:55]: Works too. Thanks.

components/ExperimentStatus.test.tsx
```diff
@@ -0,0 +1,62 @@
+import { render } from '@testing-library/react'
```
dhurlburtusa, line 1: This and `components/ExperimentStatus.tsx` will go away after #144 is merged.

components/LabelValuePanel.test.tsx
```diff
@@ -0,0 +1,83 @@
+import { render } from '@testing-library/react'
```
dhurlburtusa, line 1: This and `components/LabelValuePanel.tsx` will go away after #145 is merged.

.eslintrc.js
yanirs: OK, I'm personally not a fan of auto-focus. As it says in [the rule description](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-autofocus.md):

> Autofocusing elements can cause usability issues for sighted and non-sighted users, alike.

We can discuss it further if auto-focusing is required in the future.

pages/experiments/[id].tsx
Comment (dhurlburtusa) [reply pages/experiments/[id].tsx:31]: I'll get to these soon.

pages/experiments/[id].tsx
Comment (dhurlburtusa) [reply pages/experiments/[id].tsx:81]: I'll get to this soon.

pages/experiments/[id].tsx
Comment (dhurlburtusa) [reply pages/experiments/[id].tsx]: I'll get to this soon.

helpers/fixtures.ts
```diff
@@ -16,6 +16,8 @@ import {
   Platform,
```
dhurlburtusa, line 16: This file will go away after #146 is merged.

models/ExperimentFull.test.ts
```diff
@@ -1,8 +1,7 @@
 import Fixtures from '@/helpers/fixtures'
```
dhurlburtusa, line 1: This file will go away after #146 is merged.

models/ExperimentFull.ts
```diff
@@ -3,7 +3,16 @@ import { ApiDataSource } from '@/api/ApiDataSource'
 import { ExcludeMethods } from '@/types/ExcludeMethods'
```
dhurlburtusa, line 3: This file will go away after #146 is merged.

types/ExcludeMethods.ts
```diff
@@ -1,4 +1,4 @@
 /**
```
dhurlburtusa, line 1: This file will go away after #157 is merged.

dhurlburtusa:
Closing. Superseded by #160, #166, #167, and #168.

Issue #108, opened 2020-06-08
Use collectCoverageFrom


yanirs:
Our calls to `jest` with `--coverage` don't fail for untested files. We should use `--collectCoverageFrom` to specify the directories and files we expect to be covered with each test command.

See: https://jestjs.io/docs/en/configuration#collectcoveragefrom-array

PR #109, opened 2020-06-08
Add result page and update navigation and layout

yanirs:
Another part of addressing #95: 
* Add `AnalysisSummary` as a standalone component (more exciting content to come in #99)
* Split the experiment details page to `/experiments/<id>` and `/experiments/<id>/results`
* Move some common code to the `Layout` component.

## How has this been tested?
* Added some tests
* Manually clicked through the changed pages to verify they're not broken:
  * Homepage: https://abacus-git-add-result-page.a8c.now.sh/
  * Detail page: https://abacus-git-add-result-page.a8c.now.sh/experiments/112
  * Result page: https://abacus-git-add-result-page.a8c.now.sh/experiments/112/results
  * Result page in debug mode: https://abacus-git-add-result-page.a8c.now.sh/experiments/112/results?debug=true (intentionally not linked from anywhere)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
     executor:
       name: node/default
       tag: '10.18.1'
-      size: large
+    resource_class: large
     steps:
       - checkout
       - node/with-cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,6 @@ jobs:
     executor:
       name: node/default
       tag: '10.18.1'
-    resource_class: large
     steps:
       - checkout
       - node/with-cache:
@@ -39,7 +38,7 @@ jobs:
                 command: npm run lint
             - run:
                 name: Unit Testing
-                command: npm run test:unit -- --ci --coverage
+                command: npm run test:unit -- --ci --coverage --maxWorkers 2
             - run:
                 name: Integration Testing
                 command: npm run test:integration -- --ci
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,7 @@ jobs:
     executor:
       name: node/default
       tag: '10.18.1'
+      size: large
     steps:
       - checkout
       - node/with-cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,10 +41,10 @@ jobs:
                 command: npm run test:unit -- --ci --coverage --maxWorkers 2
             - run:
                 name: Integration Testing
-                command: npm run test:integration -- --ci --maxWorkers 2
+                command: npm run test:integration -- --ci
             - run:
                 name: End-to-End Testing
-                command: npm run test:e2e -- --ci --maxWorkers 2
+                command: npm run test:e2e -- --ci
 
 workflows:
   build-and-verify:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,10 +41,10 @@ jobs:
                 command: npm run test:unit -- --ci --coverage --maxWorkers 2
             - run:
                 name: Integration Testing
-                command: npm run test:integration -- --ci
+                command: npm run test:integration -- --ci --maxWorkers 2
             - run:
                 name: End-to-End Testing
-                command: npm run test:e2e -- --ci
+                command: npm run test:e2e -- --ci --maxWorkers 2
 
 workflows:
   build-and-verify:
--- a/components/AnalysisSummary.test.tsx
+++ b/components/AnalysisSummary.test.tsx
@@ -182,18 +182,3 @@ test('shows the analyses JSON in debug mode', () => {
   )
   expect(container.querySelector('pre')).not.toBeNull()
 })
-
-// if (analyses.length === 0) {
-//   return <h2>No analyses yet for {experiment.name}.</h2>
-// }
-//
-// return (
-//   <>
-//     <h2>Analysis summary</h2>
-//     <p>
-//       Found {analyses.length} analysis objects in total. There are {metrics.length} metrics in the system.
-//     </p>
-//
-//     {debugMode ? <pre>{JSON.stringify(analyses, null, 2)}</pre> : ''}
-//   </>
-// )
--- a/components/AnalysisSummary.test.tsx
+++ b/components/AnalysisSummary.test.tsx
@@ -1,0 +1,199 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+
+import AnalysisSummary from './AnalysisSummary'
+import {
+  Analysis,
+  AnalysisStrategy,
+  AttributionWindowSeconds,
+  ExperimentFull,
+  MetricAssignment,
+  Platform,
+  RecommendationReason,
+  Status,
+  Variation,
+} from '@/models'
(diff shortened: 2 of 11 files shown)


components/AnalysisSummary.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+  AttributionWindowSeconds,
+  ExperimentFull,
+  MetricAssignment,
+  Platform,
+  RecommendationReason,
+  Status,
+  Variation,
+} from '@/models'
+
+const experiment: ExperimentFull = new ExperimentFull({
```
yanirs, line 17: These constants are a bit of an overkill for the current tests, but they will be useful later.

@dhurlburtusa @withinboredom do you have any thoughts on sharing such test objects between tests and stories?

components/ExperimentTabs.tsx
```diff
@@ -0,0 +1,24 @@
+import Link from 'next/link'
+import React from 'react'
+import { ExperimentFull } from '@/models'
+
+/**
+ * Experiment tab component. Used to switch between experiment details and results.
+ */
+export default function ExperimentTabs({ experiment }: { experiment: ExperimentFull | null }) {
+  // TODO: Render using Material UI tabs and add tests
```
yanirs, line 9: Our coverage checks don't catch this even though this isn't covered by tests. Added #108 to address the general problem.

For now, I don't think it's worth testing since we'll replace it all with MUI components.

components/ExperimentsTable.tsx
```diff
@@ -18,21 +18,20 @@ const isoDateRenderer = (date: Date) => (
...
 
-interface Props {
-  experiments: ExperimentBare[]
-}
-
 /**
  * Renders a table of "bare" experiment information.
  */
-const ExperimentsTable = (props: Props) => {
+const ExperimentsTable = ({ experiments }: { experiments: ExperimentBare[] }) => {
```
yanirs, line 24: I mentioned it on #106, but I think that in such cases this sort of inlining makes it easier to see what's going on without jumping around.

pages/experiments/[id].tsx
```diff
@@ -1,55 +1,55 @@
...
-        </tr>
-        <tr>
-          <td>Status</td>
-          <td>{experiment.status}</td>
-        </tr>
-        <tr>
-          <td>Platform</td>
-          <td>{experiment.platform}</td>
-        </tr>
+        <tbody>
```
Comment (yanirs) [inline pages/experiments/[id].tsx:19]: Just added the `tbody` to avoid warnings about DOM structure validity. Not critical as we're saying goodbye to this table anyway.

pages/experiments/[id].tsx
```diff
@@ -70,30 +69,15 @@ export default function ExperimentPage() {
...
 
   return (
-    <Layout title='Experiment: insert_name_here'>
-      <Container>
-        <h1>Experiment insert_name_here</h1>
-        {fetchError && <ErrorsBox errors={[fetchError]} />}
-        {experiment && <ExperimentDetails experiment={experiment} />}
-        {analyses && (analyses.length === 0 ? <p>No analyses yet.</p> : <pre>{JSON.stringify(analyses, null, 2)}</pre>)}
-        <p>
-          TODO: Fix the flash-of-error-before-data-load. That is, the `ErrorsBox` initially renders because
```
Comment (yanirs) [inline pages/experiments/[id].tsx:93]: This is a general issue so added #110 instead of tracking it in a TODO.

components/AnalysisSummary.test.tsx
dhurlburtusa: We can setup some test fixtures. I don't recall reading too much about them in the JavaScript work but they are used in Ruby on Rails.

A quick search dug up: https://www.npmjs.com/package/jest-fixtures and https://dev.to/davidimoore/using-fixtures-for-testing-a-reactredux-app-with-jest--enzyme-3hd0. (I haven't dug into them yet.)

pages/experiments/[id]/results.tsx
```diff
@@ -0,0 +1,64 @@
...
+import ExperimentsApi from '@/api/ExperimentsApi'
+import Layout from '@/components/Layout'
+import { Analysis, ExperimentFull, MetricBare } from '@/models'
+import AnalysisSummary from '@/components/AnalysisSummary'
+import MetricsApi from '@/api/MetricsApi'
+import ExperimentTabs from '@/components/ExperimentTabs'
+
+const debug = debugFactory('abacus:pages/experiments/[id]/results.tsx')
+
+export default function ResultPage() {
```
Comment (dhurlburtusa) [inline pages/experiments/[id]/results.tsx:16]:
```suggestion
export default function ResultsPage() {
```

components/AnalysisSummary.test.tsx
yanirs: Yeah, I was thinking that simply exporting pre-defined instances would be a good start. We can also add some simple factory methods with dummy defaults. In this case, I want to use the same analysis list in the stories and tests, so maybe I'll move it as part of #95 and we can take it from there.

.circleci/config.yml
```diff
@@ -38,13 +38,13 @@ jobs:
                 command: npm run lint
             - run:
                 name: Unit Testing
-                command: npm run test:unit -- --ci --coverage
+                command: npm run test:unit -- --ci --coverage --maxWorkers 2
```
yanirs, line 41: Added to deal with the memory issues. I think that Jest doesn't correctly detect the number of CPUs, so too many jobs end up running parallel.

components/ExperimentsTable.tsx
dhurlburtusa: I agree. I have started moving them as I come across.

components/ExperimentTabs.tsx
```diff
@@ -0,0 +1,24 @@
...
+  if (!experiment) {
+    return <></>
```
dhurlburtusa, line 11: I've never tried an empty fragment. I usually return `null`. Do you know of any difference in the resulting code?

components/Layout.tsx
```diff
@@ -3,17 +3,13 @@ import Link from 'next/link'
 import Head from 'next/head'
 import React, { ReactNode } from 'react'
 
-import { onRenderError } from '@/event-handlers/index'
+import { onRenderError } from '@/event-handlers'
 
 import RenderErrorBoundary from './RenderErrorBoundary'
 import RenderErrorView from './RenderErrorView'
+import ErrorsBox from '@/components/ErrorsBox'
```
dhurlburtusa, line 10: Mind if I add the simple-import plugin sooner than later so that there is less to clean up? It'll take about 5 to 10 minutes to add plus some time to clean up and test.

components/ExperimentTabs.tsx
yanirs: Doesn't seem like there's a difference in what's rendered. I didn't give it much thought, but I guess it mostly affects the return type inferred by TypeScript. I don't mind either way.

components/Layout.tsx
yanirs: Go for it. :slightly_smiling_face:

components/AnalysisSummary.test.tsx
dhurlburtusa: Sounds good.

PR #111, opened 2020-06-09
Adjust to Use Client-side Routing

dhurlburtusa:
The dynamic URL require extra prop to use client-side routing.

## How has this been tested?

No page network requests when switching tabs -- just API requests.


dhurlburtusa:
Hopefully Next.js will improve it so we don't have to redeclare the URL pattern each time.

Issue #64, opened 2020-05-07
Test full OAuth flow in e2e tests


yanirs:
With #63, we temporarily disabled authentication on non-production servers. Ideally, we should test the full authentication flow in e2e tests. We just need to figure out how and whether we want to do it with a WP.com user.

yanirs:
Closing as it seems unlikely we'd address it anytime soon.

Issue #52, opened 2020-05-03
Add E2E Testing of Auth Flow in Development/Vercel


dhurlburtusa:
The auth flow is in place. To authenticate, an A12n logins in with their WP.com credentials. Then they accept/denies the request for OAuth from Abacus.

But how do we test this with our E2E tests? We don't want to save credentials in the repo.

Note: If there are different user types/roles, then we should test the difference between them. That is, if Abacus should behave differently based on the user's type/role, then that should be tested.

yanirs:
Closing since it's unlikely we'll get to this anytime soon, if at all.

PR #113, opened 2020-06-09
Update Jest Coverage Configuration

### LINKED ISSUE #108: Use collectCoverageFrom

dhurlburtusa:
Update Jest coverage configuration to catch new files. Added coverage reporting to integration and e2e tests.

New tests added to complete unit test coverage.

Ignored some files for integration and e2e to get coverage at 100% for the files being tested.

Resolves #108 

## How has this been tested?

Ran `npm run verify`. Can still new coverage reports for integration and e2e tests. All three test types have 100% coverage due to some files being ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,23 +111,37 @@ This will open a browser window of the Storybook webapp.
 
 Runs the unit tests with Jest.
 
-The unit tests are found throughout the project excluding tests in the `__tests__` directory which is being reserved for integration tests.
+The unit tests are found throughout the project excluding tests in the `__tests__` directory which is being reserved for integration tests and the `e2e` directory which is being reserved for e2e tests.
 
 ```sh
 npm run test:unit
 npm run test:unit -- --coverage
 npm run test:unit -- --watch
+npm run test:unit -- --coverage --watch
+```
+
+**test:integration**
+
+Runs the integration tests with Jest. These are in the `__tests__` directory.
+
+```sh
+npm run test:integration
+npm run test:integration -- --coverage
+npm run test:integration -- --watch
+npm run test:integration -- --coverage --watch
 ```
 
 **test:e2e**
 
-Runs the end-to-end tests with Jest and Puppeteer in a headless Chrome browser.
+Runs the end-to-end tests with Jest and Puppeteer in a headless Chrome browser. These are in the `e2e` directory.
 
 ```sh
 # Run all the E2E tests
 npm run test:e2e
+npm run test:e2e -- --coverage
 # Run just the E2E smoke tests (intended to be used in development for quick iterations)
 npm run test:e2e:smoke
+npm run test:e2e:smoke -- --coverage
 ```
 
 For debugging, you'll likely want to run with a full visual browser. To do that:
--- a/README.md
+++ b/README.md
@@ -115,9 +115,7 @@ The unit tests are found throughout the project excluding tests in the `__tests_
 
 ```sh
 npm run test:unit
-npm run test:unit -- --coverage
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -111,23 +111,37 @@ This will open a browser window of the Storybook webapp.
 
 Runs the unit tests with Jest.
 
-The unit tests are found throughout the project excluding tests in the `__tests__` directory which is being reserved for integration tests.
+The unit tests are found throughout the project excluding tests in the `__tests__` directory which is being reserved for integration tests and the `e2e` directory which is being reserved for e2e tests.
 
 ```sh
 npm run test:unit
 npm run test:unit -- --coverage
```
dhurlburtusa, line 118: Would simplify documentation and configuration if we always run coverage reports.

api/utils.ts
```diff
@@ -16,9 +16,11 @@ const PRODUCTION_API_URL_ROOT = 'https://public-api.wordpress.com/wpcom/v2/exper
  * @throws UnauthorizedError
  */
 async function fetchApi(method: string, path: string, body: ApiDataSource | null = null) {
+  /* istanbul ignore next; code branch not reachable in integration tests -- we don't hit production */
```
dhurlburtusa, line 19: I think we can avoid these two `istanbul ignores` if the API_URL_ROOT is determined at build time.

jest.integration.config.js
```diff
@@ -1,4 +1,18 @@
 module.exports = {
+  collectCoverageFrom: [
+    '<rootDir>/api/**/*.ts',
+    '!**/node_modules/**',
+    // FIXME: TODO: Get tests on the following.
+    '!<rootDir>/api/UnauthorizedError.ts',
```
dhurlburtusa, line 6: This is only thrown in the branch of code that only runs on production. We don't plan to test production with integration tests. So, `UnauthorizedError.ts` may be permanently ignored. If so, we can add a comment and remove the `FIXME` and `TODO` comment. Thoughts?

jest.e2e.config.js
```diff
@@ -1,4 +1,23 @@
 module.exports = {
+  collectCoverageFrom: [
+    '<rootDir>/pages/**/*.{ts,tsx}',
+    '!**/node_modules/**',
+    // FIXME: TODO: Get tests on the following.
+    '!<rootDir>/pages/_app.tsx',
+    '!<rootDir>/pages/auth.tsx',
+    '!<rootDir>/pages/index.tsx',
+    '!<rootDir>/pages/experiments/[id].tsx',
+    '!<rootDir>/pages/experiments/[id]/results.tsx',
```
dhurlburtusa, line 10: Should we address these now or the final (or is it Week 0?) week of the milestone?

README.md
yanirs: Yeah, I think it makes sense to always run coverage reports.

jest.e2e.config.js
yanirs: Let's add an issue to address these when the pages are more stable.

jest.integration.config.js
yanirs: Can note that this is related to #52 and #64. We should be able to cover it if we test against the production API or a more sophisticated mock.

jest.e2e.config.js
dhurlburtusa: Done.

PR #115, opened 2020-06-10
Add MetricsApi.findById

dhurlburtusa:
Add API update.

## How has this been tested?

New tests added for new code. Ran `npm run verify`. All looks good.


dhurlburtusa:
@withinboredom I should have clarified in the intro that this code is from the #106. That PR currently has 34 files and is considered too large. I was ask by @yanirs to break it up into smaller PRs, 4 or 5. So, I decided to combine a bunch of misc changes together. Otherwise this would be about 4 or 5 PRs with 1 to 3 files each. It seemed like a lot of administrative overhead breaking it up into multiple PRs.

Out of curiosity, how do you see breaking this into smaller, focused PRs benefits the team? I know one benefit is that the Git logs might correspond more closely to what's in the change.

jest.integration.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   collectCoverageFrom: [
     '<rootDir>/api/**/*.ts',
     '!**/node_modules/**',
+    '!<rootDir>/api/index.ts', // Note: `ignore istanbul file` does not work.
```
yanirs, line 5: Can we change the way we import rather than exclude this file?

utils/currency.ts
```diff
@@ -0,0 +1,10 @@
+const usCurrencyDollarFormatter = new Intl.NumberFormat('us', { style: 'currency', currency: 'USD' })
+
+/**
+ * Formats the number as US dollar.
+ */
+function formatUsCurrencyDollar(value: number) {
```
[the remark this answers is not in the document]
yanirs: Yeah, I'm not a fan of tiny files. Might be better to add a general formatting file, as we're not going to have any other currencies in the foreseeable future.

yanirs:
> @withinboredom I should have clarified in the intro that this code is from the #106. That PR currently has 34 files and is considered too large. I was ask by @yanirs to break it up into smaller PRs, 4 or 5. So, I decided to combine a bunch of misc changes together. Otherwise this would be about 4 or 5 PRs with 1 to 3 files each. It seemed like a lot of administrative overhead breaking it up into multiple PRs.
> 
> Out of curiosity, how do you see breaking this into smaller, focused PRs benefits the team? I know one benefit is that the Git logs might correspond more closely to what's in the change.

Not going to speak for @withinboredom, but just to clarify, I meant 4-5 as an example of a number that's greater than 1 and smaller than 34. I should have been clearer that it's just an example &ndash; the general idea is to keep PRs focused.

As far as benefits go, a clear Git history is one, yes. The main other benefits I see are around "reviewability". Just like code readability, it's somewhat subjective. However, a big PR that includes many loosely-connected changes may be blocked on specific issues. The first iteration of this PR is a good example: The metric caching discussion is blocking non-controversial changes because everything is lumped together. Breaking it up means that we can ship the other parts faster, and potentially split up the review load across the team.

components/AnalysisSummary.tsx
```diff
@@ -127,7 +127,7 @@ function LatestResults({
           <div>
             <strong>Last analyzed: </strong>
             {/* TODO: use component to handle formatting */}
-            {format(latestAnalyses[0].analysisDatetime, 'yyyy-MM-dd')}
+            {formatIsoUtcOffset(latestAnalyses[0].analysisDatetime)}
```
dhurlburtusa, line 130: Just a temporary fix until the `DatetimeText` component is available.

api/MetricsApi.ts
```diff
@@ -1,8 +1,50 @@
...
+  //     this.cacheById = {}
+  //   }
+  // }
+
+  put(metric: MetricFull) {
+    this.cacheById[metric.metricId] = { cachedAt: new Date(), metric }
+  }
+}
+
+const cache = new MetricsCache()
```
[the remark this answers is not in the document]
yanirs: I agree. I also don't see what problem this solves, especially as caching is only for five minutes. 

From looking at #106, the cached full metrics are used to render the dialog that's shown when a user clicks on a metric assignment. Given the rarity of such clicks, I think it's fine to make the request on click rather than introduce custom caching complications. In addition to making things less complicated, it'd reduce the request overhead when initially loading the page.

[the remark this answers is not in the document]
dhurlburtusa: The caching was mainly to address the missing `parameterType` in `MetricBare`, but since we'll be adding that information, the caching won't be as useful anymore.

api/MetricsApi.ts
```diff
@@ -14,8 +56,32 @@ async function findAll(): Promise<MetricBare[]> {
...
+ * @throws UnauthorizedError
+ */
+async function findById(metricIds: number[]): Promise<MetricFull[]> {
+  const cachedMetrics = metricIds.map((metricId) => cache.get(metricId)).filter(Boolean) as MetricFull[]
+  const cachedMetricIds = cachedMetrics.map((metric) => metric.metricId).filter(Boolean)
+  const toFetchMetricIds = _.difference(metricIds, cachedMetricIds)
+
+  const metrics = await Promise.all(
+    toFetchMetricIds.map((metricId) =>
+      fetchApi('GET', `/metrics/${metricId}`).then((apiData) => MetricFull.fromApiData(apiData)),
```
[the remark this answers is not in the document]
dhurlburtusa: We'll need to update the API we test with to return an appropriate response in this case. The mock API server returns the same metric regardless. Issue #116 was updated to remind us to improve this.

api/index.ts
```diff
@@ -0,0 +1,6 @@
+import AnalysesApi from './AnalysesApi'
```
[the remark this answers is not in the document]
dhurlburtusa: Has been removed from this PR and will be althogether.

models/Event.ts
```diff
@@ -12,7 +12,7 @@ export class Event {
   /**
    * Event properties that further constrain this event selection.
    */
-  public readonly props?: { [key: string]: string }
+  public readonly props?: { [key: string]: unknown }
```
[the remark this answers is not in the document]
dhurlburtusa: This file is no longer in this PR but we still need to determine whether the change is correct or a bug with the API.
@withinboredom Is this something you can help me determine?

utils/currency.ts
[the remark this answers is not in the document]
dhurlburtusa: This has been moved out of this PR. A new `formatters` module has been created to contain our different formatters.

jest.integration.config.js
dhurlburtusa: `api/index.ts` has been removed and this file has been updated. It's done in a different PR now.

api/MetricsApi.ts
```diff
@@ -14,8 +14,20 @@ async function findAll(): Promise<MetricBare[]> {
...
 
+/**
+ * Find the metric by ID.
+ *
+ * Note: Be sure to handle any errors that may be thrown.
+ *
+ * @throws UnauthorizedError
+ */
+async function findById(metricId: number): Promise<MetricFull> {
+  return MetricFull.fromApiData(await fetchApi('GET', `/metrics/${metricId}`))
```
withinboredom, line 25: FWIW, this will fail spectacularly with an undefined/nonexistant metric id. However, that's out of scope for this PR. Maybe this weekend or next week, I'll dip my toes into some TS and see what I come up with :) if someone doesn't beat me to it.

There's two different ways to handle errors:

1. Throw an exception in `fetchApi()` and rely on error boundaries and try-catch-finally
2. Return the api object OR an error object.

I'm partial to first one because I think that errors are exceptional in this case and that it allows us to uniformly handle errors. I'm also super annoyed dealing with functions that may or may not return errors (:sigh: PHP)

api/MetricsApi.ts
```diff
@@ -14,8 +14,20 @@ async function findAll(): Promise<MetricBare[]> {
...
+}
+
 const MetricsApi = {
   findAll,
+  findById,
 }
 
 export default MetricsApi
```
withinboredom, line 33: Any reason to not use:

```suggestion
export default {
  findAll,
  findById,
}
```

dhurlburtusa: The namespacing helps with self-documentation and avoids name conflicts. If I come across just `findAll`, then I need to see which module it is coming from. I just tried it in VSCode and tested what information would be available on hover. If the comment and return types are adequate, then you can figure out the which function it is. But this does not work during a code review.

Also, we are using `findAll` and `findById` in other `*Api.ts` modules that would conflict. There are ways around this conflict with aliases, but I think the namespacing provides the best DX overall.

api/MetricsApi.ts
dhurlburtusa: I can temporarily hook this function up to the production API so I can write the error handling code now since the mock API is very limited now.

I am good with throwing an exception or returning an error object. I think it is best not to handle the error within the call because that couples the UI too tightly to the API code. (Not implying you were suggesting to handle the error here in the API layer.)

The Swagger Yaml says a 404 should be returned. I think throwing an `NotFound` error would be good. The UI that calls this API function can then handle it appropriately.

dhurlburtusa: @withinboredom @yanirs I temporarily hooked up the API code to the production API. A status of 404 is properly returned along with a JSON body matching the Error schema:
```
{
  "code": "metric_not_found",
  "message": "The metric was not found",
  "data": {
    "status": 404
  }
}
```

FYI, here is the list of status codes the API is documented as returning:

200 - Success
201 - Created
400 - BadRequest
401 - Unauthorized
403 - Forbidden
404 - NotFound
409 - Conflict
501 - NotImplemented

We can handle this several ways.
1) Assuming all works as expected, the API always returns a body with JSON regardless of the response status.
2) In `api/utils.ts`, check the response status, parse the body as JSON, if response is anything but 200 or 201, then parse body should be an error object. This can be passed to an ApiError class we create and it can be thrown.
3) Same as #2 but return the ApiError instance.
4) Similar to #2 but have separate error instances for the different error statuses. For example,  a NotFoundError for 404s, UnauthorizedError (this error already exists) for 401s, etc.

Before making a decision on how to handle it in the API layer, it might be useful to think about how the UI layer will handle an API error. Here is what I am imagining:

If the error is a 401, the UI redirects or prompts the user to re-authenticate.
If the error is a 403 or a 404, then give the user an appropriate UI message.
I think the other error states (400, 409, 501, and any other error condition the server may response with) can be handled as such:
  a) Display an appropriate UI message. Can just be generic since the details are usually meaningless to the user.
  b) Console log the error details.

Also, to keep the current code clean and simple I think throwing errors is better.
Currently, we have something like:
```
  return (await fetchApi('GET', '/metrics')).metrics.map((apiData: ApiData) => MetricBare.fromApiData(apiData))
```
This works perfectly in a non-error situation. But if we don't throw an error in `api/utils.ts`, then each one of the functions like the one above will need special, repeated error handling. I think it would be clean if the error trickles up to the UI layer for appropriate handling. Then these functions can concentrate on the success state.

Thoughts? P2 this?

withinboredom: You can also let it bubble up to an error boundary, which imho is probably cleaner and more reusable. When looking at the code, you need to only consider the “happy path”

```jsx
<NotFound show={<>Unable to locate metric</>}>
  <MetricThing forId={42} />
</ NotFound>
```

dhurlburtusa: The error boundary and throwing an `ApiError` that basically wraps the API error information can help centralize the UI error handling logic.

api/MetricsApi.ts
yanirs: +1 to namespacing, it make things clearer. :slightly_smiling_face:

api/MetricsApi.ts
yanirs: +1 to centralising the logic for throwing and handling API errors by type, as suggested. Definitely out of scope for this PR, so can you please open an issue? It feels like a lower priority to me than getting the basic views out &ndash; it's more of a polishing step.

__tests__/api/MetricsApi.test.ts
```diff
@@ -14,4 +15,36 @@ describe('MetricsApi.ts module', () => {
...
+      expect(Array.isArray(metric.eventParams)).toBe(true)
+      metric.eventParams?.forEach((eventParam) => {
+        expect(typeof eventParam.event).toBe('string')
+        expect(typeof eventParam.props).toBe('object')
+        expect(eventParam.props).not.toBe(null)
+      })
+      expect(metric.revenueParams).toBe(null)
+    })
+
+    // TODO: Unskip this once the mock API stops returning the mock metric regardless
```
yanirs, line 38: Can you please add an issue for this so that we don't forget about it?

utils/formatters.ts
```diff
@@ -1,6 +1,6 @@
 import format from 'date-fns/format'
 
-const usCurrencyDollarFormatter = new Intl.NumberFormat('us', { style: 'currency', currency: 'USD' })
+const usCurrencyDollarFormatter = new Intl.NumberFormat('en-US', { style: 'currency', currency: 'USD' })
```
yanirs, line 3: @dhurlburtusa I pulled the PR to merge `master` and test locally, and the new formatter test failed because it formatted the currency as `USD 0.02` for me. My guess that this is because `us` isn't a valid locale, and it falls back to the default system locale. Looks like changing to `en-US` fixes it, so I figured I might as well push it here since I'm about to merge the PR.

__tests__/api/MetricsApi.test.ts
dhurlburtusa: Done. #124.

models/Event.ts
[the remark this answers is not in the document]
dhurlburtusa: Created #125 to track this.

PR #117, opened 2020-06-10
Add Initial Stylesheets

dhurlburtusa:
Add initial set of Stylesheets

## How has this been tested?

Ran `npm run verify`. Stylelint had no complaints.
These are the styles being used in the experiment details.

### DIFF
--- a/styles/partials/_elements.scss
+++ b/styles/partials/_elements.scss
@@ -1,0 +1,103 @@
+/**
+| Element Styles:
+============================================================================= */
+
+/**
+| Border-Box Reset:
+===================================== */
+
+html {
+  box-sizing: border-box;
+}
+
+*,
+*::before,
+*::after {
+  /*
+  Inherit box-sizing to make it easier to change the property for components that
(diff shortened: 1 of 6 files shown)


yanirs:
> I didn't test it locally, but I trust that you did.

That's what Vercel and CircleCI are for. :slightly_smiling_face:

styles/partials/_elements.scss
```diff
@@ -0,0 +1,103 @@
+/**
+| Element Styles:
+============================================================================= */
```
yanirs, line 3: Very minor, but where does this style of commenting come from?

dhurlburtusa: It's something I copied from one of my other projects. The length of `=` makes it so the line is 80 characters long.

It probably isn't too useful for us. Do you have a preference with comment style?

It was something I used before pre-processors like Sass or when our designer insisted on 4000 line long CSS files. The comments made it easier to find the different groups.

yanirs: Not too fussed. Can have a look around to see what others are doing. A quick search yielded [this](https://css-tricks.com/options-programmatically-documenting-css/). In any case, it's very low priority. :slightly_smiling_face:

PR #118, opened 2020-06-11
Add datetime and attribution window formatting

dhurlburtusa:
Add Reusable Components. These were extracted from #106.

This PR was originally about 12 files for the different components but it got larger after merging in new stuff from `master` and making changes from some feedback. I hope it is not too large. Some of the file diffs are really tiny.

## How has this been tested?

New tests. `verify` script passes.


components/OwnerAvatar.tsx
```diff
@@ -0,0 +1,20 @@
...
+ */
+const OwnerAvatar = (props: { ownerLogin: string }) => (
+  <span className='d-inline-block'>
+    <Avatar className='small'>
+      {props.ownerLogin
+        .split('_')
+        .slice(0, 2)
+        .map((parts) => parts[0])
+        .join('')
+        .toUpperCase()}
```
dhurlburtusa, line 15: This just gets the first two initials of the owner login. Based on what I've seen, the owner logins are underscore separated.

models/MetricFull.ts
```diff
@@ -43,6 +43,7 @@ export class MetricFull extends MetricBare {
         ? apiData.event_params.map((rawEvent: ApiData) => Event.fromApiData(rawEvent))
         : null,
       revenueParams: apiData.revenue_params ? MetricRevenueParams.fromApiData(apiData.revenue_params) : null,
+      parameterType: apiData.revenue_params ? 'revenue' : 'conversion',
```
dhurlburtusa, line 46: Will the parameter type be coming back with MetricFull or is it okay that it is calculated here?

utils/formatters.ts
```diff
@@ -0,0 +1,20 @@
+import format from 'date-fns/format'
+
+const usCurrencyDollarFormatter = new Intl.NumberFormat('us', { style: 'currency', currency: 'USD' })
+
+/**
+ * Formats the date in ISO-8601 format with a UTC timezone value plus a `+00:00`
```
dhurlburtusa, line 6: I'll get this changed to show UTC instead of the offset.

.prettierignore
```diff
@@ -1,11 +1,11 @@
 .dockerignore
 .prettierignore
 
+*.snap
```
yanirs, line 4: I don't mind the ordering, but it'd be good to leave such changes out of PRs in the future and/or justify them.

components/AnalysisSummary.tsx
```diff
@@ -127,7 +127,7 @@ function LatestResults({
           <div>
             <strong>Last analyzed: </strong>
             {/* TODO: use component to handle formatting */}
-            {format(latestAnalyses[0].analysisDatetime, 'yyyy-MM-dd')}
+            {formatIsoUtcOffset(latestAnalyses[0].analysisDatetime)}
```
yanirs, line 130: Please use the `DatetimeText` component. In general, as noted internally, the `DatetimeText` component should have been its own PR, as we also want to use it in multiple existing places.

Also, the attribution window TODO above should be resolved by this PR.

components/AttributionWindow.tsx
```diff
@@ -0,0 +1,24 @@
...
+  [AttributionWindowSeconds.OneWeek]: '1 week',
+  [AttributionWindowSeconds.TwoWeeks]: '2 weeks',
+  [AttributionWindowSeconds.ThreeWeeks]: '3 weeks',
+  [AttributionWindowSeconds.FourWeeks]: '4 weeks',
+}
+
+/**
+ * Renders the attribution window in concise, human readable text.
+ */
+const AttributionWindow = (props: { attributionWindowSeconds: AttributionWindowSeconds }) => {
```
yanirs, line 20: Do we need this to be a component? Why not follow the same approach from #99 and just add a static mapping like `AnalysisStrategyToHuman`? This feels like a lot of code and tests to achieve the same goals.

components/BooleanText.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+/**
+ * Renders the boolean value in human readable text.
+ */
+const BooleanText = (props: { value: boolean }) => (
```
yanirs, line 6: Similarly to `AttributionWindow`, I wonder if this would be better served by a formatter function or static mapping.

components/ExperimentsTable.tsx
```diff
@@ -5,7 +5,7 @@ import { useRouter } from 'next/router'
 import React from 'react'
 
 import { ExperimentBare } from '@/models'
-import { formatIsoUtcOffset } from '@/utils/date'
+import { formatIsoUtcOffset } from '@/utils/formatters'
```
yanirs, line 8: We should use `DatetimeText`, not `formatIsoUtcOffset()`.

components/MetricMinimumDifference.test.tsx
```diff
@@ -0,0 +1,32 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+
+import { MetricBare } from '@/models'
+
+import MetricMinimumDifference from './MetricMinimumDifference'
+
+test('renders dollars for conversion metrics', () => {
+  // TODO: Get from fixtures.
```
yanirs, line 9: Please fix these TODOs in this PR.

components/MetricMinimumDifference.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+import { MetricBare } from '@/models'
+import { formatUsCurrencyDollar } from '@/utils/formatters'
+
+/**
+ * Renders the attribution window in concise, human readable text.
+ */
+const MetricMinimumDifference = (props: { metric: MetricBare; minDifference: number }) => {
+  const { metric, minDifference } = props
```
yanirs, line 10: This can be inlined in the function declaration.

components/OwnerAvatar.tsx
yanirs: I might be missing something, but where did you get this requirement from? Owner logins aren't guaranteed to have underscores. Most of them don't.

I think we're better off dropping this component.

models/MetricBare.ts
```diff
@@ -16,6 +16,11 @@ export class MetricBare {
    */
   public readonly description: string
 
+  /**
+   * Type type of parameters of this metric.
```
yanirs, line 20: `Type type` -> `Type`. :slightly_smiling_face:

helpers/fixtures.ts
```diff
@@ -256,6 +256,7 @@ function createMetricsBares(numMetrics = 3) {
         metricId: i,
         name: `metric_${i}`,
         description: `This is metric ${i}`,
+        parameterType: 'revenue',
```
yanirs, line 259: As noted on Slack and on the PR this was extracted from, the `parameterType` change should really be its own standalone PR.

models/MetricFull.ts
yanirs: Not sure. Can add a TODO to address this once we extend the API.

components/DatetimeText.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+import React from 'react'
+
+import DatetimeText from './DatetimeText'
+
+test('renders as ISO 8601 UTC', () => {
+  const input = new Date(Date.UTC(2020, 4, 2))
+  const { getByText, getByTitle } = render(<DatetimeText value={input} />)
+
+  expect(getByText('2020-05-02T00:00:00.000Z')).toBeInTheDocument()
+  // The following test is locale dependent.
```
yanirs, line 11: I'd rather we avoid locale-dependent tests. Ideally, the behaviour of the tests should be independent of the machine where they run.

components/DatetimeText.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+  expect(getByTitle(input.toLocaleString())).toBeInTheDocument()
+})
+
+test('renders as ISO 8601 UTC without time', () => {
+  const input = new Date(Date.UTC(2020, 4, 2))
+  const { getByText, getByTitle } = render(<DatetimeText time={false} value={input} />)
+
+  expect(getByText('2020-05-02')).toBeInTheDocument()
```
yanirs, line 19: Won't this also be true when the `time` parameter is `true`?

components/DatetimeText.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+
+const TIME_PART_RE = /T\d{2}:\d{2}:\d{2}(?:\.\d+(?:Z)?)?/
+
+/**
+ * Renders the date value in ISO 8601 format UTC.
+ */
+const DatetimeText = (props: { value: Date; time?: boolean }) => {
```
yanirs, line 8: Can we please destructure `props` (here and elsewhere) and use more descriptive argument names? Something like `value` -> `datetime` and `time` -> `showTime`.

components/AnalysisSummary.tsx
dhurlburtusa: I tried to use the `DatetimeText` component but ran into the local date issue and how to address if we use snapshots. I tried setting the Timezone but Node didn't change how it formatted the tooltip date. The only way I can think of how to resolve this is to replace the snapshot with individual assertions, but that is a lot of manual work.

components/BooleanText.tsx
dhurlburtusa: Formatter sounds good.

components/DatetimeText.test.tsx
dhurlburtusa: Before I make changes, I just want to make sure I understood the requirements. It was my understanding that the tooltip would be in the user's local time and would always include the time portion of the date regardless of the `time` flag that affects the main body content of the component.

If this is correct, any suggestions how we can make this locale-independent?

components/DatetimeText.test.tsx
dhurlburtusa: Good catch. I like the snapshot idea for small components like this.

components/AttributionWindow.tsx
dhurlburtusa: I'll give that a go.

.prettierignore
dhurlburtusa: Sure. I was keeping the file patterns and the directory patterns in their separate groups.

components/OwnerAvatar.tsx
dhurlburtusa: It was just something to make the implementation look better until more decisions were made. In hind sight, I think I should have asked on Slack or a P2 to get clarification about the mockup.

components/DatetimeText.test.tsx
dhurlburtusa: I tried setting time to true. The test fails. `getByText`, with a string, matches only the entire text of a node. See https://testing-library.com/docs/dom-testing-library/api-queries#bytext and https://testing-library.com/docs/dom-testing-library/api-queries#textmatch for more details. Using a `precision` option with `exact` set to `false` or a regexp without start and end boundaries (`^` and `$`) would let time leak in.

components/ExperimentsTable.tsx
```diff
@@ -38,8 +28,18 @@ const ExperimentsTable = ({ experiments }: { experiments: ExperimentBare[] }) =>
     <MaterialTable
       columns={[
         { title: 'Name', field: 'name' },
-        { title: 'Start', field: 'startDatetime', render: (experiment) => isoDateRenderer(experiment.endDatetime) },
-        { title: 'End', field: 'endDatetime', render: (experiment) => isoDateRenderer(experiment.endDatetime) },
+        {
+          title: 'Start',
+          field: 'startDatetime',
+          // eslint-disable-next-line react/display-name
```
dhurlburtusa, line 34: Instead of ignoring this rule here, maybe we should turn it off altogether. I don't find it very useful myself. Having a display name is useful with the React Dev Tools since it will use the display name. This can make it easier to reason about/orient yourself in the component tree. I use this tool so infrequently, I'd prefer to turn off the rule. For someone using the tool more frequently, maybe they can add missing display names as the come across them.

Thoughts about what's best for the team?

utils/formatters.ts
dhurlburtusa: Actually, going to keep this function around to format the date we send to the API for POSTs, PUTs, etc.

components/MetricMinimumDifference.test.tsx
dhurlburtusa: @yanir I think I'll need to pull this (`components/MetricMinimumDifference.test.tsx`) and `components/MetricMinimumDifference.tsx` out into a PR stacked on the PR (to be created) for `parameterType` since it depends on it.

components/ExperimentsTable.tsx
yanirs: I'm happy to drop the rule if it's not useful.

components/MetricMinimumDifference.test.tsx
yanirs: Makes sense, but now that I'm looking at it more closely, I'm not sure why we need it to be a component. We can add a method `parameterTypeToHuman()` to `Metric*` with implementation `return metric.parameterType === 'revenue' ? formatUsCurrencyDollar(minDifference) : `${minDifference} pp`, or simply inline the call if it only happens once or twice in the codebase.

components/DatetimeText.test.tsx
yanirs: The requirements are correct. I meant we'd make it locale independent by running jest with a consistent locale.

yanirs: Alright, with https://github.com/Automattic/abacus/pull/118/commits/b220fc2aaee427a2a17d63eee79a0b4f3bf16199 test runs should be independent of the machine's timezone and locale, so please switch to snapshots.

components/DatetimeText.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * Renders the date value in ISO 8601 format UTC.
+ */
+const DatetimeText = ({ datetime, excludeTime }: { datetime: Date; excludeTime?: boolean }) => {
+  const datetimeText = datetime.toLocaleString()
+  let text = datetime.toISOString()
+  if (excludeTime) {
+    text = text.replace(TIME_PART_RE, '')
+  }
+  return (
+    <span className='whitespace-no-wrap' title={datetimeText}>
```
yanirs, line 15: Why use a utility class rather than give this a unique class name and specific formatting?

.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ jobs:
                 command: npm run lint
             - run:
                 name: Unit Testing
-                command: npm run test:unit -- --ci --maxWorkers 2
+                command: npm run test:unit -- -- --ci --maxWorkers 2
```
yanirs, line 41: This is an annoying artefact of nested `npm run` calls. See `package.json` for a discussion.

package.json
```diff
@@ -96,6 +97,8 @@
     "docker-start": "docker run -p 8888:3000 abacus:latest",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
+    "@comment jest": "Run jest with consistent TZ and LANG",
+    "jest": "cross-env NODE_ICU_DATA=node_modules/full-icu TZ=EST LANG=en_GB jest --coverage",
```
yanirs, line 101: Is there a better way to share environment variables across scripts? This is a bit clunky, as it sometimes requires multiple `--` to pass in custom flags.

components/DatetimeText.test.tsx
dhurlburtusa: Yep, I knew what you meant. 😀 I agree. I would prefer to avoid locale-dependent tests too. I just wanted to make sure I had the requirement correct before trying to find a solution. No point to finding a solution to an incorrect understanding of the requirement.

Thanks for your help with finding a locale/timezone -independent solution.

package.json
dhurlburtusa: I know with a "pure" node call (e.g., `node my-script`), there is a `--require` CLI option (e.g., `node --require something my-script`) which can be used to set the env vars (or other things) for the node process. Jest has [--env=&lt;environment>](https://jestjs.io/docs/en/cli.html#--envenvironment). It sounds similar.

Note: The `--` is only required with `npm`. `yarn` doesn't require it and actually issues a warning if you use `--`. I never tried nested `yarn` calls, however.

.circleci/config.yml
dhurlburtusa: Ew. 🤢

package.json
dhurlburtusa: I just tried this solution (to clarify: your solution with setting the TZ, etc, not the `--env` in my previous comment) on Windows. I am getting the same results (still formatted in local time with a en-US locale) with my attempt. I assume you are not using Windows.

snapshot-failed-on-windows

Darn you Windows! 😠

dhurlburtusa: I tried explicitly passing the locale (via `process.env.LANG`) and the timezone (via `process.env.TZ`) directly to `Date#toLocaleString`. This is the only way I found to actually affect the output of this function. I did have to make some tweaks to the values: Used `en-GB` for the `LANG` and used `America/New_York` for `TZ`. Otherwise I would get error messages about `en_GB` and `EST`. I'll push what I came up with to see if it works in CI. If it does, hopefully it works on other OSes in different timezones. This solution appears to work in the browser too.

dhurlburtusa: Okay. It works on my system and for CI. @yanirs Do the tests still pass on your system?

.circleci/config.yml
yanirs: Yeah, not great. :disappointed:

yanirs: Good enough for now, though. :slightly_smiling_face: Can open a low priority issue to come up with something better.

components/DatetimeText.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+
+const TIME_PART_RE = /T\d{2}:\d{2}:\d{2}(?:\.\d+(?:Z)?)?/
+
+/**
+ * Renders the date value in ISO 8601 format UTC.
+ */
+const DatetimeText = ({ datetime, excludeTime }: { datetime: Date; excludeTime?: boolean }) => {
+  const datetimeText = datetime.toLocaleString(process.env.LANG, { timeZone: process.env.TZ })
```
yanirs, line 9: It's worth adding a comment saying why we're explicitly using `process.env` variables.

components/DatetimeText.tsx
```diff
@@ -0,0 +1,21 @@
...
+  let text = datetime.toISOString()
+  if (excludeTime) {
+    text = text.replace(TIME_PART_RE, '')
```
yanirs, line 12: Let's use `.substring(0, ISO_DATE_LENGTH)` instead of a regex, with `const ISO_DATE_LENGTH = 10`.

package.json
```diff
@@ -96,6 +97,8 @@
     "docker-start": "docker run -p 8888:3000 abacus:latest",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
+    "@comment jest": "Run jest with consistent TZ and LANG",
+    "jest": "cross-env NODE_ICU_DATA=node_modules/full-icu TZ=America/New_York LANG=en-GB jest --coverage",
```
yanirs, line 101: Setting `America/New_York` may be problematic, as it will change depending on DST. We don't want our tests to break in a few months. :slightly_smiling_face: Is it possible to set it to something that includes the DST information (like EST, which I had before)?

dhurlburtusa: I don't think we have to worry about DST in this situation because we are using specific dates: 2020-05-02T00:00:00 UTC and 2020-05-09T00:00:00 UTC. For these particular dates, they with always be eastern daylight time. It's not until Nov to March will the time have a different offset from UTC. Now if we were using a moving target like `new Date()`, then I think we would have an issue. I could be wrong about this but I've done a lot research about working with JavaScript dates and that is my understanding.

I tried `EST` but for some reason I was getting errors about it not being recognized.

yanirs: Ah yeah, that make sense. :slightly_smiling_face:

package.json
yanirs: Yeah, the tests still pass on my system. Thanks for sorting it out!

components/DatetimeText.tsx
dhurlburtusa: I agree.

components/DatetimeText.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const datetimeText = datetime.toLocaleString(process.env.LANG, { timeZone: process.env.TZ })
+  let text = datetime.toISOString()
+  if (excludeTime) {
+    text = text.substring(0, ISO_DATE_LENGTH)
+  }
+  return (
+    <span className='datetime-text' title={datetimeText}>
+      {text}
+    </span>
+  )
```
yanirs, line 23: I should have noted this before, but I this can be further cleaned up with clearer variable names:

```suggestion
  const localeString = datetime.toLocaleString(process.env.LANG, { timeZone: process.env.TZ })
  const isoString = datetime.toISOString()
  return (
    <span className='datetime-text' title={localeString}>
      {excludeTime ? isoString.substring(0, ISO_DATE_LENGTH) : isoString}
    </span>
  )
```

components/DatetimeText.tsx
```diff
@@ -0,0 +1,26 @@
...
+const ISO_DATE_LENGTH = 10
+
+/**
+ * Renders the date value in ISO 8601 format UTC.
+ */
+const DatetimeText = ({ datetime, excludeTime }: { datetime: Date; excludeTime?: boolean }) => {
+  // In order to force a consistent locale and timezone for the unit tests, we set
+  // the following env vars. In the browser, we don't have these set and the function
+  // behaves as if no parameters were passed to it. Note: Setting the env vars and
+  // not explicit setting them here works in non-Windows environments. We are only
```
yanirs, line 12:
```suggestion
  // not explicitly setting them here works in non-Windows environments. We are only
```

PR #119, opened 2020-06-11
Add jest:clear NPM Script

dhurlburtusa:
The need to clear the Jest cache happens once in awhile but frequently enough that it would be nice to have a convenience script to handle it instead of looking up the command each time.(Someday I might have `./node_modules/.bin/jest --clearCache` memorized, but I'd still rather type `npm run jest:clear`.)

## How has this been tested?

Ran the script, get a message saying what directory was cleared. Then running the tests again makes them succeed again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "docker-start": "docker run -p 8888:3000 abacus:latest",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
+    "jest:clear": "jest --clearCache",
     "lint": "npm run lint:ts && npm run lint:css",
     "lint:css": "stylelint **/*.{css,scss}",
     "lint:css:fix": "stylelint **/*.{css,scss} --fix",


package.json
```diff
@@ -96,6 +96,7 @@
     "docker-start": "docker run -p 8888:3000 abacus:latest",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
+    "jest:clear": "jest --clearCache",
```
dhurlburtusa, line 99: Maybe rename to `jest:clear:cache` or `jest:clear-cache`.

yanirs:
I wonder about the need for this versus simply calling `jest`. It has a lot of CLI arguments, and we don't want to add a command for each one...

dhurlburtusa:
When Jest is called with `--clearCache`, all other CLI options are effectively igored.
Because of that, I could call `npm run test:unit -- --clearCache && npm run test:unit`.

Not a big deal if this is not accepted. Just a little convenience. When I start using it frequently enough, I eventually remember the command.

yanirs:
Thanks. Let's leave it out for now. Can always revisit in the future. :slightly_smiling_face:

PR #121, opened 2020-06-12
Style App Header

dhurlburtusa:
Styles the app header as part of styling the experiment details.

## How has this been tested?

Ran the app and it visually matches the mockup.


components/Layout.tsx
```diff
@@ -15,37 +16,43 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
...
-          </header>
+          <AppBar color='inherit' position='relative'>
+            <div className='top'>
+              <Container maxWidth='xl'>
+                <span className='app-name'>Abacus</span>
+              </Container>
+            </div>
+            <div className='bottom'>
+              <Container maxWidth='xl'>
+                <nav>
```
dhurlburtusa, line 33: Can style the nav to match closer to the mockup before merging or can wait until later.

components/Layout.tsx
```diff
@@ -15,37 +16,43 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
...
-                <Link href='/metrics'>
-                  <a>Metrics</a>
-                </Link>
-              </nav>
-            </Container>
-          </header>
+          <AppBar color='inherit' position='relative'>
+            <div className='top'>
+              <Container maxWidth='xl'>
+                <span className='app-name'>Abacus</span>
```
dhurlburtusa, line 28: Still need to add a logo, but that can be added later.

yanirs: Can use `/img/logo.png`.

styles/partials/_theme.scss
```diff
@@ -78,6 +78,40 @@ h6 {
...
+    font-size: 1.5rem;
+  }
+}
+
+/**
+| MUI AppBar:
+===================================== */
+.MuiAppBar {
+  &-root {
+    background: #194661 !important;
```
yanirs, line 102: Can we do this without `!important`?

yanirs:
Mostly looks better, but can you please fix the tests and reintroduce `margin-top` to `h1-6` tags? I'm not sure why it was reduced to zero in #117.

components/Layout.tsx
yanirs: Probably worth doing now so we don't have to revisit it.

styles/partials/_theme.scss
dhurlburtusa: I tried to get MUI to not set a background color but I don't see a way around it without taking the time to learn their CSS-in-JS solution or theming solution. There has been talk about using a custom theme which may affect this.
There are two possible solutions I know of:
1) Use a selector with a higher specificity. Currently, the one here and the one MUI uses are the same but MUI's comes later in the cascade.
2) Use an inline style. (I'd prefer to avoid these in almost all cases.)

components/Layout.tsx
dhurlburtusa: I'll go ahead and use that for now.

components/Layout.tsx
```diff
@@ -15,37 +16,43 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
...
-                <Link href='/metrics'>
-                  <a>Metrics</a>
-                </Link>
-              </nav>
-            </Container>
-          </header>
+          <AppBar color='inherit' position='relative'>
+            <div className='top'>
+              <Container maxWidth='xl'>
+                <img alt='logo' className='mr-1' src='/img/logo.png' width='24' />
```
yanirs, line 28: I thought we were going to use classes like `mr-1` sparingly. :slightly_smiling_face: 
Won't it be better to style this with a specific class (e.g., `app-logo`, following the `app-name` example). This would also include the width setting rather than inlining it.

components/Layout.tsx
```diff
@@ -15,37 +16,43 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
       return renderError ? (
         <RenderErrorView renderError={renderError} />
       ) : (
-        <>
+        <div className='app-content'>
```
yanirs, line 19: Why do we need this extra class? Looking at the generated markup, we already have a `div` with class `app` and the `app-content` `div` is its only child.

styles/partials/_theme.scss
yanirs: Let's use a more specific selector. Or better, follow the MUI approach.

Looks like their [CssBaseline](https://material-ui.com/components/css-baseline/) obviates the need for some of the CSS that was recently introduced.

dhurlburtusa: Will handle in #122.

styles/partials/_utilities.scss
```diff
@@ -41,6 +41,10 @@
   margin: 0;
 }
 
+.mr-1 {
```
yanirs, line 44: Can we remove this now?

PR #127, opened 2020-06-15
Clean up stylesheets

### LINKED ISSUE #122: Clean up stylesheets
#117 introduced some changes that we probably don't want:

* Zero `margin-top` for `h1-6` and `p`.
* Questionable splits of styling of elements (e.g., `h1-6` rules are introduced under `_theme.scss` and `_elements.scss`).
* Duplication of MUI's rules, e.g., around https://material-ui.com/components/css-baseline/.
* Unintended changes to input forms (see screenshot below).
* A commenting style that's hard to follow (https://github.com/Automattic/abacus/pull/117#discussion_r438454357).

Let's clean it up before things get too hard to maintain. :slightly_smiling_face: 

image

dhurlburtusa:
Clean up stylesheets.

Resolves #122.

## How has this been tested?

No new code to test, but ran `npm run verify` anyway to double check. All is good.


styles/main.scss
```diff
@@ -1,4 +1,67 @@
-@import 'partials/elements';
-@import 'partials/forms';
-@import 'partials/utilities';
-@import 'partials/theme';
+/**
```
dhurlburtusa, line 1: After removing all the styles we are not currently using on `master`, I decided to just move all that was left to this main file.

styles/main.scss
```diff
@@ -1,4 +1,67 @@
...
+| App:
+===================================== */
```
dhurlburtusa, line 3: We don't really have documentation of CSS needs at the moment. But lines created using `=======` makes it easier, at least for me, to find the different sections in the file.
If you have a suggestion of a different style, I'll update them. I took a look at Knyle, MDCSS, and others. I recall looking at these many years ago. If we needed documentation beyond a component/section label, I'd probably pick MDCSS.

styles/main.scss
yanirs, line 1: I wonder if we need these comment headings. They basic repeat the name of the class, but we are more like to search for the class.

styles/main.scss
```diff
@@ -1,4 +1,54 @@
...
+    }
+
+    .bottom {
+      background: #fff;
+      padding: 0.75rem 0;
+    }
+  }
+}
+
+.datetime-text {
```
yanirs, line 45: Thinking about CSS in JSX again, `DatetimeText` and `ErrorsBox` feel like natural candidates for it. Since we already include `styled-jsx`, what do you think about moving these styles there?

styles/main.scss
```diff
@@ -1,4 +1,54 @@
...
+      &:last-child {
+        margin-right: 0;
+      }
+    }
+  }
+}
+
+.MuiAppBar {
+  &-root {
+    background: #194661 !important;
```
yanirs, line 32: Can we make this more specific and avoid the `!important` or use a MUI-recommended way to set the background colour?

jessie-ross: The proper way to do this in MUI is with a Theme context, there is a good guide for setting it up

components/ErrorsBox.tsx
```diff
@@ -1,5 +1,15 @@
+import { makeStyles } from '@material-ui/core/styles'
 import React from 'react'
 
+const useStyles = makeStyles({
+  root: {
+    background: '#f8d7da',
+    border: '1px solid #f5c6cb',
```
dhurlburtusa, line 7: @jessie-ross For this border color, do think it best to hard code the value here or should I get it from the theme's palette via something like `theme.palette.error.main`?

jessie-ross: Yep you should get it fro the theme object.

If you use a function parameter for `makeStyles` it will pass in the `theme` object via context. This is the TypeScript way to do it:
```js
const useStyles = makeStyles((theme: Theme) => createStyles({
  root: {
    borderWidth: 1;
    borderStyle: 'solid';
    borderColor: theme.palette.error.main;
  },
}));
```

components/ErrorsBox.tsx
```diff
@@ -1,5 +1,19 @@
+import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
+import Typography from '@material-ui/core/Typography'
 import React from 'react'
 
+const useStyles = makeStyles((theme: Theme) =>
```
dhurlburtusa, line 5: @jessie-ross How's this look?

I also added a new background color for error.

jessie-ross:
This looks fine for now.


@dhurlburtusa
 **FYI** :
I am not sure if you've seen this but it just isn't applicable here, there is an automated version of snapshot testing:
https://jestjs.io/docs/en/snapshot-testing

Basically you just use:
```
expect(component).toMatchSnapshot()
```
The first time the test runs it sets the snapshot into a special snapshot file (it should get committed into git).
On subsequent runs it checks the output html against the snapshot.
Then run the documented command to update the snapshot when the component changes.

yanirs:
> @dhurlburtusa **FYI** :
> There is an automated version of snapshot testing:
> https://jestjs.io/docs/en/snapshot-testing
> 
> I am not sure if you've seen this but it just isn't applicable here.
> Instead of inlineSnapshot you might use the following:
> 
> ```
> expect(component).toMatchSnapshot()
> ```
> 
> The first time the test runs it sets the snapshot into a special snapshot file (it should get committed into git).
> On subsequent runs it checks the output html against the snapshot.
> Then run the documented command to update the snapshot when the component changes.

Thanks @jessie-ross. This sounds similar to [a discussion I had with @withinboredom on a recent PR](https://github.com/Automattic/abacus/pull/99#discussion_r437994507): [`toMatchInlineSnapshot()` can also be managed automatically by Jest](https://jestjs.io/docs/en/snapshot-testing#inline-snapshots) (with the help of Prettier). It's pretty nice for small snapshots since it obviates the need to jump between files. :slightly_smiling_face:

PR #128, opened 2020-06-16
DEMO: Style AppBar with Component Class Override

dhurlburtusa:
This PR demonstrates how to override classes of a built-in component by overriding it via a theme.

### DIFF
--- a/components/Layout.tsx
+++ b/components/Layout.tsx
@@ -22,7 +22,7 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
             <meta charSet='utf-8' />
             <meta name='viewport' content='initial-scale=1.0, width=device-width' />
           </Head>
-          <AppBar color='inherit' position='relative'>
+          <AppBar position='relative'>
             <div className='top'>
               <Container maxWidth='xl'>
                 <img alt='logo' className='app-logo' src='/img/logo.png' />
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -1,6 +1,7 @@
 import '@/styles/main.scss'
 
 import CssBaseline from '@material-ui/core/CssBaseline'
+import { createMuiTheme, ThemeProvider } from '@material-ui/core/styles'
 import debugFactory from 'debug'
 import { AppProps } from 'next/app'
 import Head from 'next/head'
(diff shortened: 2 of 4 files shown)


components/Layout.tsx
```diff
@@ -22,7 +22,7 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
             <meta charSet='utf-8' />
             <meta name='viewport' content='initial-scale=1.0, width=device-width' />
           </Head>
-          <AppBar color='inherit' position='relative'>
+          <AppBar position='relative'>
```
dhurlburtusa, line 25: Change to using the default color class which is 'primary'. `inherit` is not overridable (at least that's how it was behaving).

pages/_app.tsx
```diff
@@ -14,6 +15,16 @@ import { getAuthClientId, getExperimentsAuthInfo, saveExperimentsAuthInfo } from
 
 const debug = debugFactory('abacus:pages/_app.tsx')
 
+const theme = createMuiTheme({
+  overrides: {
```
dhurlburtusa, line 19: To override, you use this `overrides` property keyed by the component's name. The value is an object with the component's classes to be overridden (`colorPrimary` in this case). Each class is an object representing the CSS styles.

pages/_app.tsx
```diff
@@ -14,6 +15,16 @@ import { getAuthClientId, getExperimentsAuthInfo, saveExperimentsAuthInfo } from
...
+    MuiAppBar: {
+      colorPrimary: {
+        backgroundColor: '#194661',
```
dhurlburtusa, line 22: Implication: All `MuiAppBar`'s that set the `color` prop to `primary` will have this style. This can be overridden per component instance if desired.

styles/main.scss
```diff
@@ -29,8 +29,6 @@
 
 .MuiAppBar {
   &-root {
-    background: #194661 !important;
-
     .top {
```
dhurlburtusa, line 32: This demo is not focusing on how to style the children of a MUI component.

pages/_app.tsx
```diff
@@ -58,14 +69,16 @@ const App = React.memo(function App(props: AppProps) {
...
           </Head>
-          <CssBaseline />
-          {renderError ? (
-            <RenderErrorView renderError={renderError} />
-          ) : (
-            <div className='app'>
-              <Route {...routeProps} />
-            </div>
-          )}
+          <ThemeProvider theme={theme}>
```
dhurlburtusa, line 72: Must explicitly set the theme for the overrides to take effect.

pages/_app.tsx
```diff
@@ -1,6 +1,7 @@
 import '@/styles/main.scss'
```
dhurlburtusa, line 1: I'm not sure why, but this implementation didn't explicitly need the extra code for the SSR.

pages/_app.tsx
```diff
@@ -14,6 +15,16 @@ import { getAuthClientId, getExperimentsAuthInfo, saveExperimentsAuthInfo } from
 
 const debug = debugFactory('abacus:pages/_app.tsx')
 
+const theme = createMuiTheme({
```
dhurlburtusa, line 18: I plan to move this theme creation to the `styles` directory in its own module.

PR #129, opened 2020-06-16
DEMO: Style AppBar by Redefining Component's Class Styles

dhurlburtusa:
This PR demonstrates how to style the AppBar component instance by redefining the component's class styles using CSS-in-JS.


components/Layout.test.tsx
```diff
@@ -20,7 +20,7 @@ test('renders layout with declared title and children', () => {
   // new links, are being tested.
   expect(headerElmt).toMatchInlineSnapshot(`
     <header
-      class="MuiPaper-root MuiAppBar-root MuiAppBar-positionRelative MuiAppBar-colorInherit MuiPaper-elevation4"
+      class="MuiPaper-root MuiAppBar-root MuiAppBar-positionRelative MuiAppBar-colorPrimary makeStyles-colorPrimary-1 MuiPaper-elevation4"
```
dhurlburtusa, line 23: This solution adds another CSS class to the DOM.

components/Layout.tsx
```diff
@@ -10,52 +11,62 @@ import { onRenderError } from '@/event-handlers'
 import RenderErrorBoundary from './RenderErrorBoundary'
 import RenderErrorView from './RenderErrorView'
 
-const Layout = ({ title, error, children }: { title: string; error?: Error | null; children?: ReactNode }) => (
```
dhurlburtusa, line 13: Most of this file is indentation changes. Not sure why GH is diffing it like it does.

components/Layout.tsx
```diff
@@ -10,52 +11,62 @@ import { onRenderError } from '@/event-handlers'
...
-            <h1>{title}</h1>
-            {error && <ErrorsBox errors={[error]} />}
-            {children}
-          </Container>
-          <footer>
+const useStyles = makeStyles({
+  colorPrimary: {
+    background: '#194661',
+  },
+})
```
dhurlburtusa, line 18: This creates a React hook with the classes and styles we want to add/override.

components/Layout.tsx
```diff
@@ -10,52 +11,62 @@ import { onRenderError } from '@/event-handlers'
...
+
+const Layout = ({ title, error, children }: { title: string; error?: Error | null; children?: ReactNode }) => {
+  const classes = useStyles()
```
dhurlburtusa, line 21: Then we use the React hook to get the dynamically generated class.

components/Layout.tsx
```diff
@@ -10,52 +11,62 @@ import { onRenderError } from '@/event-handlers'
...
+        return renderError ? (
+          <RenderErrorView renderError={renderError} />
+        ) : (
+          <>
+            <Head>
+              <title>{title} | Abacus</title>
+              <meta charSet='utf-8' />
+              <meta name='viewport' content='initial-scale=1.0, width=device-width' />
+            </Head>
+            <AppBar classes={classes} position='relative'>
```
dhurlburtusa, line 35: And we set the new class on the component.

components/Layout.tsx
```diff
@@ -10,52 +11,62 @@ import { onRenderError } from '@/event-handlers'
...
+              <div className='top'>
+                <Container maxWidth='xl'>
```
dhurlburtusa, line 37: The rest of this file is just indentation changes.

pages/_app.tsx
```diff
@@ -50,14 +57,6 @@ const App = React.memo(function App(props: AppProps) {
     <RenderErrorBoundary onError={onRenderError}>
       {({ renderError }) => (
         <>
-          <Head>
```
dhurlburtusa, line 53: These were moved to `_document.jsx`. That file wasn't working if I removed it's `<Head>` component. So I just moved all the stylesheet imports there. I didn't investigate how this affects SSR. But it seems to be the right place to put it.

pages/_document.tsx
```diff
@@ -0,0 +1,44 @@
+// Note: This file was adapted from
```
dhurlburtusa, line 1: This file enables server-side rendering with MUI. Without it, the styles still work but are not ready until the JavaScript kicks in. Get a FOUC.
Also, without this, an error is issued client-side saying that the DOM sent by the server doesn't match with the client.

components/Layout.tsx
yanirs: FYI: You can view the diff without indentation changes by adding `w=1` to the query string: https://github.com/Automattic/abacus/pull/129/files?w=1

yanirs:
Closing since we went with #128.

dhurlburtusa:
Closing. Went with #128 and #130

PR #130, opened 2020-06-16
DEMO Style Custom Components using MUI CSS-in-JS

dhurlburtusa:
This PR demonstrates styling custom components using MUI's CSS-in-JS solution.

### DIFF
--- a/components/DatetimeText.tsx
+++ b/components/DatetimeText.tsx
@@ -1,4 +1,11 @@
+import { makeStyles } from '@material-ui/core/styles'
 import React from 'react'
+
+const useStyles = makeStyles({
+  root: {
+    whiteSpace: 'nowrap',
+  },
+})
 
 const ISO_DATE_LENGTH = 10
 
@@ -6,6 +13,7 @@ const ISO_DATE_LENGTH = 10
  * Renders the date value in ISO 8601 format UTC.
  */
 const DatetimeText = ({ datetime, excludeTime }: { datetime: Date; excludeTime?: boolean }) => {
+  const classes = useStyles()
   // In order to force a consistent locale and timezone for the unit tests, we set
   // the following env vars. In the browser, we don't have these set and the function
   // behaves as if no parameters were passed to it. Note: Setting the env vars and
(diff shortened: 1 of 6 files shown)


components/ErrorsBox.tsx
```diff
@@ -8,14 +18,13 @@ interface Props {
...
   return (
-    <div className='errors-box'>
+    <div className={classes.root}>
       {props.errors.map((err) => (
         <React.Fragment key={err.message}>
-          <div className='error-box'>
-            <span className='error'>
-              <span className='error__msg'>{err.message}</span>
-            </span>
+          <div className='error-box_js'>
```
dhurlburtusa, line 26: I've adopted a convention for classes that are queried by JavaScript. The class is suffixed with `_js`. This is only used by the unit tests. We can consider a different solution. I might change it to use a snapshot instead.

components/ErrorsBox.tsx
```diff
@@ -1,5 +1,15 @@
+import { makeStyles } from '@material-ui/core/styles'
 import React from 'react'
 
+const useStyles = makeStyles({
```
dhurlburtusa, line 4: Note: There are other ways to use `makeStyles` to be dynamic/customizable. If a function is passed instead of an object literal, then the theme is make available which can be used to set the style's values. Or the properties can be functions instead of values as string. In this case, the component's `props` are passed.

components/DatetimeText.tsx
```diff
@@ -1,11 +1,19 @@
+import { makeStyles } from '@material-ui/core/styles'
 import React from 'react'
 
+const useStyles = makeStyles({
+  root: {
```
yanirs, line 5: Is calling it `root` the convention here? It'd be nice to keep the more informative class names if possible.

dhurlburtusa: It's the convention that MUI uses for their components and many times in the documentation they follow this convention. It's just a convention. That is, we can use one that better matches our needs. In the end, all but global styles get a unique class name. The actual name can be controlled using a class generator.

PR #131, opened 2020-06-17
Replace Sass with CSS-in-JS

dhurlburtusa:
Replace Sass with CSS-in-JS.

I think we might as well go all the way with CSS-in-JS.

Here is what I came up with.


components/AnalysisSummary.stories.tsx
```diff
@@ -1,5 +1,3 @@
-import '@/styles/main.scss'
```
dhurlburtusa, line 1: We end up loosing all the app styles once we get rid of Sass. I am going to test wrapping the storybook components with a theme provider to bring the styles back.

components/Layout.tsx
```diff
@@ -10,52 +11,85 @@ import { onRenderError } from '@/event-handlers'
...
+      color: '#4f4f4f',
+      fontFamily: 'Roboto, sans-serif',
+      marginRight: '1rem',
+      textDecoration: 'none',
+      '&:last-child': {
+        marginRight: 0,
+      },
+    },
+  },
+})
```
dhurlburtusa, line 42: I am wondering what you think about putting a bunch of styles together in one place like this. Think we should break these into smaller, individual components?

components/Layout.tsx
```diff
@@ -10,52 +11,85 @@ import { onRenderError } from '@/event-handlers'
...
-          <Container>
-            <h1>{title}</h1>
-            {error && <ErrorsBox errors={[error]} />}
-            {children}
-          </Container>
-          <footer>
+const useStyles = makeStyles({
+  appBarBottom: {
+    background: '#fff',
+    padding: '0.75rem 0',
```
dhurlburtusa, line 17: Should we use the theme's spacing to control these values?

components/Layout.tsx
```diff
@@ -10,52 +11,85 @@ import { onRenderError } from '@/event-handlers'
...
+            <Head>
+              <title>{title} | Abacus</title>
+              <meta charSet='utf-8' />
+              <meta name='viewport' content='initial-scale=1.0, width=device-width' />
+            </Head>
+            <AppBar position='relative'>
+              <div className={classes.appBarTop}>
+                <Container maxWidth='xl'>
+                  <img alt='logo' className={classes.appLogo} src='/img/logo.png' />
+                  <span className={classes.appName}>Abacus</span>
```
dhurlburtusa, line 62: I'm going to change this to use the `<Typography>` component.

package.json
```diff
@@ -99,9 +95,7 @@
     "format:check": "prettier --check .",
     "@comment jest": "Run jest with consistent TZ and LANG",
     "jest": "cross-env NODE_ICU_DATA=node_modules/full-icu TZ=America/New_York LANG=en-GB jest --coverage",
-    "lint": "npm run lint:ts && npm run lint:css",
-    "lint:css": "stylelint **/*.{css,scss}",
-    "lint:css:fix": "stylelint **/*.{css,scss} --fix",
+    "lint": "npm run lint:ts",
     "lint:ts": "eslint . --ext .ts,.tsx",
     "lint:ts:fix": "eslint . --ext .ts,.tsx --fix",
```
dhurlburtusa, line 100: Since we are not linting styles, how about:
```suggestion
    "lint": "eslint . --ext .ts,.tsx",
    "lint:fix": "eslint . --ext .ts,.tsx --fix",
```

pages/_app.tsx
```diff
@@ -16,9 +14,17 @@ import { getAuthClientId, getExperimentsAuthInfo, saveExperimentsAuthInfo } from
 
 const debug = debugFactory('abacus:pages/_app.tsx')
 
+const useStyles = makeStyles({
+  app: {
+    background: '#f4f6f8',
```
dhurlburtusa, line 19: I don't see anything in the palette of the default theme that would work well for this. Think we should keep this hard coded here or extend the palette? Or do something else?

components/Layout.tsx
```diff
@@ -10,52 +11,85 @@ import { onRenderError } from '@/event-handlers'
...
+    '& a': {
+      color: '#4f4f4f',
+      fontFamily: 'Roboto, sans-serif',
+      marginRight: '1rem',
+      textDecoration: 'none',
+      '&:last-child': {
+        marginRight: 0,
+      },
+    },
+  },
```
dhurlburtusa, line 41: I'm wondering if we should break these out into their own standalone components. They don't necessary need to be in a module for each one. They can live in this `Layout.tsx` module until they are reused elsewhere. Thoughts?

components/Layout.tsx
jessie-ross: Yep, `theme.spacing(sizeOfSpace)` but we can add it in progressively.

components/Layout.tsx
jessie-ross: Not a bad idea but I wouldn't say it is necessary for the moment, especially with such a minimalist Header.

components/Layout.tsx
jessie-ross: I think we can safely leave this as is.
The `Typography` component is useful for sharing styles whereas this logotype should always be unique.

package.json
jessie-ross: I personally don't mind leaving as is.

pages/_app.tsx
jessie-ross: Yeah changing the default theme is the right way to go here

components/Layout.tsx
dhurlburtusa: Sounds good.

pages/_app.tsx
```diff
@@ -14,6 +15,16 @@ import { getAuthClientId, getExperimentsAuthInfo, saveExperimentsAuthInfo } from
 
 const debug = debugFactory('abacus:pages/_app.tsx')
 
+const theme = createMuiTheme({
+  overrides: {
+    MuiAppBar: {
+      colorPrimary: {
+        backgroundColor: '#194661',
```
jessie-ross: Yeah this is generally done by altering the theme palette instead.

jessie-ross:
Overall this PR is fine, only recommendation is using the theme palette rather than using overrides, see the code comments for specifics.

I am not too I feel this repo's culture is but I think this is good to go, with a future PR addressing the theme palette.

pages/_app.tsx
dhurlburtusa: @jessie-ross Which way do you suggest we alter the theme palette?

1) Change the default value of an existing property?
  a) palette.primary.main?
  b) palette.background.default?
  c) Other?
2) Extend the default palette.
  a) Add palette.background.appHeader?
  b) Other?

components/AnalysisSummary.stories.tsx
jessie-ross: Ah I think I misunderstood when you were talking about MUI and storybook earlier.
Yep I would be very surprised if there wasn't a way to do this.

components/Layout.tsx
jessie-ross: I think this is nothing bad, there is nothing going on here that would complicate things if in the future we did have to seperate things.

pages/_app.tsx
jessie-ross: The Material Design system generally works off of a Primary and Secondary colour, they provide a sense of theme to an App. Ideally Primary and Secondary would have different meanings which would add another dimension to the UI.
https://material.io/design/color/the-color-system.html

This is similar the the Typography point, and I am realising now that I am not sure if y'all have bought into "Material Design" like I am imagining. This is something I will mention in a P2.

Well the first step here are to:
- Define Primary and Secondary colours for the App.
- One of Primary and Secondary will likely be the AppBar blue.
- We set the AppBar as Primary/Secondary

(Which colour will be Primary and which Secondary is an exercise in design semantics.)

dhurlburtusa: @jessie-ross I changed the theme from

```
const theme = createMuiTheme({
  overrides: {
    MuiAppBar: {
      colorPrimary: {
        backgroundColor: '#194661',
      },
    },
  },
  palette: {
    background: {
      error: '#f8d7da',
    },
  },
})
```
to
```
const theme = createMuiTheme({
  palette: {
    background: {
      error: '#f8d7da',
    },
    primary: {
      main: '#194661',
    },
  },
})
```
Now I get a server-side rendering conflict. Plus the icon renders at full size instead of.

To solve the SSR conflicts in a previous DEMO PR (#129), we had to introduce `pages/_document.tsx` (https://github.com/Automattic/abacus/pull/129/files?diff=unified&w=1#diff-8f3a5cf5019173ef1f56a54e45e44747R1). This is a Next.js thing if you are new to Next.js.

PR #144, opened 2020-06-20
Add ExperimentStatus Component

dhurlburtusa:
Add a reusable ExperimentStatus component. Plan to use in the experiments table and experiment details.

## How has this been tested?

Added new unit test.

Automated test continue to pass at 100% coverage.


yanirs:
Oh, I realised after merging that this should also have included using the component in the experiment table. Added #148.

PR #145, opened 2020-06-20
Add a LabelValuePanel Component

dhurlburtusa:
Add a LabelValuePanel component which displays a title along with label/value pairs rendered in a columnar fashion.

This is a general component where this style of data presentation is needed. We at least plan to use it the experiment details.

## How has this been tested?

Added a new unit test to test the component.

All automated tests still continue to run with 100% coverage.


components/LabelValuePanel.tsx
```diff
@@ -0,0 +1,46 @@
...
+import Table from '@material-ui/core/Table'
+import TableBody from '@material-ui/core/TableBody'
+import TableCell, { TableCellProps } from '@material-ui/core/TableCell'
+import TableHead from '@material-ui/core/TableHead'
+import TableRow from '@material-ui/core/TableRow'
+import Typography from '@material-ui/core/Typography'
+import React, { ReactNode } from 'react'
+
+/**
+ * A panel to display a label followed by it's value. The label/value pairs are
```
yanirs, line 11: `it's` -> `its`

components/LabelValuePanel.tsx
```diff
@@ -0,0 +1,46 @@
...
+  title: string
+}) {
+  const { data, title } = props
+  return (
+    <Paper>
+      <Table>
+        <TableHead>
+          <TableRow>
+            <TableCell colSpan={2}>
+              <Typography color='textPrimary' variant='h3'>
```
yanirs, line 25: It seems a bit odd to be using `h3` inside a `th`. Can't we simply apply the colour class to the cell?

yanirs: Actually, do we even need a `TableHead` for this heading? Is it better off as a standalone component followed by the table? That is, `<h3>title</h3><table><tbody>...data</tbody></table>`.

jessie-ross: (@yanirs, I'm not sure if there is confusion about this but I may as well reclarify: variant just means the style of an H3 is applied, not the html tag `<h3>`)

In the MUI examples they seem to put table titles above the table wrapped in a toolbar component:
https://material-ui.com/components/tables/#sorting-amp-selecting

But all up it might be efficient here to wrap around MaterialTable: 
https://material-ui.com/components/tables/#material-table

They provide a title attribute and it would be a matter of turning off the column headers/paging etc.

yanirs: > @yanirs, I'm not sure if there is confusion about this but I may as well reclarify: variant just means the style of an H3 is applied, not the html tag `<h3>`

@jessie-ross I don't think that's right. If you look at the snapshot tests, we do get a `<h3>` tag.

> But all up it might be efficient here to wrap around MaterialTable:
> https://material-ui.com/components/tables/#material-table
>
> They provide a title attribute and it would be a matter of turning off the column headers/paging etc.

Yeah, it's worth looking into if we can render it nicely.

components/LabelValuePanel.tsx
dhurlburtusa: Isn't English great. Use the apostrophe for contractions except for `it`.

components/LabelValuePanel.tsx
dhurlburtusa: @yanirs The `variant` determines the styles _and_ the default HTML element. `variant='h3'` means that the `Typography` component will be styled as defined by the `h3` styles (comes from the default theme with our current configuration) and it will render with an `h3` HTML element. This is because most of the `variants` map to an element with the same name. An example of an exception is the `body1` variant. It maps to a `p` HTML element but will be styled with `body1` styles.

components/LabelValuePanel.tsx
```diff
@@ -0,0 +1,51 @@
...
+  data,
+  title,
+}: {
+  data: { label: string; padding?: TableCellProps['padding']; value: ReactNode }[]
+  title: string
+}) {
+  const classes = useStyles()
+  return (
+    <Paper>
+      <Typography className={classes.title} color='textPrimary' variant='h3'>
```
yanirs, line 32: I'm still confused by this, but I posted internally so we can move forward.

PR #146, opened 2020-06-20
Allow To Resolve Segment Assignments and Group by Type

dhurlburtusa:
Adds the ability to resolve the segment ID of a segment assignment to the actual segment.
Adds the ability to group a list of resolved segment assignments by segment type.

## How has this been tested?

Unit tests added for new functions.

All tests continue to pass with 100% coverage.


models/SegmentAssignment.ts
```diff
@@ -45,3 +47,68 @@ export class SegmentAssignment {
...
+          })
+        } else {
+          console.error(`Unable to lookup segment with ID ${segmentAssignment.segmentId}.`)
+        }
+        return null
+      })
+      .filter(Boolean)
+  }
+
+  static groupByType(resolvedSegmentAssignments: ResolvedSegmentAssignment[]) {
```
jessie-ross, line 101: Is there a reason why we wouldn't use a generic group-by function instead of creating a specialised one? (e.g. _.groupBy)

Personally I am a fan of the "It is better to have 100 functions operate on one data structure than 10 functions on 10 data structures" philosophy.

models/ExperimentFull.ts
```diff
@@ -142,6 +151,31 @@ export class ExperimentFull implements ApiDataSource {
...
 
+  /**
+   * Resolves the segment ID of the segment assignment with the actual segment.
+   * If the ID cannot be resolved, then the "resolved" segment assignment will
+   * have a `null` `segment` property.
+   *
+   * @param segmentAssignments - The segment assignments to be resolved.
+   * @param segments - The segments to associate with the assignments.
+   */
+  static resolveSegmentAssignments(segmentAssignments: SegmentAssignment[], segments: Segment[]) {
```
dhurlburtusa, line 162: Could require that `segments` is an object of segments keyed by segment ID instead of an array of segments.

models/SegmentAssignment.ts
```diff
@@ -45,3 +47,68 @@ export class SegmentAssignment {
     })
   }
 }
+
+/**
+ * An segment assignment with the segment resolved.
+ */
+export class ResolvedSegmentAssignment {
```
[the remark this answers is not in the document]
dhurlburtusa: Updated.

models/SegmentAssignment.ts
dhurlburtusa: Removed. I'll use `_.groupBy`.

helpers/fixtures.ts
```diff
@@ -263,11 +265,26 @@ function createMetricsBares(numMetrics = 3) {
   return metrics
 }
 
+function createSegments(count: number) {
```
yanirs, line 268: Let's use the same approach @jessie-ross followed for `createMetricBares()` (`_.range().map()`, now on `master`).

models/ExperimentFull.ts
```diff
@@ -142,6 +151,31 @@ export class ExperimentFull implements ApiDataSource {
...
+    return segmentAssignments.map((segmentAssignment) => {
+      const segment = segments.find((segment) => segment.segmentId === segmentAssignment.segmentId) || null
```
yanirs, line 164: Let's not call `segments.find()` for every `segmentAssignment`. :slightly_smiling_face:

models/ExperimentFull.ts
yanirs: Maybe just do the mapping in this function?

Also, it's still unclear to me how this will be used and why it's a static function now that it's in `ExperimentFull` &ndash; it can be a method without the first argument. In any case, can you please update #106 to provide context?

models/ExperimentFull.ts
dhurlburtusa: That's the reason why I was suggesting the object keyed by segment ID instead of the array.

yanirs: This function can still do the mapping regardless of what it's passed. :slightly_smiling_face:  It's hard to know what would make more sense without the extra context of usages.

models/ExperimentFull.ts
dhurlburtusa: I thought about making it an instance method where the `segmentAssignments` come from the instance. Just the `segments` are passed in.

Either way works for me. The advantage to the static function is that an `ExperimentFull` instance doesn't have to exist to get the list of segment assignments. These can come from where ever.

But basically, it is for the AudiencePanel to display the segment assignments. The design groups the segment assignments by segment type. We also include the segment name.

```
const segmentsByType = _.groupBy(ExperimentFull.resolveSegmentAssignments(experiment.segmentAssignments, segments), _.property('segment.type'))
```
Or (if an instance method)
```
const segmentsByType = _.groupBy(experiment.resolveSegmentAssignments(segments), _.property('segment.type'))
```

I'll update #106 to show context. I'll let you know when that is done.

dhurlburtusa: @yanirs I pushed the code using an instance method to #106 so that you can see how it will be used in context. Its use can be seen at https://github.com/Automattic/abacus/pull/106/commits/1be7c987d8d20a0bde5243809f74427afc5adbc3#diff-3f30fd336375d65d6ecca21ba9c201e0R34.

yanirs: Thanks for that. I have some thoughts, but it'll all just get scattered across PRs if we proceed in this manner. Can you please follow the comments I left on #106 last week and split it by panel? You can either close the current PR, or add the Audience Panel here so we can discuss the necessity and implementation of this method with full context.

yanirs:
Closing since this will be a part of #160.

PR #150, opened 2020-06-22
Clean up layout

jessie-ross:
Mainly fixing the annoying footer:
- Sticky footer
- Typography on Footer
- Typography on the content header

Screenshot: https://d.pr/i/co23Tx

## How has this been tested?
- Manually tested
- Automated tests pass


components/Layout.test.tsx
```diff
@@ -20,34 +20,34 @@ test('renders layout with declared title and children', () => {
   // new links, are being tested.
   expect(headerElmt).toMatchInlineSnapshot(`
     <header
-      class="MuiPaper-root MuiAppBar-root MuiAppBar-positionRelative MuiAppBar-colorPrimary MuiPaper-elevation4"
+      class="MuiPaper-root MuiAppBar-root MuiAppBar-positionRelative MuiAppBar-colorPrimary makeStyles-appBar-2 MuiPaper-elevation4"
     >
       <div
-        class="makeStyles-appBarTop-2"
```
dhurlburtusa, line 26: I was wondering how stable these classes were going to be. @jessie-ross In your experience, do you find that you need to maintain these a lot? Or does it only happen on fairly rare occasions? Do you know of a better way to test?

Since Jest can auto-update the snapshots, it's usually not much of a maintenance issue -- at least in my experience.

components/Layout.tsx
```diff
@@ -49,13 +74,13 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
...
         ) : (
-          <>
+          <div className={classes.root}>
             <Head>
               <title>{title} | Abacus</title>
               <meta charSet='utf-8' />
               <meta name='viewport' content='initial-scale=1.0, width=device-width' />
             </Head>
-            <AppBar position='relative'>
+            <AppBar position='relative' className={classes.appBar}>
```
dhurlburtusa, line 83: My personal preference is to keep React component props in alphabetical order in most cases. Some of the exceptions I use:
1) `key` first.
2) `ref` first or up front.
3) Put props that affect the applicability of other props up front.
4) `type` prop of button.
5) `type` prop of inputs.

There might be others but I am not thinking of them now.

Unfortunately, this is quite opinionated. But if there is some common ground we agree on, then https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md may help if it can be configured to enforce what we agree on.

The [reservedFirst](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md#reservedfirst) option might be something we want to enforce.

@jessie-ross, @withinboredom, @yanirs Thoughts?

yanirs: I'm happy with having a consistent order as long as it's enforceable automatically.

components/Layout.tsx
```diff
@@ -75,17 +100,19 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
                 </Container>
               </div>
             </AppBar>
-            <Container>
-              <h1>{title}</h1>
+            <Container className={classes.content}>
+              <Typography variant='h4' component='h2' className={classes.contentTitle}>
```
yanirs, line 104: Why the switch from `h1`? I know we discussed it partly on Slack, where you said:
> H1's are supposed to be used once per page, they are almost always not “text” and generally wrap a logotype in the header. When they aren’t wrapping the logo they are found as a “Hero” component, those huge texts on landing pages, this is why they provide such a large H1.

That's not entirely accurate. For example, if you look at WordPress sites, the post title on a post page is `h1`. Using a `h2` styled as `h4` feels for the main page title feels like an unnecessary hack to me.

yanirs: If you disagree, please make a P2 post so we can discuss this in general.

components/Layout.tsx
```diff
@@ -49,13 +74,13 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
...
+          <div className={classes.root}>
             <Head>
               <title>{title} | Abacus</title>
               <meta charSet='utf-8' />
               <meta name='viewport' content='initial-scale=1.0, width=device-width' />
             </Head>
-            <AppBar position='relative'>
+            <AppBar position='relative' className={classes.appBar}>
               <div className={classes.appBarTop}>
                 <Container maxWidth='xl'>
```
yanirs, line 85: While you're here, can you please make the logo and name link to the root page?

components/Layout.tsx
dhurlburtusa: The HTML Living Standard (https://html.spec.whatwg.org/multipage/dom.html#sectioning-content and https://html.spec.whatwg.org/multipage/sections.html#headings-and-sections) explains how sectioning content (`section`, `aside`, `nav`, and `article`) are to be used with heading content (`h1` .. `h6` and `hgroup`).

When I first started with HTML and CSS, I used to use `h1`s for logos/brand names. Then many, many years later I learned about the [HTML Living standard](https://html.spec.whatwg.org/multipage/) and the expected semantic meaning. So, I got out of using `h1`s just because I wanted something with a large font-size. I try to only use the heading content to creating a correct document outline. However, I have not gone so far as to install a tool to tell me how the heading and sectioning content is affecting the outline. I'm not sure how many developers even take the outline into account. I don't take it too far because I have not ever had to do a compliant a11y project.

One thing that has not been clear to me is whether the site brand name (i.e. Abacus) should be the top-level item in the generated outline. I was thinking the main title (aka page title) for the particular page should be the top-level item.

Sometimes the hero title/caption represents the page's title. Other times it might be repeated across pages in the same group/category/etc.

But because the outline, sectioning, semantic HTML has never been a priority for any of my employers or clients, I never dug in deep enough to make sure the markup was correct in this sense.

yanirs: Well, in this case, this is the main heading of the page so `h1` seems appropriate, no?

components/Layout.test.tsx
jessie-ross: These will change whenever the classes change

jessie-ross: If you are adding/changing/removing a plain class you would need to run `--updateSnapshot` as well - so `--updateSnapshot`-wise I would call them equivalent.

They do add a bit more noise into the test diffs but for UI snapshots they are going to be noisy anyway.

components/Layout.tsx
jessie-ross: Yeah I am not too fussed about this at the moment - this is one of the easier problems to change later on. I will revert this back to a H1.

components/Layout.tsx
jessie-ross: No problemo

components/Layout.tsx
jessie-ross: I am going to leave this particular snippet as is until we have the automated tooling in place.

components/Layout.tsx
```diff
@@ -75,17 +102,19 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
                 </Container>
               </div>
             </AppBar>
-            <Container>
-              <h1>{title}</h1>
+            <Container className={classes.content}>
+              <Typography variant='h4' component='h1' className={classes.contentTitle}>
```
yanirs, line 106: I still don't get why we can't just leave it as `h1`. It looks better on `master`:

image


This branch:

image

components/Layout.tsx
yanirs: Thanks. I meant both the logo image and the name. Also, we probably don't want it underlined.

components/Layout.tsx
yanirs: Alright, I posted internally on this since it's a general issue, so not a blocker for this PR.

PR #153, opened 2020-06-23
Disable Coverage Via Jest for E2E Tests

### LINKED ISSUE #152: E2E Test Coverage Silently Failing
When CircleCI runs the e2e tests, the coverage part fails because the JSX has not been parsed. However, this parse failure does not fail the tests. It prints the errors but does not fail the tests.

Some possible solutions:
1) Add the `ts-jest` to `globals` in `jest.e2e.config.js`.
2) Remove coverage for e2e tests. 

The snippet for #\1:
```
  globals: {
    // Must specify a custom tsconfig for tests because we need the TypeScript
    // transform to transform JSX into js rather than leaving it as JSX which the
    // next build requires.
    'ts-jest': {
      babelConfig: true,
      tsConfig: '<rootDir>/tsconfig.jest.json',
    },
  },
```

After further testing, it looks like the test coverage doesn't work the same as with integration and unit tests. So, it looks like the solution is to disable coverage for E2E tests.

Note: The reason why this appeared to be working correctly before new routes/pages were added was because we were ignoring all these files.

dhurlburtusa:
Coverage isn't actually working correctly for the E2E tests with Puppeteer.

Tried to install https://www.npmjs.com/package/jest-puppeteer-istanbul but was unable to get it to work correctly. It behaved like two coverage reports were being generated. One report would always report that there was no coverage, the other report which seemed to be coming from `jest-puppeteer-istanbul` would report as expected.

We can revisit this in the future. For now, I think we should just disable the coverage reports since we are ignoring everything anyway.

Resolves #152.

## How has this been tested?

Automated tests still work but no attempt to generate coverage is done for e2e tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "@comment jest": "Run jest with consistent TZ and LANG",
-    "jest": "cross-env NODE_ICU_DATA=node_modules/full-icu TZ=America/New_York LANG=en-GB jest --coverage",
+    "jest": "cross-env NODE_ICU_DATA=node_modules/full-icu TZ=America/New_York LANG=en-GB jest",
     "lint": "npm run lint:ts",
     "lint:ts": "eslint . --ext .ts,.tsx",
     "lint:ts:fix": "eslint . --ext .ts,.tsx --fix",
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,5 @@
 module.exports = {
+  collectCoverage: true,
   collectCoverageFrom: [
     // Note: Dot directories (e.g., .storybook) are also excluded automatically.
     '**/*.{ts,tsx}',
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -94,7 +94,7 @@
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "@comment jest": "Run jest with consistent TZ and LANG",
-    "jest": "cross-env NODE_ICU_DATA=node_modules/full-icu TZ=America/New_York LANG=en-GB jest --coverage",
```
dhurlburtusa, line 97: There were two ways to enable/disable coverage.
1) Leave the `--coverage` option for the `jest` script and add a `--coverage=false` to each of the `test:e2e` scripts.
2) Remove the `--coverage` option from the `jest` script and only enable for unit and integration. Enabling could have been done via config file or CLI options.

I opt'd for option 2 since the CLI overrides the config file's values.

jessie-ross: Option 2 sounds good to me

PR #155, opened 2020-06-23
Fix E2E Tests

### LINKED ISSUE #154: E2E Tests Failing in Local Development Environment
The e2e tests are failing in my local development environment due to waiting for a navigation (i.e., client-side redirect) to occur.

This is happening on the current `master` branch.

I think I have a solution that will allow the tests to work locally on all developer's machines and in the CI environment.

dhurlburtusa:
Fix e2e tests by checking first the URL before determining whether we need to wait for a navigation.

Fixes #154.

## How has this been tested?

Ran tests locally. The tests pass. Hopefully, they will pass everywhere too with this change.

### DIFF
--- a/e2e/navigation.test.tsx
+++ b/e2e/navigation.test.tsx
@@ -6,8 +6,12 @@ jest.setTimeout(30000)
 
 describe('Dashboard', () => {
   it('should redirect to /experiments', async () => {
-    // This helps prevent race conditions around waitForNavigation and the redirect.
-    await Promise.all([page.goto('http://a8c-abacus-local:3000'), page.waitForNavigation({ timeout: 2000 })])
+    await page.goto('http://a8c-abacus-local:3000')
+
+    // Sometimes the redirect has not occurred yet. So, we need to wait.
+    if (new URL(page.url()).pathname === '/') {
+      await page.waitForNavigation({ timeout: 2000 })
+    }
 
     expect(new URL(page.url()).pathname).toBe('/experiments')
   })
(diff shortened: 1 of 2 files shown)


yanirs:
Works on my machine and LGTM, but I'd like to wait until @jessie-ross tests it locally before merging.

e2e/navigation.test.tsx
```diff
@@ -6,8 +6,12 @@ jest.setTimeout(30000)
 
 describe('Dashboard', () => {
   it('should redirect to /experiments', async () => {
-    // This helps prevent race conditions around waitForNavigation and the redirect.
-    await Promise.all([page.goto('http://a8c-abacus-local:3000'), page.waitForNavigation({ timeout: 2000 })])
+    await page.goto('http://a8c-abacus-local:3000')
+
+    // Sometimes the redirect has not occurred yet. So, we need to wait.
+    if (new URL(page.url()).pathname === '/') {
```
jessie-ross, line 12: This theoretically has the same race condition: the page can navigate between line 12 and line 13.

jessie-ross:
This works on mine too and I am fine with it.

e2e/navigation.test.tsx
dhurlburtusa: The call to `await page.goto` by default resolves when the `load` event of the browsing context is fire. The redirect code may or may not have run by then.

Without knowing the internals of Puppeteer, there may be a race condition between completing line 9 and the read of the current page URL.

jessie-ross: Yep that is right as well

jessie-ross: I think the problem here is that waitForNavigation fails the test if it times out, maybe we could switch to a setTimeout

PR #158, opened 2020-06-24
WIP: Extract Out Reusable Experiment Page

dhurlburtusa:
This extracts out the common parts of the experiment page for the different sections (currently details and results but snippets will be added later).

This should help ease the implementation of the "toolbar" which depends highly on the current section being rendered.

This is a WIP while attempting to create mocks for the API calls.

This intended to support #106 (experiment details).

## How has this been tested?

Has been manually tested. However, automated tests would be nice. But the page component, which is now under the `components` directory, makes API calls. The API calls used to be in the individual "route" components and had no tests since the plan was to write e2e tests.

If the API calls can be mocked, then we can write unit tests.


jest.config.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
     '!**/*stories.tsx',
     '!**/node_modules/**',
     '!<rootDir>/api/**', // We test these with integration tests.
+    '!<rootDir>/components/ExperimentsPage.tsx', // Ignore for unit testing.
```
dhurlburtusa, line 9: Until how to test is decided, ignoring this during the unit tests.

components/ExperimentsPage.tsx
```diff
@@ -0,0 +1,79 @@
...
+          setMetrics(metrics)
+          return
+        })
+        .catch(setFetchError)
+    }
+  }, [experimentId, section])
+
+  return (
+    <Layout title={`Experiment: ${experiment ? experiment.name : 'Not Found'}`} error={fetchError}>
+      <ExperimentTabs experiment={experiment} />
```
dhurlburtusa, line 70: The "toolbar" mentioned in the description will go here. It will have the tabs plus the "action" buttons that depend on state and current section/tab.

yanirs:
@dhurlburtusa I'm not a fan of this approach. It feels like a premature and possibly wrong abstraction. Both pages are fairly distinct and will evolve. Let's focus on getting #106 fully implemented before extracting common parts, which may be better served by a common layout and some helpers rather than the abstraction in this PR.

Feel free to reopen if you disagree.

PR #161, opened 2020-06-25
Add metric property parameter_type

jessie-ross:
- Add metric property parameter_type

## How has this been tested?
- Existing tests pass except for integration - swagger will be updated shortly

### DIFF
--- a/__tests__/api/MetricsApi.test.ts
+++ b/__tests__/api/MetricsApi.test.ts
@@ -12,6 +12,7 @@ describe('MetricsApi.ts module', () => {
         expect(typeof metric.metricId).toBe('number')
         expect(typeof metric.name).toBe('string')
         expect(typeof metric.description).toBe('string')
+        expect(typeof metric.parameterType).toBe('string')
       })
     })
   })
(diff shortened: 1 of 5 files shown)


yanirs:
> Existing tests pass except for integration - swagger will be updated shortly

Done, and tests are passing. :heavy_check_mark: 

Can you please also update `MetricFull.fromApiData()`?

dhurlburtusa:
@jessie-ross Would you please take a look at `models/MetricFull.ts` around line 46? There is a TODO there that is related to this change.

jessie-ross:
> Would you please take a look at models/MetricFull.ts around line 46? There is a TODO there that is related to this change.

Good catch! I did have some thoughts about that.
I think we should derive it on the server since MetricBare doesn't include the eventParams/revenueParams. 

I will add a commit fixing this to the PR.

jessie-ross:
It is still failing, I am going to take a look at the swagger file...

yanirs:
> It is still failing, I am going to take a look at the swagger file...

Looks like it's the unit tests, not Swagger's fault. :slightly_smiling_face:

PR #168, opened 2020-06-27
Add Experiment Details Conclusions Panel

dhurlburtusa:
Add Experiment details Conclusions panel.

## How has this been tested?

New tests for new component.
Automated tests still pass with 100% coverage.

### DIFF
--- a/components/ConclusionsPanel.tsx
+++ b/components/ConclusionsPanel.tsx
@@ -1,0 +1,41 @@
+import { createStyles, makeStyles } from '@material-ui/core/styles'
+import React from 'react'
+
+import LabelValuePanel from '@/components/LabelValuePanel'
+import { ExperimentFull } from '@/models'
+
+const useStyles = makeStyles(() =>
+  createStyles({
+    conclusionless: {
+      opacity: 0.6,
+    },
+  }),
+)
+
+/**
+ * Renders the conclusion information of an experiment in a panel component.
+ *
+ * @param props.experiment - The experiment with the conclusion information.
+ */
+function ConclusionsPanel({ experiment }: { experiment: ExperimentFull }) {
+  const classes = useStyles()
+  const deployedVariation = experiment.variations.find(
+    (variation) => experiment.deployedVariationId === variation.variationId,
+  )
+  const isConclusionless = !(experiment.endReason || experiment.conclusionUrl || deployedVariation)
+  const data = [
+    { label: 'Description for ending experiment', value: experiment.endReason },
+    {
+      label: 'Conclusion URL',
+      value: !!experiment.conclusionUrl && (
+        <a href={experiment.conclusionUrl} rel='noopener noreferrer' target='_blank'>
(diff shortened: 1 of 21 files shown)


components/ExperimentDetails.tsx
```diff
@@ -29,6 +30,8 @@ function ExperimentDetails({
       <AudiencePanel experiment={experiment} segments={segments} />
       <br />
       <MetricAssignmentsPanel experiment={experiment} metrics={metrics} />
+      <br />
```
dhurlburtusa, line 33: Responsive layout of the panels will come in a later PR.

components/ConclusionsPanel.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * Renders the conclusion information of an experiment in a panel component.
+ *
+ * @param props.experiment - The experiment with the conclusion information.
+ */
+function ConclusionsPanel({ experiment }: { experiment: ExperimentFull }) {
+  const classes = useStyles()
+  const deployedVariation = experiment.variations.find(
+    (variation) => experiment.deployedVariationId === variation.variationId,
+  )
+  const isConclusionless = !(experiment.endReason || experiment.conclusionUrl || deployedVariation)
```
dhurlburtusa, line 25: Note: Not using the status as a factor to determine whether there is not conclusion any information. The API should ensure that conclusion data can't be entered when status is staging or running.

yanirs:
@dhurlburtusa can you please make this PR depend on `master` rather than stacked on the others? As far as I can tell, there is no need for it to depend on other PRs.

components/ConclusionsPanel.tsx
yanirs: Yeah, it's good not to rely on the status. However, I think it may be better to not render the panel at all if there are no conclusions. @aaronyan what do you think?

components/ConclusionsPanel.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const data = [
+    { label: 'Description for ending experiment', value: experiment.endReason },
```
yanirs, line 27: I know the label came from the design, but it doesn't seem exactly right. @aaronyan what do you think of labelling it "Reason the experiment ended" or something along these lines? We already auto-populate it with `Discarded` on transition from `staging` to `disabled`.

components/ConclusionsPanel.tsx
aaronyan: Yeah, for simplicity let's not render the Conclusions panel when no conclusions are present. I like the idea of keeping the Conclusions panel visible (buy greyed out), but I think that would introduce unnecessary confusion at this stage, unless I figure out how to communicate when the Conclusions can be edited. For now, let's not render if no conclusions exist.

components/ConclusionsPanel.tsx
aaronyan: Sounds good to me! I like your revised label "Reason the experiment ended".

PR #169, opened 2020-06-27
Add Experiment Details Responsive Layout

dhurlburtusa:
Add Experiment Details Responsive Layout.

## How has this been tested?

Added new tests for updated component.
Manually tested.


dhurlburtusa:
@yanirs It's because the data table in Metrics panel won't shrink any narrower than about 404px wide.

If we reduce the font-size of the table head cells and reduce the padding of all table cells for the metrics data table, then we can get everything to fit nicely on a 360px wide screen. Thoughts?

Usually, I try to style down to 320px wide but no smaller. Most devices are 360px or wider. Because we don't have a general audience, I think supporting 360px or wider should be sufficient. Thoughts?

yanirs:
> @yanirs It's because the data table in Metrics panel won't shrink any narrower than about 404px wide.
> 
> If we reduce the font-size of the table head cells and reduce the padding of all table cells for the metrics data table, then we can get everything to fit nicely on a 360px wide screen. Thoughts?
> 
> Usually, I try to style down to 320px wide but no smaller. Most devices are 360px or wider. Because we don't have a general audience, I think supporting 360px or wider should be sufficient. Thoughts?

@dhurlburtusa from some quick testing with production data it seems like it's not only the Metrics panel. It looks like even the description in the `General` panel doesn't wrap properly. We may need to rethink the use of tables. In any case, 360px seems reasonable for now. Reducing the padding sounds like it'd work, but it'd be good also get @jessie-ross's opinion.

jessie-ross:
@yanirs @dhurlburtusa

**Firstly I don't think Mobile is a great idea for this app.**

Generally tables aren't well suited for phone-size devices, neither are high information-density apps.

Mobile design requires a lot of reworking to get it usable and it needs to be considered at every UI choice. Then it adds more complexity in development by having to maintain two different designs intertwined in the same App.

If we imagine an Abacus Experimenters (a Product Engineer I suspect) and Experiment Viewers (perhaps Business and Strategy people), even considering future dreams, I think they would at most be using this application on a tablet if not a laptop/desktop.

I think we could do ourselves a great service if we scope to tablets and above.
We could have a simpler application.
In my opinion simplicity will get use further in our primary goal of a high correctness application.

**But what if we really want mobile?**

I should note here that I consider tablets and desktops together as they have similar enough levels of real-estate and usability. I should also note that responsive design is not equivalent to it working on mobile, and I am for having a responsive design.

There is a certain spectrum of strategies for mobile/web design:
- Mobile first, provides a compromised view on tablets/desktops.
- Mobile/Desktop, compromised on both.
- Desktop first, provides a compromised view on mobile 

Sometimes a perfect solution exists: 
- Mobile/Desktop, not compromised, but likely should need a bigger team and budget, more design work.

I think if we do want to go down the mobile path we should be mostly Desktop first, based on my appraisal of our users above. This means that we make it look as good as possible on desktop, making decisions like using tables which are great for desktop, and then providing workarounds to make it usable on Mobile. This is in contrast to changing out things which work great for Desktop users in order to provide improved usability for mobile. 

In this particular case, it would just be fiddling with font-size, paddings, margins to get them to fit. 
Certain grids may need to be flattened somehow, I find it easiest to do this with divs and breakpoints.

aaronyan:
+1 to the "Desktop first, mobile second" when prioritizing experience and design.

> In this particular case, it would just be fiddling with font-size, paddings, margins to get them to fit. Certain grids may need to be flattened somehow, I find it easiest to do this with divs and breakpoints.

I have been putting off the mobile mockups for some time. But maybe it's time I get those for y'all! :) That may help with guiding how the font-size, paddings, margins, etc to get things to fit. I'll spend some time tomorrow thinking about and drafting up what this details page might look like on mobile.

Things like the experiment list is going to be pretty tricky on mobile. We'll likely sacrifice lots of information to make the limited horizontal real estate fit. This is kind of a tangent, but just wanted to mention that we may add the "remove information" technique in order to get things to fit on mobile.

yanirs:
Thanks @jessie-ross and @aaronyan!

> +1 to the "Desktop first, mobile second" when prioritizing experience and design.

Agreed, but we need to make mobile usable, and not as an afterthought. Experimenters should be able to see experiment details and basic results on mobile. That's the rationale behind many of our decisions so far &ndash; making it work on tablets isn't enough. Also, there's an official directive from Matt that internal tools should be responsive (p10gg3-7TI-p2#comment-26337).

> I have been putting off the mobile mockups for some time. But maybe it's time I get those for y'all! :) That may help with guiding how the font-size, paddings, margins, etc to get things to fit. I'll spend some time tomorrow thinking about and drafting up what this details page might look like on mobile.

I'm not sure that more mockups are a good use of our time, as the implementation reality will be different from mockups. Why not fiddle with the sizing in the context of PRs like this one? The changes shouldn't be huge if we're going for a compromised-but-usable mobile view.

> Things like the experiment list is going to be pretty tricky on mobile. We'll likely sacrifice lots of information to make the limited horizontal real estate fit. This is kind of a tangent, but just wanted to mention that we may add the "remove information" technique in order to get things to fit on mobile.

Given that we're using MUI, I don't think they have to be _that_ tricky. It _should_ be fairly simple to collapse columns to DetailsView elements, similarly to the implementation in #173.

components/ExperimentDetails.tsx
```diff
@@ -22,21 +25,40 @@ function ExperimentDetails({
...
-      <br />
-      <MetricAssignmentsPanel experiment={experiment} metrics={metrics} />
-      {experiment.hasConclusionData() && (
-        <>
-          <br />
-          <ConclusionsPanel experiment={experiment} />
-        </>
+    <Grid container spacing={2}>
+      <Grid item xs={12}>
+        <h2>Experiment details</h2>
```
yanirs, line 34: I think we can remove this heading since we already have the main page heading and panel headings.

Issue #164, opened 2020-06-26
Decide whether we want to use Chip or a different component


yanirs:
See https://github.com/Automattic/abacus/pull/160#discussion_r445210297

This probably also applies to `ExperimentStatus`: #148

aaronyan:
The [Chip component](https://material-ui.com/components/chips/) sounds great! I say we also adjust the `ExperimentStatus` to use the Chip component as well.

jessie-ross:
I would prefer not to, based solely on the style guide's recommendations:
> Usage: Chips allow users to enter information, make selections, filter content, or trigger actions.
https://material.io/components/chips#usage

This is the reasoning I see about applying the above to our project:
1. Chips are very useful components for interactivity. 
2. Good design separates interactive from non-interactive - part of the language of our design.
3. We should want to preserve them for interactive use.

I like the alternate idea we had - caption font, uppercase, and mild transparency.
RE: The ExperimentStatus in the #170: the legibility problem seems to be a bug due to double opacity being applied.

dhurlburtusa:
If chips are intended for interactivity, then avoiding them for what we were thinking of using them for sounds good to me. What we are using them for is what many UI libraries call a Label component. I've seen it in Semantic UI. It's also what GitHub calls their components used for similar purposes. There are several "labels" on this page -- the purple Merged label and this PR's status label (currently Open).

Basically, we need a component that highlights are particular piece of textual data.

aaronyan:
Thanks for the context and reasoning. Let's move ahead with not using a Chip component for `ExperimentStatus`. Our status will be more of a highlight of existing text, rather than an interactive button/chip.

Combining these ideas together:

> highlights are particular piece of textual data
> I like the alternate idea we had - caption font, uppercase, and mild transparency

<img width="1257" alt="Screen Shot 2020-06-29 at 5 44 55 PM" src="https://user-images.githubusercontent.com/4505888/86069965-4bbb8b00-ba30-11ea-932c-e633cdf1e7d0.png">

Also commenting here with that same information. https://github.com/Automattic/abacus/pull/170#pullrequestreview-438832710

yanirs:
@aaronyan as noted on https://github.com/Automattic/abacus/pull/170#discussion_r447412738, combining three types of highlighting (uppercasing, background colour, font colour) in that manner seems excessive and doesn't seem to be inline with the guidelines.

[...]

Issue #80, opened 2020-05-20
Add styling and site layout


aaronyan:
Once we have some mockups and styles, let's add them to Abacus! :)

Site layout:
- Headers, footers
- Menu, logo

Styling:
- Color scheme
- Font choices
- General style guides (margins, paddings)
- Heading and paragraph styles

aaronyan:
> Styling

We're going with [Material UI](https://material-ui.com/) for our design framework. And then we'll likely go with a [paid theme](https://material-ui.com/store/items/devias-kit-pro/) to style the color, typography, spacing, and overall look and feel.

> Headers, footers
> Menu

The header and footer will be basic:
- Dashboard
- Experiments
- Metrics

We will likely use a vertical left-side navigation drawer to be mobile friendly and give it a more app-like feel.

> Logo

Still got to make one! :). We can use Experimentation Platform logo for now, but I think I would like to create an Abacus logo to better match the theme we choose.

dhurlburtusa:
I'm removing myself from this as it is really an ongoing, team effort.

aaronyan:
Closing this for now since these gets added over time, and I don't think we'll ever stop iterating on this.

PR #170, opened 2020-06-28
Style experiments table

### LINKED ISSUE #148: Use ExperimentStatus component in the experiment table and adjust its styling
Following #144, we should use `ExperimentStatus` to style the statuses in the experiment table.

Styling adjustments should be done as suggested by @jessie-ross: https://github.com/Automattic/abacus/issues/164#issuecomment-651685173

jessie-ross:
Before:
Screen Shot 2020-06-29 at 12 04 06 am

After:
Screen Shot 2020-06-28 at 11 51 40 pm

See the figma mocks for the goals.

- Change column order
- Add monospace fonts
- Polish ExperimentStatus
- Simplify table features
- Polish up table with weights and opacity - see commit notes
- Moved the search to the left after trying and failing to efficiently style it - see commit notes

See also: p1593107552230200-slack-CNTD4G117

Resolves #148.

## Notes
- It might be worth considering another library like ag-grid past the MVP
- MetricTable will be styled after this PR and the MetricDetail PR for convenience's sake.

## How has this been tested?
- AT


components/ExperimentStatus.tsx
```diff
@@ -1,35 +1,43 @@
...
+    },
+  }),
+)
 
 function ExperimentStatus({ status }: { status: Status }) {
-  const classes = useStatusStyles()
-  return <span className={clsx(classes.root, classes[status])}>{status}</span>
+  const classes = useStyles()
+  return (
+    <Typography className={clsx(classes.root, classes[status])} variant='caption'>
```
dhurlburtusa, line 37: There has been discussion about "Chipifying" this status. Please see #164.

yanirs: Yeah, let's use this opportunity to chipify the status.

components/ExperimentsTable.test.tsx
```diff
@@ -34,12 +39,15 @@ test('with one page of experiments, renders a table', () => {
       status: Status.Staging,
     },
   ]
-  const { container } = render(<ExperimentsTable experiments={experiments} />)
+  const { container } = render(
+    <ThemeProvider>
```
yanirs, line 43: Do we need this wrapping in the tests? Feels like it defeats the purpose of isolating components.

components/ExperimentsTable.tsx
```diff
@@ -28,20 +32,49 @@ const ExperimentsTable = ({ experiments }: { experiments: ExperimentBare[] }) =>
...
+          field: 'ownerLogin',
+          cellStyle: {
+            fontFamily: theme.custom.fonts.monospace,
+          },
+        },
         {
           title: 'Start',
           field: 'startDatetime',
           render: (experiment) => <DatetimeText datetime={experiment.startDatetime} excludeTime />,
+          cellStyle: {
```
yanirs, line 66: I don't think we should use monospace for the start and end datetimes, but if we do, let's do it on the `DatetimeText` component level.

pages/_document.tsx
```diff
@@ -11,6 +11,10 @@ export default class Document extends NextDocument {
         <Head>
           <link rel='stylesheet' href='https://fonts.googleapis.com/css2?family=Comfortaa:wght@300;400&display=swap' />
           <link rel='stylesheet' href='https://fonts.googleapis.com/css2?family=Roboto:wght@400;700&display=swap' />
+          <link
```
yanirs, line 14: With this, we end up with ten requests to Google fonts. Do we really need them all?

utils/material-table.ts
```diff
@@ -1,7 +1,20 @@
...
+export const defaultTableOptions: Options = {
   emptyRowsWhenPaging: false,
   pageSize: 25,
-  pageSizeOptions: [25, 50, 100],
   showEmptyDataSourceMessage: false,
   showTitle: false,
+  search: true,
+  toolbar: true,
+  searchFieldVariant: 'standard',
+  searchFieldAlignment: 'left',
```
yanirs, line 11: I'm not a fan of left-aligning the search field as it distracts from the main table content. Any reason not to keep it on the right?

components/ExperimentStatus.tsx
```diff
@@ -1,35 +1,43 @@
...
+      padding: '3px 5px',
+      borderRadius: 3,
+      textTransform: 'uppercase',
+    },
+    completed: {
+      color: '#4CAF50',
+      background: 'rgba(76, 175, 80, 0.08)',
+    },
+    running: {
+      color: '#FF9800',
```
yanirs, line 20: This colour is barely readable on my screen. Do we really need colours and uppercasing for status? It'd be good to get @aaronyan's input. In my opinion, it doesn't add much to the current design.

components/ExperimentsTable.test.tsx
dhurlburtusa: We should need these any more. Storybook is now decorating each of the stories in the ThemeProvider. This code is on master. So, it can be done now.

pages/_document.tsx
dhurlburtusa: Whichever way this (i.e., contents of the `<Head>` component) is changed, we also need to sync it with `.storybook/preview.head.html`.

I'd prefer a way to auto-sync but the effort is likely more trouble than it is worth.

I think we should at least add a comment here that changes here need to be reflected in `.storybook/preview.head.html`.

components/ExperimentsTable.tsx
```diff
@@ -28,20 +32,49 @@ const ExperimentsTable = ({ experiments }: { experiments: ExperimentBare[] }) =>
   return (
     <MaterialTable
       columns={[
-        { title: 'Name', field: 'name' },
+        {
+          title: 'Name',
+          field: 'name',
+          cellStyle: {
+            fontFamily: theme.custom.fonts.monospace,
+            fontWeight: 600,
```
yanirs, line 40: Can we get this from the theme so that we use consistent weights for boldface?

components/ExperimentsTable.test.tsx
jessie-ross: > Do we need this wrapping in the tests? Feels like it defeats the purpose of isolating components.

It is needed here as we `useTheme` and we rely on a custom theme property.
The unit test crashes without it.

> We shouldn't need these any more. Storybook is now decorating...

As this is a unit test the Storybook config shouldn't affect it. 
I think the code is on this branch as well.

utils/material-table.ts
jessie-ross: The reasoning here was to match it to the mocks and to do something with the whitespace on the left - it felt like an imbalance, but I am fine with moving it back to the right.

yanirs: Let's get @aaronyan's opinion. :slightly_smiling_face:

components/ExperimentStatus.tsx
jessie-ross: I think the legibility problem is a bug where the opacity is being doubled.

pages/_document.tsx
jessie-ross: > With this, we end up with ten requests to Google fonts. Do we really need them all?

**There are two questions I see here:**
1. Do we need this many fonts?
2. Do we need this many requests?

**In short,** we can reduce the amount of fonts as well as requests, but I would say it isn't necessary at this point.


### Do we need this many fonts?

- I have added Roboto Mono italics bold and regular. I would also add italics for Roboto Sans.
   I think it is good to have these available for our fonts at this stage but they aren't being used.
- Comforta as far as I can see is only being used for the logo.
   We could simply include the logo as a `png`

### Do we need this many requests?

We can reduce the number of requests by including them all on the same stylesheet link:
https://fonts.googleblog.com/2010/09/optimizing-use-of-google-font-api.html


### Do we need to be worried about this?
 
I would say not so much.

This is a performance problem of a very low order - not only does it only affect first startup time, it only affects first startup if these very common fonts haven't been loaded yet.

To compound on that, the CSS requests are made in parallel:
Screen Shot 2020-06-29 at 7 07 39 pm

And the fonts are mostly loaded in parallel:
Screenshot on 2020-06-29 at 19-14-22

This can be fixed to become all fonts:
https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization

On top of all that, this is one of the easier things to fix at a later stage.

---

This is a rather long comment for something like this, but I wrote it mainly to put a spotlight on the fact that this seems like a performance issue.
**In general I think that this sort of performance problem shouldn't be in our scope.**

utils/material-table.ts
aaronyan: I'm not too opinionated on this. I'd say leave on left for now in preparation for the future filters options. If we feel it's too distracting, we can make the search a lighter grey to de-emphasize the search.

components/ExperimentStatus.tsx
aaronyan: Statuses having uppercase and colors I think is a valuable design distinction from the rest of the items in the table, so imo I think we should keep it.

Yeah, the background color is quite hard to see, so let's change it to be 100% opacity of the status background, and then white text. Also let's change to a darker orange like `rgba(255, 123, 0, 1)`. And then it would look like this:

Screen Shot 2020-06-29 at 1 07 50 PM

aaronyan:
Thanks for your work on this @jessie-ross. It's looking really good! We [may end up choosing a paid theme](https://github.com/Automattic/abacus/issues/80) (like [this one](https://material-ui.com/store/items/devias-kit-pro/)), so you won't have to worry too much atm on polishing specific to the mocks. But I appreciate getting the other items in line like the column ordering, the tests, and using the `ExperimentStatus` component.

components/ExperimentsTable.test.tsx
dhurlburtusa: Oops! I saw the `<ThemeProvider>` and immediately thought this was a story. My bad.

dhurlburtusa:
@aaronyan I was mistaken about the custom MUI themes They are more like web app templates than simple skinning themes. It might be easier to look a custom theme and then update the custom theme we started to match the look.

components/ExperimentStatus.tsx
aaronyan: I just caught up on [this issue](https://github.com/Automattic/abacus/issues/164), and I like the idea of "highlighting text". Here's a new recommendation!
- Change `running` color to even darker orange: `rgb(255, 106, 0)` /  `#ff6a00`
- Change `completed` color to even darker green: `rgb(9, 173, 75)` / `#09ad4b`
- Change status to use `font-family: "Roboto Mono", monospace` to be a "highlighted text"
- Increase highlight to `0.15` opacity
- Keep uppercase

Screen Shot 2020-06-29 at 5 55 56 PM

aaronyan:
Ah I see. Thanks for the heads up and info @dhurlburtusa! Too bad styling and themes are not as easy as I initially thought, but all good. Sounds like that polish will very useful then! :)

components/ExperimentStatus.tsx
yanirs: Hmm... I don't get why we need three types of highlight for the status:
* Text colour
* Background colour
* Uppercasing

I don't mind the first two, but uppercasing seems excessive. Also, colouring the text [doesn't seem to be in line with the Material Design guidelines](https://material.io/design/color/text-legibility.html#text-backgrounds):

> Black text is recommended for use on light backgrounds, and white text on dark backgrounds. If your app has both light and dark themes, make sure the text is available in a contrasting color against each theme.
>
> Colored backgrounds or typography additionally change the rules regarding text opacity and different states of text.

How about we stick to a single form of highlight: Background colour + appropriate white/black text for legibility?

pages/_document.tsx
yanirs: Alright, let's defer this. I agree that it's not a priority for now. Added #174 for future reference.

components/ExperimentsTable.tsx
jessie-ross: I might do it for now to match all the columns.

I will add it on the `DateTimeText` component.

jessie-ross:
Just to summarise the changes for this PR:
- Move monospacing to the `DateTimeText` component
- Change the ExperimentStatus style as per #164

jessie-ross:
This is the latest look:
Screen Shot 2020-07-01 at 1 44 01 pm

I have also wrapped the test render function so it automatically adds ThemeProvider, it was becoming a problem as one component would contaminate others on top.

components/DateTimeText.tsx
```diff
@@ -1,18 +1,21 @@
...
+  }),
+)
 
 const ISO_DATE_LENGTH = 10
 
 /**
  * Renders the date value in ISO 8601 format UTC.
  */
-const DatetimeText = ({ datetime, excludeTime }: { datetime: Date; excludeTime?: boolean }) => {
+const DateTimeText = ({ datetime, excludeTime }: { datetime: Date; excludeTime?: boolean }) => {
```
yanirs, line 18: For consistency, can we please keep `DatetimeText`? We use `datetime` as a single word with no separators throughout the code, including on this line.

In general, it's better to make such refactoring changes separate to keep the PR diffs clean.

helpers/render.tsx
```diff
@@ -0,0 +1,9 @@
+import { Queries, render as actualRender, RenderOptions } from '@testing-library/react'
+import React from 'react'
+
+import ThemeProvider from '@/styles/ThemeProvider'
+
+const render: typeof actualRender = <Q extends Queries>(ui: React.ReactElement, options?: RenderOptions<Q>) =>
```
yanirs, line 6: Shall we add this to `helpers/test-utils.ts` instead? It'd also be good to add a docstring explaining why it's needed.

yanirs:
> This is the latest look:

Looks good! By the way, I usually look at the auto-deployed Vercel instance (https://abacus-git-add-style-experiments-table.a8c.vercel.app/experiments in this case), so there's no need to include screenshots. :slightly_smiling_face: 

> I have also wrapped the test render function so it automatically adds ThemeProvider, it was becoming a problem as one component would contaminate others on top.

Sounds good. See inline comment about moving it to `test-utils.ts` (available on `master`).

Can you also please resolve conflicts with `master` and address the other two comments (new: https://github.com/Automattic/abacus/pull/170#discussion_r448160592 and old: https://github.com/Automattic/abacus/pull/170#discussion_r446723533)? I find it useful to resolve conversations on PRs to keep track of what needs addressing &ndash; I just resolved most of them. :slightly_smiling_face: 

I think we should be good to go after that. Good work!

components/DateTimeText.tsx
jessie-ross: Yep makes sense, I will revert this.

And you are right, I'll do this kind of thing as a seperate PR in the future.

aaronyan:
Woohoo, that looks so slick. Nice work y'all!

jessie-ross:
> Looks good. Can you please resolve this comment before merging: [#170 (comment)](https://github.com/Automattic/abacus/pull/170#discussion_r446723533)?

Ah thanks, I missed that one

Issue #162, opened 2020-06-25
Add metric details page


aaronyan:
A page to view metric details:
- Displays rich details of metric (resolves #149)
- Allow inline editing (resolves #14)

Let's follow the same design as the experiment details:
- Metric: <metric_name> at the top left
- A single General card with all of the details shown in expanded metric experiment page view
- Leaving out the edit button until we support metric editing

<img width="512" alt="experiments-view-with-tabs-and-metric-details" src="https://user-images.githubusercontent.com/4505888/85796786-61a71400-b6ef-11ea-809a-b9100dd4d966.png">

For future iterations, we can consider a card below General showing all of the experiments that use this metric (needs API endpoint).

yanirs:
> Metric name at top left

Should we make it `Metric: <metric_name>`? We currently have `Experiment: <experiment_name>` for experiment pages. This is useful because it's also used for the page title and it's not always clear from the name if it's a metric or an experiment.

> Edit button top right

Shall we leave this out until we support editing?

> For future iterations, we can consider a card below General showing all of the experiments that use this metric (needs API endpoint).

:+1:

aaronyan:
Thanks for the suggestions. Accepted all and edited the original description per your suggested changes:
- make it Metric: <metric_name>
- leave out edit button until we support metric editing

jessie-ross:
I have sort of implemented this already using the grid expansion feature.
Do we need this implemented further, would it make sense to wait for needing edit capabilities?

aaronyan:
Yeah, I think the way you've implemented the grid expansion feature is good enough for now. +1 let's wait for a need to edit before we add a metric details page.

I'd like to see the grid expansion in the Experiment Overview tab, but I'll create a separate issue for that. Lowering the priority on this issue for now.

jessie-ross:
Especially since this is only going to be accessible by "Admin"s (us), I recommend doing this by a single modal that is used for both adding and editing:
- Edit button in the metric rows.
- Add button under the table.

yanirs:
If we put it on the `/metrics` page then it'll be accessible to everyone, no?

jessie-ross:
I thought we were going to be putting this behind some kind of wall, might just be a request parameter.

yanirs:
Ah, makes sense. :slightly_smiling_face: No objections to doing it as a modal if we hide it behind a request parameter.

aaronyan:
[...]

PR #173, opened 2020-06-28
Implement MetricDetails

jessie-ross:
## ~~This is a WIP PR welcoming suggestions.~~

**A minimalist implementation of MetricDetails:**

Screen Shot 2020-06-29 at 1 13 16 am

- Uses the MaterialTable DetailsView feature
- Prints the parameters as JSON, I think this is reasonable for the MVP.

### Known issues to be fixed in PR:
- Testing
- Styling 
  - I know it doesn't look pretty now, but it can be polished up real nice. 
  - Alignment, background and scroll-overflow.

## How has this been tested?
- AT & MT


components/MetricsTable.tsx
```diff
@@ -14,22 +15,73 @@ const tableColumns = [
...
+  return (
+    <>
+      {isLoading && <LinearProgress />}
+      <Snackbar open={!!error} message='Oops! Something went wrong while trying to load a Metric.' />
+      {!isLoading && !error && metricFull && (
+        <dl className={classes.root}>
+          <Typography component='dt'>Direction:</Typography>
+          <Typography component='dd'>{metricFull.higherIsBetter ? 'Higher is better.' : 'Lower is better.'}</Typography>
+          <Typography component='dt'>Parameters:</Typography>
+          <Typography component='dd'>{paramsStringified}</Typography>
```
yanirs, line 63: I think it'd be better to render the parameters as more user-friendly text or not render them at all. As noted in #162, the idea is to make it a standalone page, so we have more space to play with, though I'm not too opposed to using the detail panel.

components/MetricsTable.tsx
```diff
@@ -14,22 +15,73 @@ const tableColumns = [
...
+  const paramsStringified = params && JSON.stringify(params, null, 4)
+
+  return (
+    <>
+      {isLoading && <LinearProgress />}
+      <Snackbar open={!!error} message='Oops! Something went wrong while trying to load a Metric.' />
+      {!isLoading && !error && metricFull && (
+        <dl className={classes.root}>
+          <Typography component='dt'>Direction:</Typography>
+          <Typography component='dd'>{metricFull.higherIsBetter ? 'Higher is better.' : 'Lower is better.'}</Typography>
```
yanirs, line 61: Can make this `Higher is better?` with `Yes/No` in the `dd`, and then we can use the `formatBoolean()` util.

components/MetricsTable.tsx
```diff
@@ -14,22 +15,73 @@ const tableColumns = [
...
 
+const useMetricDetailStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      display: 'grid',
+      gridTemplateColumns: '10rem auto',
+      gridRowGap: theme.spacing(1),
+      padding: theme.spacing(1, 4),
+      '& dt': {
+        fontWeight: 600,
```
yanirs, line 26: As noted on the other PR, this 600 should probably be a constant somewhere. :slightly_smiling_face:

components/MetricsTable.tsx
jessie-ross: Yeah I think it is worth doing it like this until other higher priority items are ticked off.
(If this were to load into its own page it would look very empty at the moment.)

jessie-ross:
I am holding on this one until #170 goes through, as I can apply a lot of that here.

jessie-ross:
The latest look https://d.pr/v/Ca1xfS (I do like including these...)

- In a sub-grid with hover-colour background.
- Contained the JSON visually with a border and an `overflow: scroll`.

components/MetricsTable.tsx
```diff
@@ -1,35 +1,141 @@
...
 import { defaultTableOptions } from '@/utils/material-table'
 
 const debug = debugFactory('abacus:components/MetricsTable.tsx')
 
-const tableColumns = [
-  { title: 'Name', field: 'name' },
-  { title: 'Description', field: 'description' },
-  { title: 'Parameter Type', field: 'parameterType' },
-]
+/* istanbul ignore next */
```
yanirs, line 25: Having so many coverage suppression comments defeats the purpose of checking for coverage. I'm happy to reopen the testing discussion on P2, but if the entire file is ignored, let's suppress on the config level rather than sprinkle suppression comments everywhere.

templates/component.tsx
```diff
@@ -7,6 +7,7 @@
 import React from 'react'
 import { makeStyles, createStyles, Theme } from '@material-ui/core/styles'
 
+/* istanbul ignore next */
```
yanirs, line 10: Is this meant to be in this PR?

yanirs:
> The latest look https://d.pr/v/Ca1xfS (I do like including these...)

Nice video! Looks good. I suppose it's no longer WIP?

I'm not a fan of suppressing all coverage, but I see how this would be tricky to test well now that we rely more heavily on MUI components and MaterialTable magic. I'm happy to suppress for now, but as noted, let's do it once per file rather than sprinkle warnings everywhere, and aim to have the discussion around testing and coverage moving forward early next week.

templates/component.tsx
jessie-ross: Yep, code coverage is complaining about the function being created here, I couldn't find a way to turn it off

components/MetricsTable.tsx
jessie-ross: Yep, code coverage is complaining about the style function. I think they shouldn't be tested.

It is just this function and not the whole file, e.g. with the current decision around testing, if another function were made we should be testing that.

yanirs: There are three `istanbul ignore next` comments in this file, which seem to pretty much cover everything.

jessie-ross:
> Nice video! Looks good. I suppose it's no longer WIP?

Good point, I have just updated the title.

> I'm not a fan of suppressing all coverage

Firstly not all coverage is being suppressed at the moment, e.g. this component:
https://github.com/Automattic/abacus/blob/f0f11ce2ce5caa2c4ac9abe5a638877b9c24790b/components/MetricsTable.tsx#L97-L100

Secondly, the parts we are ignoring fall into three specific categories:
- Trivial and unnecessary - ignoring the theme injecting function in makeStyles.
   I couldn't figure out a way to permanently ignore all of these.
- Incredibly trivial functions that are covered by e2e, e.g. these two arrow functions:
https://github.com/Automattic/abacus/blob/f0f11ce2ce5caa2c4ac9abe5a638877b9c24790b/components/MetricsTable.tsx#L129-L138
- Side-effectful data loading component that is covered by e2e, e.g. this fella:
https://github.com/Automattic/abacus/blob/f0f11ce2ce5caa2c4ac9abe5a638877b9c24790b/components/MetricsTable.tsx#L53-L56

I am turning off coverage for these specific purposes but I don't necessarily want to turn it off in general, as it fits to have it on with this codebase.

templates/component.tsx
dhurlburtusa: This is bizarre. We are already explicitly excluding all matching files in the `templates/` directory. See `jest.config.js`, line 14. We shouldn't need this.

yanirs: I tried removing it locally and I'm not seeing any issues.

components/MetricsTable.tsx
yanirs: I played with it a bit locally, and simply adding a click on a row to the unit tests obviates the need for all these coverage suppression comments, and moves the attention to things that aren't tested but should be.

yanirs:
> Firstly not all coverage is being suppressed at the moment, e.g. this component:

Ah yeah, I see that now.

> Secondly, the parts we are ignoring fall into three specific categories:
>
> Trivial and unnecessary - ignoring the theme injecting function in makeStyles.
> I couldn't figure out a way to permanently ignore all of these.

We shouldn't need to permanently ignore all these. It's easy to cover them by rendering the component. As noted inline, in this case adding a row click to the existing test would render the `MetricDetail` component. Alternatively, we can pull it out and render it separately.

> Incredibly trivial functions that are covered by e2e, e.g. these two arrow functions:

These functions are very trivial to cover by adding a row click to the test. :slightly_smiling_face: I think it's worth doing in this case.

> Side-effectful data loading component that is covered by e2e, e.g. this fella:

There's plenty of behaviour in `MetricDetail` that I think is worth testing. It shouldn't be too hard to do if we mock the data loading line.

jessie-ross:
Ah ok I see what you mean, I am not too familiar with Unit testing React Components, I will add the click to the unit test.

jessie-ross:
> There's plenty of behaviour in MetricDetail that I think is worth testing. It shouldn't be too hard to do if we mock the data loading line.

@yanirs, it seems like all the rest of the data loading is currently being done in `/pages` with only e2e applied so I can't seem to see any examples of mocking api hooks, I am going to look into a good way to do this.

jessie-ross:
This is actually pulling on a long thread:
- Mocking the Api call wasn't bad
- Getting the fixture data is complicated/lengthy


I didn't have a MetricFull fixture so I went to create one. I wanted to extend from MetricBare, so I thought about using the MetricFull constructor, but I couldn't use it directly as I would have to strip the methods first. The I thought about using  MetricBare.toApiData and MetricFull.fromApiData together, and I found out toApiData doesn't exist on MetricBare. I also wanted to keep it consistent with the rest of the code in the file which was just using Model constructors. At this point I was like this isn't worth it and copy-pasted it in.

Then I realised to mock the Api call I needed api data, so I was thinking I would just use MetricFull.toApiData, which didn't exist. At this point I thought it would be a good idea to write it so I did, but then I realised I would have to do the same for the children entities, so I did that.

Ok, so I thought I was all good at this point but nope, the code coverage doesn't pass now. I didn't want to do half assed testing if I was going to be testing these but these are lengthy to test well because of all the variations so I decided I should stop here.

I would say #204 would fix a lot of these problems.
- Integration testing assuring the schemas
- Only need one set of fixtures - the server-side data

Then it is easy to extend/convert the fixtures, and they are all flowing in one direction from the source data.

jessie-ross:
I've realised this morning that the mocked data doesn't need to be in ApiData format 🤦

components/MetricsTable.tsx
```diff
@@ -1,35 +1,141 @@
...
-    }
-    router.push('/metrics/[id]', `/metrics/${rowData.metricId}`)
-  }
+  const theme = useTheme()
+  const tableColumns = [
+    {
+      title: 'Name',
+      field: 'name',
+      cellStyle: {
+        fontFamily: theme.custom.fonts.monospace,
```
yanirs, line 110: Doesn't matter for this PR, but can't we just apply the monospace font higher up in the CSS? This is getting a bit repetitive.

models/Event.ts
```diff
@@ -12,7 +12,7 @@ export class Event {
   /**
    * Event properties that further constrain this event selection.
    */
-  public readonly props?: { [key: string]: string }
+  public readonly props?: { [key: string]: string | boolean }
```
yanirs, line 15: This should be a mapping from string to string since that's how `eventprops` are stored by Tracks.

models/MetricRevenueParams.ts
```diff
@@ -4,6 +4,11 @@ import { ApiData } from '@/api/ApiData'
  * Parameters for a revenue query. If `null`, then `event_params` must be given.
  */
 export class MetricRevenueParams {
+  /**
+   * Description
+   */
+  public readonly description?: string
```
yanirs, line 10: Why was this added? We have descriptions on the metric level, not on the params level.

jessie-ross: This was [based off of the swagger docs](https://app.swaggerhub.com/apis/yanir/experiments/0.1.0#/MetricFull), if you expand the revenueParams section.

models/Event.ts
jessie-ross: [Also based off of swagger](https://app.swaggerhub.com/apis/yanir/experiments/0.1.0#/management/get_metrics__metric_id_), if you look at the 200 example.

jessie-ross:
I'll make the schema changes you mentioned and add an issue for now about updating the swagger docs.

I will likely make the change to the swagger docs along with changes needed for using Schema in the codebase.

models/MetricRevenueParams.ts
yanirs: Looks like it's just the way Swagger UI formats things:

image

`description:` is slightly greyed out, indicating that it's _not_ a property. It's clearer when looking at the yaml:

```yaml
revenue_params:
  type: object
  description: >
    Parameters for a revenue query. If `null`, then
    `event_params` must be given.
  nullable: true
  properties:
    refund_days:
      type: integer
      description: Number of days to use for the refund window.
      format: int32
...
```

They should really render it like they render the array description. It's pretty confusing.

models/Event.ts
yanirs: Ah yeah, that's a bug in the conversion metric example, so it should be fixed in the Swagger file.

components/MetricsTable.tsx
jessie-ross: I've tried to do so but it seems to be a limitation of MaterialGrid, part of my motivation for #205

models/MetricRevenueParams.ts
jessie-ross: Ahhh I see

PR #177, opened 2020-07-01
Add Experiment Creation page

jessie-ross:
Adds experiment creation page.
- Loads necessary initial data: metrics, segments
- Adds `newExperimentTemplate` for new experiments
- Adds "for now" link in the top nav

Part of a series on #9 

## How has this been tested?
- MT
- e2e tests to come later

### DIFF
--- a/models/ExperimentFull.ts
+++ b/models/ExperimentFull.ts
@@ -218,3 +218,22 @@ export class ExperimentFull implements ApiDataSource {
     return !!this.endReason || !!this.conclusionUrl || typeof this.deployedVariationId === 'number'
   }
 }
+
+export const newExperimentTemplate = {
+  experimentId: null,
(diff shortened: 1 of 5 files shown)


models/ExperimentFull.ts
```diff
@@ -218,3 +218,22 @@ export class ExperimentFull implements ApiDataSource {
     return !!this.endReason || !!this.conclusionUrl || typeof this.deployedVariationId === 'number'
   }
 }
+
+export const newExperimentTemplate = {
```
dhurlburtusa, line 222: To minimize the chance that this mutable object with some mutable properties (metricAssignments, etc), I'd recommend making this a function that returns this object. Because a function will return a new object each time, there won't be any issue with accidentally mutating the template itself.

PR #178, opened 2020-07-01
Add Experiment Creation form UI for the Beginning section

jessie-ross:
This PR implements the form-part for the Beginning section.

This is also intended to set a template for the other form-parts.
- One PR for each form-part
- Note the file structure
- Storybook-ed
- Expect unpolished UI to come out, polishing to happen at a later stage
- Tests come at the end of the PR

Part of a series on #9 

## The Details

Screen Shot 2020-07-02 at 2 00 33 am

Having said that this is supposed to be unpolished, I couldn't help myself with this one:
- Couldn't get the text to come out nice like on Mocks, ended up with widows so I moved the text to two lines.
- Changed the order of the first two lines, providing motivation before prompting the user to create a P2.
- Ensmallened the Begin button, it didn't seem right in material style to be at the mock's size.

How do you feel about me making these sorts of design changes? I can build more directly from the mocks.

I omitted the P2 references, are we including those?  

## How has this been tested?
- MT
- AT


components/experiment-creation/form-parts/Beginning.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const classes = useStyles()
+
+  return (
+    <div className={classes.root}>
+      <Typography variant='body1'>
+        We think one of the best ways to prevent a failed experiment is by documenting what you hope to learn.
+      </Typography>
+      <Typography variant='body1' className={classes.p2LinkLine}>
+        To create a new experiment, please first <Link href='#TODO'>post on this P2</Link>.{' '}
+      </Typography>
```
yanirs, line 39: I think we can just drop this sentence. We will have a Field Guide page with full instructions, the other sentences provide enough context, and it'll solve the link problem.

yanirs:
Good start, thanks for pushing the WIP PR!

> This is also intended to set a template for the other form-parts.
> 
>     * One PR for each form-part
>     * Storybook-ed
>     * Expect unpolished UI to come out, polishing to happen at a later stage

Sounds good. :slightly_smiling_face: 
 
>     * Note the file structure

I agree with the need for the `/components/experiment-creation` directory, as `/components` is now too busy to be flat, but I wonder if we need the `/form-parts` subdirectory for now? Having too many levels of nesting can make navigation a bit of a pain. How about we add `/form-parts` later if `/experiment-creation` becomes too busy?

>     * Tests come at the end of the PR

Do you mean at the end of each PR, but not necessarily at the WIP stage? If so, sounds good.

> Having said that this is supposed to be unpolished, I couldn't help myself with this one:
> 
>     * Couldn't get the text to come out nice like on Mocks, ended up with widows so I moved the text to two lines.
>     * Changed the order of the first two lines, providing motivation before prompting the user to create a P2.
>     * Ensmallened the Begin button, it didn't seem right in material style to be at the mock's size.
> 
> How do you feel about me making these sorts of design changes? I can build more directly from the mocks.

Sounds good, especially for experiment creation, as the design predates the decision to go with Material UI. Also, it's natural to come across design issues when doing the implementation. At this point it's up to our team to make design decisions, as we don't have a dedicated designer.

> I omitted the P2 references, are we including those?

See inline comment. I think we can drop that line, as we will provide additional context in the Field Guide, which is easier to quickly edit than the code.

jessie-ross:
> I agree with the need for the /components/experiment-creation directory, as /components is now too busy to be flat, but I wonder if we need the /form-parts subdirectory for now? Having too many levels of nesting can make navigation a bit of a pain. How about we add /form-parts later if /experiment-creation becomes too busy?

Sure

> Do you mean at the end of each PR, but not necessarily at the WIP stage? If so, sounds good.

Yep that's what I meant

> See inline comment. I think we can drop that line, as we will provide additional context in the Field Guide, which is easier to quickly edit than the code.

Easy

components/experiment-creation/Beginning.test.tsx
```diff
@@ -0,0 +1,8 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+
+import Beginning from './Beginning'
+
+test('renders without crashing', () => {
```
yanirs, line 6: Can we make this a snapshot test? It may feel silly now, but it becomes informative with subsequent diffs.

PR #179, opened 2020-07-01
Add Experiment Toolbar

### LINKED ISSUE #8: Add experiment details page
Aaron edited: Let's make this details page the same as the second page in the New Experiment creation flow. We just need to add the p2 link to the "basic info" section underneath the experiment name.

When a user clicks on an experiment from the experiments list page, then they will be brought to this experiment details page. We'll likely have two tabs/sections on this page:

1. Analysis
2. Experiment details

We'll default to 1 if experiment has results, otherwise default to 2.

All the buttons will be greyed out (uneditable) until we have have the components to update the experiment details (#10).

dhurlburtusa:
Add Experiment Toolbar to manage the various "edit" modes during an experiment's lifecycle.

Resolves #8.

## How has this been tested?

Added new tests for updated components and new components.

Some interactions deferred to e2e tests.


components/ExperimentTabs.tsx
```diff
@@ -1,25 +1,75 @@
...
+    [theme.breakpoints.down(700)]: {
+      tab: {
+        minWidth: 90,
+      },
+    },
+    [theme.breakpoints.down(640)]: {
+      tab: {
+        minWidth: 110,
+      },
+    },
```
dhurlburtusa, line 23: These styles allow the tabs and toolbar buttons to remain side-by-side as much as possible until 640px or narrower. They take into account the widest the toolbar will be based on the possible combination of buttons that will be shown at any given moment.

components/ExperimentToolbar.test.tsx
```diff
@@ -0,0 +1,276 @@
...
+// | staging     | n/a          | details  | view     | yes
+// | n/a         | n/a          | results  | view     | yes
+// | n/a         | n/a          | snippets | view     | yes
+
+test('with a complete experiment, no conclusions, on details section, and in conclude mode will render expected buttons', () => {
+  const experiment = Fixtures.createExperimentFull()
+  const { container } = render(<ExperimentToolbar experiment={experiment} mode='conclude' section='details' />)
+
+  // Expect "Cancel" and "Save Conclusions" buttons.
+  expect(container).toMatchSnapshot()
```
dhurlburtusa, line 34: Doing explicit checks with the "testing library" will likely be less brittle than using snapshots. The snapshots are about 1700 lines long.

Doing explicit checks should only net to a couple lines more for each test but eliminate a brittle 1700 line long snapshot.

components/ExperimentToolbar.tsx
```diff
@@ -0,0 +1,148 @@
...
+    root: {
+      alignItems: 'center',
+      display: 'flex',
+    },
+    tabs: {
+      flex: '1 0 auto',
+    },
+    toolbarRoot: {
+      justifyContent: 'flex-end',
+      minHeight: 48,
```
dhurlburtusa, line 28: Please document why this value was chosen.

components/AnalysisSummary.tsx
```diff
@@ -210,6 +211,7 @@ export default function AnalysisSummary({
 
   return (
     <>
+      <ExperimentToolbar experiment={experiment} mode='view' section='results' />
```
yanirs, line 214: Why is this here? It should be on the page rather than the summary component.

dhurlburtusa:
@yanirs That sounds great. I'm trying to remember why I thought I had to embed the tabs and toolbar in each of the page's content components (i.e., ExperimentDetails and AnalysisSummary). I thought there was some heading or something that was special on the individual pages that went above the tabs and toolbar. But you're right, the tabs and toolbar can be outside. This will simplify a few things. Thanks for noticing.

components/ExperimentToolbar.test.tsx
```diff
@@ -0,0 +1,276 @@
...
+})
+
+test('with a disabled experiment, no conclusions, on details section, and in conclude mode will render expected buttons', () => {
```
jessie-ross, line 37: Can we do these tests generatively?

dhurlburtusa: @jessie-ross If I understand correctly, I've tried in the way past and ran into issues determining which test out of dozens were failing because they all ended up having the same line number as I iterated.

But it has been a long time since I tried that. So, I'll do a little research unless you already have something in mind.

jessie-ross: Yeah I am not too sure either, but I was thinking we should be able to iterate over the `test(` calls, generating test names for them.

components/ExperimentDetails.tsx
```diff
@@ -25,11 +26,73 @@ function ExperimentDetails({
   segments: Segment[]
 }) {
   debug('ExperimentDetails#render')
+  const [mode, setMode] = useState<ExperimentToolbarMode>('view')
```
jessie-ross: I like to use Enums for these, but types seem to work just as well

components/ExperimentToolbar.tsx
```diff
@@ -0,0 +1,148 @@
...
+  section: ExperimentToolbarSection
+}) {
+  const classes = useStyles()
+  const { status } = experiment
+  const hasConclusionData = experiment.hasConclusionData()
+  const concludable = (status === Status.Completed || status === Status.Disabled) && !hasConclusionData
+  const editable =
+    status === Status.Staging ||
+    status === Status.Running ||
+    ((status === Status.Completed || status === Status.Disabled) && hasConclusionData)
```
jessie-ross, line 84: I think these guys are pretty core business logic that should be moved out towards the model file.

styles/theme.ts
```diff
@@ -47,6 +47,14 @@ const theme = createMuiTheme({
         },
       },
     },
+    MuiButton: {
```
dhurlburtusa, line 50: This changes all MUI Buttons using the `primary` color and `contained` variant to be this green color. Without having mockups for the rest of the app, not sure if this is the best way to style them.

styles/theme.ts
```diff
@@ -47,6 +47,14 @@ const theme = createMuiTheme({
...
+      containedPrimary: {
+        backgroundColor: '#219653',
+        '&:hover': {
+          backgroundColor: '#006729',
```
dhurlburtusa, line 54: Took color from the mockup, set in in https://material.io/resources/color/, and created this color.

components/ExperimentToolbar.tsx
dhurlburtusa: @jessie-ross I started to move these to the model but then thought about it a little more. I think it is close to core business logic but I think it is more view/presentation logic.

**Core Business Logic:**

isConcludable => (status === Status.Completed || status === Status.Disabled)
// That is, only able to be concluded if in the correct status, regardless of whether conclusion data have been entered.

isEditable => true
// That is, something is always editable but what is editable depends on the status.

**View Logic:**

In the view, what is meant by `concludable` is that it is in a concludable state AND conclusion data has not been entered. This is because we don't want to show the Edit button in this state because we want to encourage the user to fill out the conclusion data.

**Combining the two for an updated view logic**

`concludable` => experiment.isConcludable && !hasConclusionData
`editable` => same as before

Did you have something different in mind?

yanirs:
I'm not sure about this approach of adding action-specific handlers and the idea of a `mode`. It's hard to judge without any actual implementation of the actions, but it seems to me like a unified flow for all the actions may make more sense. That is, encapsulate the actual action calls in the toolbar component, and have a single `onAction()` callback that will refresh the details. Alternatively, we can add the buttons to the `ExperimentDetails` component since this is where the editing will happen.

Also, as we want to start using the limited UI with experimenters in the coming weeks, adding buttons that aren't wired to anything may be confusing. Since @dhurlburtusa won't be around to complete this work, I'm inclined to close this PR until @jessie-ross is ready to work on it. Then @jessie-ross can pick up parts of this PR as needed.

> Resolves #8

As noted in https://github.com/Automattic/abacus/issues/8#issuecomment-653730901, I think that #8 was already resolved without this. :slightly_smiling_face: Let's work on individual edit actions incrementally.

components/ExperimentToolbar.tsx
jessie-ross: Makes sense that an experiment is always editable but we are just hiding it to encourage the users to conclude.

We should make that very explicit as it is not something a maintainer would know:
```javascript
const showConcludeButton = isConcludable(experiment) && !hasConclusionData
// An experiment always has editable parts but we want to hide the Edit Button when it needs to be concluded.
// This is to encourage users to click the conclude button.
const showEditButton = !showConcludeButton
```

- Change the semantics of the variable from being linked to the property of the experiment to the view by referring to the buttons instead of the property of the experiment.
- Have the logic refer to the showing of the conclusion button rather than properties of the experiment.
- No need to mention the other states

styles/theme.ts
jessie-ross: Sorry I didn't comment on this earlier, green should be our secondary colour and it shouldn't be hardcoded for this variant

Issue #98, opened 2020-06-04
Add analysis debugging page


yanirs:
With liberal use of warnings (#96) and the additional depth of our analysis and system, I’m pretty sure we’ll need a quick view of the full analysis results. Rather than running manual queries, we can "hide" this on a separate page (`/experiments/<id>/results/debug`?) where we’ll show a table with all the latest results, display all the plots, and maybe even include the full JSON. We won’t link to this page, but it should be the first place we check when an experiment is flagged for manual analysis.

PR #180, opened 2020-07-02
Move processing logic from AnalysisSummary to AnalysisProcessor

yanirs:
In preparation to addressing #96 and #98, this change moves some of the logic from the `AnalysisSummary` components to a standalone `AnalysisProcessor` helper class, making the components more focused on rendering the processed results.

## How has this been tested?

Technically, it's now possible to test `AnalysisProcessor` in isolation, but for this change I'm relying on the existing tests that still give us full coverage.

<strike>I also had some weird unrelated test failures that seem to be due to SwaggerHub failures. Hopefully it'll be fine on CircleCI.</strike> Yep, looks fine.

### DIFF
--- a/components/AnalysisSummary.tsx
+++ b/components/AnalysisSummary.tsx
@@ -13,6 +13,7 @@ import {
   RecommendationWarningToHuman,
   Variation,
 } from '@/models'
+import AnalysisProcessor from '@/utils/AnalysisProcessor'
 
 /**
  * Convert a recommendation's endExperiment and chosenVariationId fields to a human-friendly description.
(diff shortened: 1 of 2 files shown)


components/AnalysisSummary.tsx
```diff
@@ -86,94 +87,74 @@ function ParticipantCounts({
...
-                        </TableCell>
-                      </>
-                    ) : (
-                      <>
-                        <TableCell>N/A</TableCell>
-                        <TableCell>N/A</TableCell>
-                        <TableCell>Not analyzed yet</TableCell>
-                      </>
-                    )}
+      {analysisProcessor.resultSummaries.map(
```
yanirs, line 93: This diff is better viewed [ignoring whitespace](https://github.com/Automattic/abacus/pull/180/files?w=1).

utils/AnalysisProcessor.ts
```diff
@@ -0,0 +1,75 @@
...
+  public readonly resultSummaries: ResultSummary[]
+
+  constructor(analyses: Analysis[], experiment: ExperimentFull, metrics: MetricBare[]) {
+    this.experiment = experiment
+    this.metricAssignmentIdToLatestAnalyses = _.mapValues(
+      _.groupBy(analyses, 'metricAssignmentId'),
+      (metricAnalyses) => {
+        metricAnalyses = _.orderBy(metricAnalyses, ['analysisDatetime'], ['desc'])
+        return _.sortBy(
+          _.filter(metricAnalyses, ['analysisDatetime', metricAnalyses[0].analysisDatetime]),
```
jessie-ross, line 48: - I don't think this line belongs here.
- This line logically belongs with the `_.orderBy` as they work together to get the set of `latestMetricAnalyses`
- If I was feeling like it I might even put them in a function together for clarity.

components/AnalysisSummary.tsx
```diff
@@ -86,94 +87,74 @@ function ParticipantCounts({
...
-        return {
-          metricAssignmentId,
-          attributionWindowSeconds,
-          metricName: metricsById[metricId].name,
-          latestAnalyses: metricAssignmentIdToLatestAnalyses[metricAssignmentId as number],
-        }
-      },
-    )
-  }, [experiment.metricAssignments, metricsById, metricAssignmentIdToLatestAnalyses])
+function LatestResults({ analysisProcessor }: { analysisProcessor: AnalysisProcessor }) {
```
jessie-ross, line 90: The naming of `analysisProcessor` is a bit awkward, it is an analysis that has already been processed and not a processor.

jessie-ross: It is also plural as it is multiple processed analyses.

utils/AnalysisProcessor.ts
```diff
@@ -0,0 +1,75 @@
...
+          'analysisStrategy',
+        )
+      },
+    )
+
+    const metricsById = _.zipObject(_.map(metrics, 'metricId'), metrics)
+    this.resultSummaries = _.orderBy(
```
jessie-ross, line 55: IMO sorting should happen at the view level

utils/AnalysisProcessor.ts
```diff
@@ -0,0 +1,75 @@
...
+   */
+  public readonly experiment: ExperimentFull
+
+  /**
+   * A mapping from each of the experiment's metric assignments to its latest analyses.
+   *
+   * The mapped analyses are guaranteed to be ordered deterministically by analysisStrategy. Determinism is based on the
+   * assumption that there's a single analysis per analysisStrategy for a given analysisDatetime.
+   */
+  public readonly metricAssignmentIdToLatestAnalyses: { [key: number]: Analysis[] }
```
dhurlburtusa, line 31: This isn't being used externally. It can be private.

utils/AnalysisProcessor.ts
```diff
@@ -0,0 +1,75 @@
...
+  metricAssignmentId: number
+  attributionWindowSeconds: AttributionWindowSeconds
+  metricName: string
+  latestAnalyses: Analysis[]
+}
+
+/**
+ * A helper class to handle the processing of an experiment's analyses.
+ */
+export default class AnalysisProcessor {
```
dhurlburtusa, line 19: I agree with the awkwardness of this class' name. To me, the name implies a service type object that would have a process method that takes some input and would return the output.

Maybe separate into a data object that contains the results of the processing/aggregating and function that does the processing/aggregating. Basically, the constructor becomes a function that returns the data object which will be used by the view.

utils/AnalysisProcessor.ts
dhurlburtusa: Yes, the choice of which sort you want to display should happen at the view level. But I think it is okay to have a default sorting of the data. If the default sorting matches the choice of sorting for the view, then no more sorting needed. If the view needs a different order, then the view will need to re-order it itself.

utils/AnalysisProcessor.ts
jessie-ross: Had a bit more of a think about this overnight: The orderBy and filter combination is the only business logic I can see in this file, the rest of it wrangles the data into a format for the UI. I do think it should be separate and tested as a function/method `Analysis[] -> Analysis[]`.

utils/AnalysisProcessor.ts
```diff
@@ -0,0 +1,75 @@
...
+      },
+    )
+
+    const metricsById = _.zipObject(_.map(metrics, 'metricId'), metrics)
+    this.resultSummaries = _.orderBy(
+      experiment.metricAssignments,
+      ['isPrimary', 'metricAssignmentId'],
+      ['desc', 'asc'],
+    ).map(({ metricAssignmentId, attributionWindowSeconds, metricId }) => {
+      return {
```
jessie-ross, line 60: I am going to mention this in a P2, it is more extensible to avoid a custom merge and have this as a collection of the data:

```
{
    metricAssignment: MetricAssignment,
    metric: Metric,
    latestAnalyses: Analysis[],
}
``` 

This makes this structure general purpose and not tied to the interface, it also allows us to use common transforms on those data types.

It also reduces the number of custom collection types.

components/AnalysisSummary.tsx
yanirs: A processor that processes multiple analyses is an analysis processor. :slightly_smiling_face: But yeah, I don't find the name great either.

utils/AnalysisProcessor.ts
yanirs: Yeah, I agree with @dhurlburtusa regarding default sorting. I was thinking we should have that for experiment variations, as it'd obviate the need to sort again everywhere variations are used.

utils/AnalysisProcessor.ts
yanirs: Yeah, the custom type made more sense as a one-off interface in the context of the view, as it reduces nesting.

yanirs:
Thanks @dhurlburtusa and @jessie-ross. I agree that the `AnalysisProcessor` isn't the best solution to simplifying the `AnalysisSummary` components, which is what I had in mind with this PR. With the switch to MaterialTables things are already simpler, and the end goal will probably be better served by standalone utility functions. I'll close this PR and follow a different approach for generating the simplified analysis table (#96).

utils/AnalysisProcessor.ts
jessie-ross: To explain my opinion a bit more: 
- It adds unnecessary code to functions that shouldn't be worrying about the view.
- If we want to see how it is sorted, or change how it is sorted we would need to go up the function tree to find out.
- When we receive a list of data down through the component tree its order isn't documented (and shouldn't be), so we should assume it is unsorted and sort before rendering.

yanirs: > When we receive a list of data down through the component tree its order isn't documented (and shouldn't be), so we should assume it is unsorted and sort before rendering.

I don't think this _has_ to be the case. If the order of a list is documented, then we can rely on this order. For example, if `experiment.variations` was always in the canonical documented order (default variation first, then by ID), then we'd be able to rely on `experiment.variations[0]` being the default variation and not need to sort it multiple times in different components. As it is, `experiment.variations` is part of the interface and it having a specific order can be a part of the interface as well.

The same goes for any processing function &ndash; it can guarantee an order as part of its contract, and it is often desirable as it makes code that relies on it more consistent and less repetitive.

PR #181, opened 2020-07-02
Change analysis tables to use MaterialTable

yanirs:
In preparation to addressing #96 and #98, this PR changes the two analysis tables to use `MaterialTable`. This makes it easier to focus on the table content that doesn't get drowned by nested markup. In addition, the upcoming addition of analysis warnings will rely on more advanced `MaterialTable` features, so I figured we should make things consistent.

## How has this been tested?

Manual testing in storybook and with the mock and production API &ndash; visually it's more or less the same as what we have on `master`, except that the column widths aren't as nice. However, we can easily tinker with it later and styling is far from "final" in this PR. Both tables will be moved to the debug mode view in the next PR anyway, where styling isn't too important.


components/__snapshots__/AnalysisSummary.test.tsx.snap
```diff
@@ -40,7 +40,10 @@ exports[`shows the analyses JSON in debug mode 1`] = `
         "wide_ci"
       ]
     },
-    "analysisDatetime": "2020-05-10T00:00:00.000Z"
+    "analysisDatetime": "2020-05-10T00:00:00.000Z",
+    "tableData": {
```
yanirs, line 44: Looks like `MaterialTable` mutates the data, but I don't think it matters for JSON display purposes.

components/AnalysisSummary.tsx
```diff
@@ -88,70 +79,60 @@ function ParticipantCounts({
...
-              {AttributionWindowSecondsToHuman[attributionWindowSeconds]}
-            </div>
-            <div>
-              <strong>Last analyzed: </strong>
+            <Typography variant={'subtitle1'}>
+              <strong>
+                <code>{metricName}</code>
+              </strong>{' '}
+              with {AttributionWindowSecondsToHuman[attributionWindowSeconds]} attribution, last analyzed on{' '}
               {DatetimeText({ datetime: latestAnalyses[0].analysisDatetime, excludeTime: true })}
```
jessie-ross, line 128: While you are at it, this should be run in jsx `<DatetimeText`

jessie-ross:
LGTM

dhurlburtusa:
I piggy-backed #185 on this PR to address a double border issue in MaterialTables (MUI Tables in general).

PR #182, opened 2020-07-02
Update ExperimentTabs to Use MUI Tabs

dhurlburtusa:
Update ExperimentTabs to Use MUI Tabs.

## How has this been tested?

Updated tests for component updates.
Manually tested.

### DIFF
--- a/components/ExperimentTabs.test.tsx
+++ b/components/ExperimentTabs.test.tsx
@@ -5,12 +5,6 @@ import { ExperimentFull, Platform, Status, Variation } from '@/models'
 
 import ExperimentTabs from './ExperimentTabs'
 
-test('renders nothing when experiment prop is `null`', () => {
-  const { container } = render(<ExperimentTabs experiment={null} />)
-
-  expect(container.hasChildNodes()).toBe(false)
-})
-
 test('renders expected links', () => {
   // TODO: Get from fixtures.
   const experiment = new ExperimentFull({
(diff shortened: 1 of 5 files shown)


components/ExperimentTabs.test.tsx
```diff
@@ -5,12 +5,6 @@ import { ExperimentFull, Platform, Status, Variation } from '@/models'
 
 import ExperimentTabs from './ExperimentTabs'
 
-test('renders nothing when experiment prop is `null`', () => {
```
dhurlburtusa, line 8: No longer allow a `null` experiment prop.

components/ExperimentTabs.test.tsx
```diff
@@ -42,8 +36,9 @@ test('renders expected links', () => {
     metricAssignments: [],
     segmentAssignments: [],
   })
-  const { getByText } = render(<ExperimentTabs experiment={experiment} />)
+  const { getByText } = render(<ExperimentTabs experiment={experiment} tab='details' />)
```
dhurlburtusa, line 39: Currently, the tests don't depend on the various values for `tab` prop.

components/ExperimentTabs.tsx
```diff
@@ -1,25 +1,65 @@
...
+import Tabs from '@material-ui/core/Tabs'
+import { useRouter } from 'next/router'
+import React, { ReactNode } from 'react'
 
 import { ExperimentFull } from '@/models'
 
+const useStyles = makeStyles(() =>
+  createStyles({
+    tab: {
+      minWidth: 110,
```
dhurlburtusa, line 12: The default min-width was too big and didn't look right. This is an arbitrary size to looks good.

components/ExperimentTabs.tsx
```diff
@@ -1,25 +1,65 @@
...
+    },
+  }),
+)
+
+function LinkTab({ as, label, url, value }: { as?: string; label: ReactNode; url: string; value: string }) {
+  const classes = useStyles()
+  const router = useRouter()
+  return (
+    <Tab
```
dhurlburtusa, line 21: MUI `Tab` and the Next.js `Link` don't work well together. So, needed a special component to handle the routing and to render as MUI intended.

styles/theme.ts
```diff
@@ -74,6 +74,9 @@ const theme = createMuiTheme({
     primary: {
       main: '#194661',
     },
+    secondary: {
```
dhurlburtusa, line 77: The "current tab"-indicator uses the theme's secondary color. So, used a color to match the mockup. At this point, I am not sure what other MUI components use this color. So, we may need to make some adjustments if this unexpectedly change's some other MUI component's color.

jessie-ross:
Looks fine for now.

PR #184, opened 2020-07-02
Adjust Heading Size and Spacing

### LINKED ISSUE #159: Adjust MUI's heading sizes to match the design
As discussed internally, let's:

* Adjust MUI's heading sizes so we can use `h1-6` with the current design.
* Change the `h1` in the `Layout` component to use `h1` variant.
* Change the `h3` in the `LabelValuePanel` to use `h2`? (unless we expect other higher-level headings on the page)

dhurlburtusa:
Adjust Heading Size and Spacing.

Resolves #159.

## How has this been tested?

Manually, visually tested.

### DIFF
--- a/components/Headings.stories.tsx
+++ b/components/Headings.stories.tsx
@@ -1,0 +1,15 @@
+import Typography from '@material-ui/core/Typography'
+import React from 'react'
+
+export default { title: 'Headings' }
+
+export const all = () => (
+  <>
+    <Typography variant='h1'>h1. Heading</Typography>
+    <Typography variant='h2'>h2. Heading</Typography>
+    <Typography variant='h3'>h3. Heading</Typography>
+    <Typography variant='h4'>h4. Heading</Typography>
+    <Typography variant='h5'>h5. Heading</Typography>
+    <Typography variant='h6'>h6. Heading</Typography>
+  </>
+)
--- a/components/Layout.tsx
+++ b/components/Layout.tsx
@@ -58,7 +58,7 @@ const useStyles = makeStyles((theme: Theme) =>
       flex: '1 0',
     },
     contentTitle: {
-      margin: theme.spacing(3, 0, 2, 0),
+      margin: theme.spacing(4, 0, 2, 0),
     },
 
     // Footer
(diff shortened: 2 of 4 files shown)


components/Headings.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import Typography from '@material-ui/core/Typography'
```
dhurlburtusa, line 1: Story added so we can see all headings together.

components/Layout.tsx
```diff
@@ -58,7 +58,7 @@ const useStyles = makeStyles((theme: Theme) =>
       flex: '1 0',
     },
     contentTitle: {
-      margin: theme.spacing(4, 0, 2, 0),
+      margin: theme.spacing(3, 0, 2, 0),
```
dhurlburtusa, line 61: The top margin looked excessive. Reduced a bit.

styles/theme.ts
```diff
@@ -21,6 +21,11 @@ declare module '@material-ui/core/styles' {
   }
 }
 
+const headingDefaults = {
```
dhurlburtusa, line 24: These defaults are borrowed from Bootstrap 4.

styles/theme.ts
```diff
@@ -75,6 +81,32 @@ const theme = createMuiTheme({
       main: '#194661',
     },
   },
+  typography: {
```
dhurlburtusa, line 84: These font-sizes are borrowed from Bootstrap 4.

components/Layout.tsx
jessie-ross: It is subjective but I felt that was a good amount of top margin, it gave a nice vertical rhythm.

yanirs: Yeah, this is also unrelated to #159 so let's keep it out of this PR.

```suggestion
      margin: theme.spacing(4, 0, 2, 0),
```

styles/theme.ts
yanirs: The sizes still feel a bit excessive, but we can iterate.

PR #185, opened 2020-07-02
Fix/double border

dhurlburtusa:
Fix Double Border in MaterialTables.

## How has this been tested?

Manually, visually confirmed double borders were removed from MaterialTables while maintaining the expected borders with raw MUI Tables.

### DIFF
--- a/styles/theme.ts
+++ b/styles/theme.ts
@@ -31,13 +31,13 @@ const theme = createMuiTheme({
   overrides: {
     MuiCssBaseline: {
       '@global': {
-        // Remove the last table cell border of a top-level MuiTable when in MuiPaper.
-        // Otherwise the paper's border butts up with the last table cell's border.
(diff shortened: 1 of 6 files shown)


styles/theme.ts
```diff
@@ -30,19 +31,13 @@ const theme = createMuiTheme({
   overrides: {
     MuiCssBaseline: {
       '@global': {
-        // Remove the last table cell border of a top-level MuiTable when in MuiPaper.
-        // Otherwise the paper's border butts up with the last table cell's border.
+        // Remove the last table cell border of a MuiTable when in MuiPaper. Otherwise the
```
yanirs, line 34: I'm not sure what's going on here and whether it's solving a problem worth solving. And why was the reference to https://github.com/mui-org/material-ui/pull/20809 removed?

dhurlburtusa:
Closing. May open a new PR based from master.

PR #187, opened 2020-07-03
Add and Use Label Component

### LINKED ISSUE #175: Use "Label" Component in Experiment Details
This is a followup/reminder to #164 to use the component we decide on with the experiment details (it is currently using a MUI Chip). #170 will address the component we decide on for the experiment table.

- Default variations.
- Excluded segments.
- Primary metrics.

dhurlburtusa:
Resolves #175.

## How has this been tested?

Manually observed it rendered in experiment details.
Component replaced the MUI Chip uses. Tests updated accordingly.

### DIFF
--- a/components/Label.tsx
+++ b/components/Label.tsx
@@ -1,0 +1,25 @@
+import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
+import clsx from 'clsx'
+import React, { ReactNode } from 'react'
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      backgroundColor: '#eee',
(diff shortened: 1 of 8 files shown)


components/Label.tsx
```diff
@@ -0,0 +1,25 @@
...
+      borderRadius: theme.spacing(2),
+      padding: theme.spacing(1, 1.5),
+    },
+  }),
+)
+
+/**
+ * A simple component for rendering and styling labeled text.
+ *
+ * Note: Don't confusion this with a label for an input control.
```
yanirs, line 18:
```suggestion
 * Note: Don't confuse this with a label for an input control.
```

PR #188, opened 2020-07-03
WIP: Add experiment schema

jessie-ross:
This PR adds experiment Schema and integrates it into the application.

This method provides minimal integration/refactoring.

For more maximal integration/factoring: see #204

### DIFF
--- a/api/ExperimentsApi.ts
+++ b/api/ExperimentsApi.ts
@@ -1,4 +1,7 @@
-import { ExperimentBare, ExperimentFull } from '@/models'
+/* eslint-disable @typescript-eslint/ban-ts-ignore */
+// This is temporary for the WIP,
+
+import { ExperimentBare, ExperimentFull, experimentFullSchema } from '@/models'
 
 import { ApiData } from './ApiData'
 import { fetchApi } from './utils'
@@ -30,8 +33,9 @@ async function findAll(): Promise<ExperimentBare[]> {
  *
  * @param id - The ID of the experiment to fetch.
  */
-async function findById(id: number) {
-  return ExperimentFull.fromApiData(await fetchApi('GET', `/experiments/${id}`))
+async function findById(id: number): Promise<ExperimentFull> {
+  // @ts-ignore: Temporary for the WIP PR, the inferred type perfectly matches the Interface except for the methods
+  return await experimentFullSchema.validate(await fetchApi('GET', `/experiments/${id}`))
 }
 
 const ExperimentsApi = {
--- a/models/ExperimentFull.ts
+++ b/models/ExperimentFull.ts
@@ -1,9 +1,14 @@
+import * as yup from 'yup'
+
 import { ApiData } from '@/api/ApiData'
 import { ApiDataSource } from '@/api/ApiDataSource'
 import { ExcludeMethods } from '@/types/ExcludeMethods'
 import { formatIsoUtcOffset } from '@/utils/formatters'
 
-import { Event, ExperimentBare, MetricAssignment, Platform, SegmentAssignment, Status, Variation } from './index'
+import { Event, ExperimentBare, MetricAssignment, SegmentAssignment, Variation } from './index'
+import { Platform } from './Platform'
+// These are seperate because of an import bug
+import { Status } from './Status'
 
 /**
  * An experiment with full data.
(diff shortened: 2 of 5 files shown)


api/ExperimentsApi.ts
```diff
@@ -30,8 +33,9 @@ async function findAll(): Promise<ExperimentBare[]> {
  *
  * @param id - The ID of the experiment to fetch.
  */
-async function findById(id: number) {
-  return ExperimentFull.fromApiData(await fetchApi('GET', `/experiments/${id}`))
+async function findById(id: number): Promise<ExperimentFull> {
+  // @ts-ignore: Temporary for the WIP PR, the inferred type perfectly matches the Interface except for the methods
+  return await experimentFullSchema.validate(await fetchApi('GET', `/experiments/${id}`))
```
dhurlburtusa, line 38: Please take a look at `types/ExcludeMethods.ts`. Maybe it will help this this TS issue.

models/ExperimentFull.ts
```diff
@@ -1,9 +1,14 @@
...
+
 import { ApiData } from '@/api/ApiData'
 import { ApiDataSource } from '@/api/ApiDataSource'
 import { ExcludeMethods } from '@/types/ExcludeMethods'
 import { formatIsoUtcOffset } from '@/utils/formatters'
 
-import { Event, ExperimentBare, MetricAssignment, Platform, SegmentAssignment, Status, Variation } from './index'
+import { Event, ExperimentBare, MetricAssignment, SegmentAssignment, Variation } from './index'
+import { Platform } from './Platform'
+// These are seperate because of an import bug
```
dhurlburtusa, line 10: This is curious. The `Platform` and `Status` worked before. If this will be required after the WIP, then updating the comment to point to the reported import bug may prove useful.

Update: After further research, it's probably best not to import from `./index` from any module within `models` -- only from outside `models`. I didn't trace it completely but in my experience when something if undefined in a group of related dependencies, then there usually is a circular dependency somewhere.

Note: I moved the schemas created/defined in this module to a separate `models/schemas.ts` module and reverted the separate imports, and I no longer ran into the Status undefined issue.

Note: I did not include `schemas.ts` in `models/index.ts`. (I didn't try either. So, used: `import { experimentFullSchema } from '@/models/schemas'` in `api/ExperimentsApi.ts`.

models/ExperimentFull.ts
```diff
@@ -218,3 +223,58 @@ export class ExperimentFull implements ApiDataSource {
     return !!this.endReason || !!this.conclusionUrl || typeof this.deployedVariationId === 'number'
   }
 }
+
+// Placing these here for now, but they should go in a schema.ts file
+//
+
+const idSchema = yup.number().integer().positive()
+
+export const metricAssignmentSchema = yup
```
dhurlburtusa, line 232: I think it would be best to hold off on exporting `metricAssignmentSchema`, `segmentAssignmentSchema`, and `variationSchema` (all internal use only, atm) until they are actually needed externally.

models/ExperimentFull.ts
```diff
@@ -218,3 +223,58 @@ export class ExperimentFull implements ApiDataSource {
     return !!this.endReason || !!this.conclusionUrl || typeof this.deployedVariationId === 'number'
   }
 }
+
+// Placing these here for now, but they should go in a schema.ts file
+//
+
+const idSchema = yup.number().integer().positive()
```
dhurlburtusa, line 230:
```suggestion
const idSchema = yup.number().integer().min(0)
```
I am not sure where the IDs actually start in the database, but I know that the mock API is returning 0 for `ExperimentFull#deployedVariationId`.

models/ExperimentFull.ts
```diff
@@ -218,3 +223,58 @@ export class ExperimentFull implements ApiDataSource {
...
+    description: yup.string().defined(),
+    startDatetime: yup.date().defined(),
+    endDatetime: yup.date().defined(),
+    status: yup.string().oneOf([Status.Staging, Status.Running, Status.Completed, Status.Disabled]).defined(),
+    platform: yup.string().oneOf([Platform.Wpcom, Platform.Calypso]).defined(),
+    ownerLogin: yup.string().defined(),
+    existingUsersAllowed: yup.boolean().defined(),
+    p2Url: yup.string().url().defined(),
+    endReason: yup.string().defined().nullable(),
+    conclusionUrl: yup.string().url().defined().nullable(),
```
dhurlburtusa, line 273: The mock API is just returning `string` as an example URL which obviously does not validate. #116 is keeping track of the issues/limitations of the mock API.

Seems we'll need to come up with a solution to this soon than later. Otherwise yup is going to complain about the invalid data being returned.

models/ExperimentFull.ts
```diff
@@ -218,3 +223,58 @@ export class ExperimentFull implements ApiDataSource {
...
+export const metricAssignmentSchema = yup
+  .object({
+    attributionWindowSeconds: yup.number().integer().positive().defined(),
+    changeExpected: yup.bool().defined(),
+    experimentId: idSchema.defined(),
+    isPrimary: yup.bool().defined(),
+    metricId: idSchema.defined(),
+    minDifference: yup.number().defined(),
+  })
+  .defined()
```
dhurlburtusa, line 241: After some experimentation, looks like we should append `.cameCase()` for `metricAssignmentSchema`, `segmentAssignmentSchema`, and `variationSchema`.

models/ExperimentFull.ts
```diff
@@ -218,3 +223,58 @@ export class ExperimentFull implements ApiDataSource {
...
+    isExcluded: yup.bool().defined(),
+  })
+  .defined()
+
+export const variationSchema = yup
+  .object({
+    experimentId: idSchema.defined(),
+    name: yup.string().max(128).defined(),
+    isDefault: yup.bool().defined(),
+    allocatedPercentage: yup.number().integer().min(1).max(99).defined(),
```
dhurlburtusa, line 256: Mock API is returning 0 for allocated percentage.

api/ExperimentsApi.ts
jessie-ross: Yeah I put it like this so we can start fixing the integration problems straight away.

models/ExperimentFull.ts
jessie-ross: Nice work looking into this, I wasn't too worried about it at this point.

models/ExperimentFull.ts
jessie-ross: We will need them fairly imminently as part of form validation.

models/ExperimentFull.ts
jessie-ross: Good point, I have committed it :)

models/ExperimentFull.ts
jessie-ross: Yep this is the result of us using more specific integration testing.

models/ExperimentFull.ts
jessie-ross: I got this from our table definition:
> `allocated_percentage` must be between 1 and 99.

Another thing to update on our Mock API

models/ExperimentFull.ts
jessie-ross: It does seem like positive should be right as MariaDB auto-increments starting from 1.

jessie-ross:
As far as the schema changes, I don't think we should be changing our schema to fit our Mock Data.
This is what integration tests/run-time data verification is for, to check the Apps assumptions on the data coming in.

models/ExperimentFull.ts
yanirs: It's trivial to fix the mock API by changing the examples in the Swagger file. Let's do that.

jessie-ross:
Closed in favour of #204

PR #189, opened 2020-07-03
Some Suggested Changes for PR 188

dhurlburtusa:
Some Suggested Changes for PR #188

This is to help clarify some of the comments added to PR #188.

@jessie-ross If you like what is in here, then you can simply merge into your branch. Or cherry pick what you like.

### DIFF
--- a/api/ExperimentsApi.ts
+++ b/api/ExperimentsApi.ts
@@ -1,7 +1,8 @@
 /* eslint-disable @typescript-eslint/ban-ts-ignore */
 // This is temporary for the WIP,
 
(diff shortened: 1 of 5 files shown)


jessie-ross:
Thanks for the help on this.

I am hoping to make some big changes with this one, I will be P2ing soon.

jessie-ross:
Looks like the main part in this is in `api/ExperimentsApi.ts` and I am aware of that.
I will close this for now.

jessie-ross:
As far as the schema changes, I don't think we should be changing our schema to fit our testing problems.

PR #190, opened 2020-07-03
Fix Double Border Issue

dhurlburtusa:
Fix Double Border Issue. (Basically a re-based version of #185).

Removes the double border between the last row in a table and the surrounding paper.

## How has this been tested?

Manually, visually tested.


jessie-ross:
I am going to close this for now as I can't see a difference.
I will revive it if I see a need for it.

PR #193, opened 2020-07-03
Replace ErrorsBox with MuiAlert

### LINKED ISSUE #183: Consider Replacing ErrorsBox with MUI Alert
Since we are using MUI, I think we can replace the `ErrorsBox` with a [MUI Alert component](https://material-ui.com/components/alert/) instead.

dhurlburtusa:
This shows about want is involved in using MuiAlert.

It is apart of Material-UI labs. That is, it is not apart of core. I am guessing it will become apart of core sometime this year. But this is just guess.

This removes three files. That is, we won't have to maintain these three files.

Resolves #183 if this is the route we want to take.

## How has this been tested?

Injected in a temporary error and ran the app.

Updated the snapshots. Automated tests continue to pass with 100% coverage.


components/Layout.tsx
```diff
@@ -18,6 +18,11 @@ const useStyles = makeStyles((theme: Theme) =>
       minHeight: '100vh',
     },
 
+    // Alert
+    alert: {
+      marginBottom: theme.spacing(2),
```
dhurlburtusa, line 23: If we use MuiAlert in many other places and they also need a bottom margin for better layout, then we can add the bottom margin to all MuiAlerts and remove it only on the few that don't need it.

components/Layout.tsx
```diff
@@ -110,7 +115,11 @@ const Layout = ({ title, error, children }: { title: string; error?: Error | nul
               <Typography variant='h1' className={classes.contentTitle}>
                 {title}
               </Typography>
-              {error && <ErrorsBox errors={[error]} />}
+              {error && (
+                <Alert className={classes.alert} severity='error'>
```
dhurlburtusa, line 119: An alternative solution is to keep the `ErrorsBox` component but update it to use an `Alert`.

package.json
```diff
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "@material-ui/core": "^4.10.0",
+    "@material-ui/lab": "^4.0.0-alpha.56",
```
dhurlburtusa, line 18: I didn't know it was in alpha until I implemented this.

styles/theme.ts
```diff
@@ -1,4 +1,5 @@
 import { createMuiTheme } from '@material-ui/core/styles'
+import type {} from '@material-ui/lab/themeAugmentation'
```
dhurlburtusa, line 2: This adds the types to the theme overrides for the components of MUI lab. Technically, we don't need them yet since we are not overriding the styles of any MUI lab components. But leaving it here prepares us for that.

yanirs:
I think this looks good. I tested it on Vercel by blocking virtserver.swaggerhub.com, which gives us this:

image

@jessie-ross do you see any issues with adding MUI labs as a dependency?

styles/theme.ts
jessie-ross: Let's not include it if it isn't needed yet.

jessie-ross:
MUI Alerts look fine as they are minimal and very well isolated so they are easy to replace.

I recommend using snackbars for Errors more generally:
https://material-ui.com/components/snackbars/

It turns the alerts into semi-notifications.

yanirs:
Thanks @jessie-ross! Are you happy with merging it then?

> I recommend using snackbars for Errors more generally: https://material-ui.com/components/snackbars/

The second example on that page uses MuiAlerts. :slightly_smiling_face:

jessie-ross:
It think it is good to use the imperative version snackbars for code-reusability, MUI recommends notistack and it looks like a decent library:
https://iamhosseindhv.com/notistack

I will PR shortly

yanirs:
Closing as this was replaced by #206.

PR #197, opened 2020-07-06
Suggestion for MetricDetails (#173): Remove coverage suppression

yanirs:
Just pushing some code I had locally from playing with my suggestions on #173. It doesn't give us full coverage yet, but it makes the suppression much less broad.

### DIFF
--- a/components/MetricsTable.test.tsx
+++ b/components/MetricsTable.test.tsx
@@ -1,4 +1,4 @@
-import { getByText } from '@testing-library/react'
+import { fireEvent, getByText, screen, waitFor } from '@testing-library/react'
 import React from 'react'
 
 import Fixtures from '@/helpers/fixtures'
@@ -18,7 +18,7 @@ test('with no metrics, renders an empty table', () => {
   expect(tBodyElmt).toHaveTextContent('')
 })
 
-test('with some metrics, renders a table', () => {
+test('with some metrics, renders a table', async () => {
   const { container } = render(<MetricsTable metrics={Fixtures.createMetricBares(2)} />)
 
   const tBodyElmt = container.querySelector('tbody') as HTMLTableSectionElement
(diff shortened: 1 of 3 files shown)


components/MetricsTable.test.tsx
```diff
@@ -27,4 +27,7 @@ test('with some metrics, renders a table', () => {
   expect(getByText(tBodyElmt, 'This is metric 1', { selector: 'tr > td' })).toBeInTheDocument()
   expect(getByText(tBodyElmt, 'revenue', { selector: 'tr > td' })).toBeInTheDocument()
   expect(getByText(tBodyElmt, 'conversion', { selector: 'tr > td' })).toBeInTheDocument()
+
+  fireEvent.click(screen.getByText('metric_1'))
+  await waitFor(() => screen.getByText('Oops! Something went wrong while trying to load a Metric.'))
```
yanirs, line 32: I think this is fairly low effort for getting the `MetricDetails` component to render. I'm also inclined to [follow the React Testing Library example](https://testing-library.com/docs/react-testing-library/example-intro) and add a simple server mock to get the successful case. Alternatively we can mock the client. This is probably better discussed in a P2 post on our testing strategy (perhaps use the existing one: pbmo2S-hZ-p2).

components/MetricsTable.tsx
```diff
@@ -65,9 +63,6 @@ const MetricDetail = ({ metricBare }: { metricBare: MetricBare }) => {
       .finally(() => setIsLoading(false))
   }, [metricBare.metricId])
 
-  const params = metricFull?.eventParams || metricFull?.revenueParams
-  const paramsStringified = params && JSON.stringify(params, null, 4)
```
yanirs, line 69: Inlining this means that we don't need to handle unreachable cases like `metricFull` or `params` being null. It's also clear from the coverage report (70-83 uncovered) that the successful case isn't tested by the unit tests.

jessie-ross: Ok, I do generally prefer non-inlined for readability

PR #198, opened 2020-07-06
Add canonical sorting methods and simplify AnalysisSummary

yanirs:
In preparation to addressing #96 and #98 and as an alternative to #180, this PR adds `getSortedVariations()` and `getSortedMetricAssignments()` to `ExperimentFull` and uses them to simplify the `AnalysisSummary` component and other components that should follow the canonical order for those fields.

## How has this been tested?

Automated testing, which now includes tests for the new methods.

### DIFF
--- a/components/AnalysisSummary.tsx
+++ b/components/AnalysisSummary.tsx
@@ -52,12 +52,11 @@ function ParticipantCounts({
   experiment: ExperimentFull
   latestPrimaryMetricAnalyses: Analysis[]
 }) {
-  const sortedVariations = _.orderBy(experiment.variations, ['isDefault', 'name'], ['desc', 'asc'])
   const tableColumns = [
     { title: 'Strategy', render: ({ analysisStrategy }: Analysis) => AnalysisStrategyToHuman[analysisStrategy] },
     { title: 'Total', render: ({ participantStats }: Analysis) => participantStats.total },
   ]
-  sortedVariations.forEach(({ variationId, name }) => {
+  experiment.getSortedVariations().forEach(({ variationId, name }) => {
     tableColumns.push({
       title: name,
       render: ({ participantStats }: Analysis) => participantStats[`variation_${variationId}`] || 0,
@@ -79,30 +78,23 @@ function ParticipantCounts({
  */
 function LatestResults({
   experiment,
-  metrics,
+  metricsById,
   metricAssignmentIdToLatestAnalyses,
 }: {
   experiment: ExperimentFull
-  metrics: MetricBare[]
+  metricsById: { [key: number]: MetricBare }
   metricAssignmentIdToLatestAnalyses: { [key: number]: Analysis[] }
 }) {
-  // TODO: It'd be better to move some mappings to model methods once things are more stable. We should be able to make
-  // TODO: calls like metricAssignment.getMetric().name and experiment.getMetricAssignmentById(123).getMetric().name
-  // TODO: rather than construct mappings in the components.
-  const metricsById = useMemo(() => _.zipObject(_.map(metrics, 'metricId'), metrics), [metrics])
   // Sort the assignments for consistency and collect the data we need to render the component.
   const resultSummaries = useMemo(() => {
-    return _.orderBy(experiment.metricAssignments, ['isPrimary', 'metricAssignmentId'], ['desc', 'asc']).map(
-      ({ metricAssignmentId, attributionWindowSeconds, metricId }) => {
-        return {
-          metricAssignmentId,
-          attributionWindowSeconds,
-          metricName: metricsById[metricId].name,
-          latestAnalyses: metricAssignmentIdToLatestAnalyses[metricAssignmentId as number],
-        }
-      },
-    )
-  }, [experiment.metricAssignments, metricsById, metricAssignmentIdToLatestAnalyses])
+    return experiment.getSortedMetricAssignments().map((metricAssignment) => {
+      return {
+        metricAssignment,
+        metric: metricsById[metricAssignment.metricId],
+        latestAnalyses: metricAssignmentIdToLatestAnalyses[metricAssignment.metricAssignmentId as number],
+      }
+    })
+  }, [experiment, metricsById, metricAssignmentIdToLatestAnalyses])
   const tableColumns = [
     { title: 'Strategy', render: ({ analysisStrategy }: Analysis) => AnalysisStrategyToHuman[analysisStrategy] },
     {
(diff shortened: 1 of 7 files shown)


components/AnalysisSummary.tsx
```diff
@@ -79,30 +78,23 @@ function ParticipantCounts({
...
-        }
-      },
-    )
-  }, [experiment.metricAssignments, metricsById, metricAssignmentIdToLatestAnalyses])
+    return experiment.getSortedMetricAssignments().map((metricAssignment) => {
+      return {
+        metricAssignment,
+        metric: metricsById[metricAssignment.metricId],
+        latestAnalyses: metricAssignmentIdToLatestAnalyses[metricAssignment.metricAssignmentId as number],
+      }
```
yanirs, line 95: @jessie-ross I also followed your suggestion from #180 to avoid early destructuring. I agree it's cleaner. :slightly_smiling_face:

components/AnalysisSummary.tsx
```diff
@@ -174,6 +166,7 @@ export default function AnalysisSummary({
   metrics: MetricBare[]
   debugMode?: boolean
 }) {
+  const metricsById = useMemo(() => _.zipObject(_.map(metrics, 'metricId'), metrics), [metrics])
```
yanirs, line 169: Moved here because it'll be needed by the new components (WIP PR coming soon).

We also build this mapping in a different way in `MetricAssignmentsPanel`. What do you think of an `asMapping` option to `MetricsApi.findAll()`? So far, it seems like with either need it as a flat array or as an ID-to-metric mapping, but in both forms.

components/MetricAssignmentsPanel.tsx
```diff
@@ -60,13 +59,8 @@ const useStyles = makeStyles((theme: Theme) =>
...
   const classes = useStyles()
   const resolvedMetricAssignments = useMemo(
-    () =>
-      _.orderBy(
-        resolveMetricAssignments(experiment.metricAssignments, metrics),
-        ['isPrimary', _.property('metric.name')],
-        ['desc', 'asc'],
-      ),
-    [experiment.metricAssignments, metrics],
+    () => resolveMetricAssignments(experiment.getSortedMetricAssignments(), metrics),
```
yanirs, line 62: The previous order wasn't deterministic since we can have multiple assignments to the same metric with different attribution windows.

components/VariationsTable.tsx
```diff
@@ -19,13 +18,12 @@ const useStyles = makeStyles((theme: Theme) =>
 )
 
 /**
- * Renders the variations in tabular formation.
+ * Renders the variations in tabular formation, in the order that they're given.
```
yanirs, line 21: Alternatively, it may make more sense to pass the experiment in the props instead of `variations`, as variations don't exist independently of an experiment.

models/ExperimentFull.ts
```diff
@@ -217,6 +219,20 @@ export class ExperimentFull implements ApiDataSource {
   hasConclusionData(): boolean {
     return !!this.endReason || !!this.conclusionUrl || typeof this.deployedVariationId === 'number'
   }
+
+  /**
+   * Return the experiment's variations sorted in the canonical order: Default first, then by name.
+   */
+  getSortedVariations(): Variation[] {
```
yanirs, line 226: Alternatively, we can sort in the constructors, but being more explicit feels nicer.

components/AnalysisSummary.tsx
jessie-ross: In the future it may even be good to take this one step further, especially when we are want to derive some display options along this dimension:

```
// Rough names
getExperimentResults: (experiment, metrics, analyses) => { experiment, metricAssignment, metric, analyses }[] 
latestAnalyses: analyses[] => analyses[]
getExperimentResultDisplayOptions: (experimentResult: { experiment, metricAssignment, metric, analyses }) => DisplayOptions

// You would use it like this
getExperimentResults(experiment, metrics, analyses)
    .map(x => { ...x, analyses: latestAnalyses(x.analyses) })
    // Here adding displayOptions as an extension, as I like to keep related items together.
    .map(x => { ...x, displayOptions: getExperimentResultDisplayOptions(x) })
```

I can see something like the `{ experiment, metricAssignment, metric, analyses }` collection being great to pass down, and we should be able to drop `metricAssignmentIdToLatestAnalyses`.

components/AnalysisSummary.tsx
```diff
@@ -139,14 +131,14 @@ function LatestResults({
...
+      {resultSummaries.map(({ metricAssignment, metric, latestAnalyses }) => (
+        <div key={metricAssignment.metricAssignmentId}>
           <Typography variant={'subtitle1'}>
             <strong>
-              <code>{metricName}</code>
+              <code>{metric.name}</code>
             </strong>{' '}
-            with {AttributionWindowSecondsToHuman[attributionWindowSeconds]} attribution, last analyzed on{' '}
-            <DatetimeText datetime={latestAnalyses[0].analysisDatetime} excludeTime={true} />
+            with {AttributionWindowSecondsToHuman[metricAssignment.attributionWindowSeconds]} attribution, last analyzed
```
jessie-ross, line 140: In the future let's use a library for this.

components/AnalysisSummary.tsx
jessie-ross: Yeah we need to have a talk about overall normalization throughout the code, in which indexing here is a specific example.

This is the usual library for it: https://github.com/paularmstrong/normalizr, (created by Dan Abrahmov).

It can be good to start normalized with everything and then pass it down, I think it would be sufficient if we load data at the top level e.g. here we might start with `normalizedExperiment` and `normalizedMetrics`. 

I am still thinking about if it is optimal in general but I do like the idea and it can be very nice to 
just have it as default throughout the code, usually good architecture comes out of it. (Denormalizing is cheap and easy with `nomalizr`.)

I am not a fan of the `asMapping` option, or adding options to methods in general.
I like functions/methods to be fairly single purpose, I prefer composing functions or at least adding a second method that uses the first. Further I don't see it as the API classes responsibility to manage an array of returned data.

components/MetricAssignmentsPanel.tsx
```diff
@@ -60,13 +59,8 @@ const useStyles = makeStyles((theme: Theme) =>
 function MetricAssignmentsPanel({ experiment, metrics }: { experiment: ExperimentFull; metrics: MetricBare[] }) {
   const classes = useStyles()
   const resolvedMetricAssignments = useMemo(
```
jessie-ross, line 61: I don't think we need to be using so many `useMemos`:
1. The front-end is very fast and these aren't expensive operations.
2. It is extremely easy to add a bunch of them on later.
3. These ones don't add much since their dependencies are the same as the Components they sits within.

jessie-ross:
I've left some general comments but LGTM

components/AnalysisSummary.tsx
yanirs: What do you mean? A library for the mapping from seconds to human-friendly format? I think we thought about it but since it's an enum it was simpler to add a static mapping rather than another dependency.

components/AnalysisSummary.tsx
yanirs: `normalizr` looks like a good option. We should definitely solve the general problem. Added #201.

components/MetricAssignmentsPanel.tsx
yanirs: Makes sense. Some of the `useMemo()` calls were added in a previous PR following a comment by Dan, though it felt like premature optimisation to me. As this call was already here and there a few others around, I added #202 to handle all calls in general.

components/AnalysisSummary.tsx
jessie-ross: Yeah that is fair enough

PR #199, opened 2020-07-06
Add simplified analysis table

### LINKED ISSUE #96: Add analysis warnings and reduce table content

### LINKED ISSUE #98: Add analysis debugging page

yanirs:
Resolves #96 by adding a simplified analysis table and moving the existing tables to be shown only when in debug mode.

Inspired by @jessie-ross's implementation of metric details in #173, I implemented the summary as a very simple table with the high-level recommendation for each metric assignment, with extra details available in an expandable `detailPanel`. Unlike the current view, we only show a single analysis strategy per metric assignment, and say that manual analysis is needed if the recommendations drawn from the different analysis strategies are inconsistent.

As our SwaggerHub mock isn't very interesting, check out Storybook to see what it looks like with richer results:

image

The styling isn't final, but it's probably better to handle it in future PRs.

Closes #98, as this gives us a basic debug view and we can add more debugging info in the future.

## How has this been tested?

* Manual testing in storybook and on the dev server
* Added and adjusted automatic tests


aaronyan:
>  implemented the summary as a very simple table with the high-level recommendation for each metric assignment, with extra details available in an expandable detailPanel

This looks great, love the addition of the table + `detailPane`. My first impression is wanting to add the participant total counts and difference interval as columns to the table (instead of hidden in the `detailPane`). I think those details would add to the recommendation, but the difference interval may be a bit early to add to the table.

The difference interval is a naturally difficult to parse. If we don't add it now, we should consider adding it to the table in a future PR when we can have some nice way to visualize it eg sparkline-like graphs or something.

> Unlike the current view, we only show a single analysis strategy per metric assignment, and say that manual analysis is needed if the recommendations drawn from the different analysis strategies are inconsistent.

:100: Keep it simple when things align. And send out an SOS when things get complicated. :)

yanirs:
> > implemented the summary as a very simple table with the high-level recommendation for each metric assignment, with extra details available in an expandable detailPanel
> 
> This looks great, love the addition of the table + `detailPane`. My first impression is wanting to add the participant total counts and difference interval as columns to the table (instead of hidden in the `detailPane`). I think those details would add to the recommendation, but the difference interval may be a bit early to add to the table.
> 
> The difference interval is a naturally difficult to parse. If we don't add it now, we should consider adding it to the table in a future PR when we can have some nice way to visualize it eg sparkline-like graphs or something.

Thanks @aaronyan! Yeah, I think that a visualisation for the difference interval would help and make it easier to digest in the context of the table. Maybe something like what Google Optimize shows?

image

Definitely not for this PR, though, so can leave it in the detail pane for now.

Regarding participant counts, I think they may be nice to include as a column, but it can also be confusing because the counts depend on the last analysis date, which varies depending on the attribution window. For example, the post-spammer counts for a metric that was analysed on 2020-07-01 would be different from the post-spammer counts for one that was analysed on 2020-07-10, even if the experiment ended in June. We can actually address this by always analysing all the metrics until the maximum of the last dates. This would make the counts identical across all metrics, so it'd make sense to present them outside the table. What do you think?

models/ExperimentFull.ts
```diff
@@ -217,6 +219,20 @@ export class ExperimentFull implements ApiDataSource {
   hasConclusionData(): boolean {
     return !!this.endReason || !!this.conclusionUrl || typeof this.deployedVariationId === 'number'
   }
+
+  /**
+   * Return the experiment's variations sorted in the canonical order: Default first, then by name.
+   */
+  getSortedVariations(): Variation[] {
```
jessie-ross, line 226: This is not a change needed here but rather me shilling my #204 PR:

Under the "lib" model we might have `Variations.sort: Variation[] => Variation[]`, and it becomes more general purpose. We would use it like `Variations.sort(experiment.variations)`.

For instance I can see me needing this function as part of the ExperimentCreation form on a list of variations. 

Same with MetricAssignments.

aaronyan:
> Maybe something like what Google Optimize shows?
Yeah, that would be amazing.

> We can actually address this by always analysing all the metrics until the maximum of the last dates.

I'm a fan! Analyzing metrics until the maximum of the last dates sounds like a good call to keep counts across metrics consistent and easy to understand.

> This would make the counts identical across all metrics, so it'd make sense to present them outside the table. What do you think?

When you say "outside the table" do you mean in a separate card (like an audience card)? I think this is what you mean, because if all metrics have the same participant counts, putting them in a separate card would avoid repeating the same participant count with each metric. 

I'd imagine the participant count being in something like an audience card for another future iteration. Hehe, so fun to dream about future iterations. :)
- percentage of assigned variations over time
- total analyzed participants
- percentage of participants that were crossovers and excluded from analysis
- percentage of participants that were spammers and excluded from analysis

I think we're in agreement with what to do for now. :) To summarize:
- Analyze all the metrics until the maximum of the last dates
- Keep the results metrics view as you have it (no participant total counts and no difference interval as columns to the table)

Sound good? If so, I can help make the issue for the analyzing all the metrics until the maximum of the last dates.

components/experiment-results/ExperimentResults.tsx
```diff
@@ -37,35 +38,42 @@ export default function ExperimentResults({
...
-      <div className='analysis-participant-counts'>
-        <h3>Participant counts for the primary metric</h3>
-        <ParticipantCounts
-          experiment={experiment}
-          latestPrimaryMetricAnalyses={
-            metricAssignmentIdToLatestAnalyses[experiment.getPrimaryMetricAssignmentId() as number]
-          }
-        />
-      </div>
+  if (debugMode) {
```
yanirs, line 41: The diff here is [mostly whitespace](https://github.com/Automattic/abacus/pull/199/files?diff=split&w=1).

components/experiment-results/__snapshots__/ExperimentResults.test.tsx.snap
```diff
@@ -1,6 +1,366 @@
...
 
-exports[`shows the analyses JSON in debug mode 1`] = `
+exports[`renders the condensed table with some analyses in non-debug mode 1`] = `
+<div
+  class="analysis-latest-results"
+>
+  <h3>
+    Latest results by metric
+  </h3>
+  <div
```
yanirs, line 10: Material Table generates so much inline styling... @jessie-ross do you know of a simple way to compare snapshots of just the table contents?

components/experiment-results/ExperimentResults.test.tsx
```diff
@@ -13,11 +14,14 @@ test('renders an appropriate message with no analyses', () => {
...
+    <ExperimentResults analyses={analyses} experiment={experiment} metrics={metrics} debugMode={true} />,
+  )
 
-  // In non-debug mode, we shouldn't have a <pre> element with the JSON.
-  expect(container.querySelector('pre')).toBeNull()
+  // In debug mode, we should have a <pre> element with the JSON.
+  expect(container).toHaveTextContent(`Found ${analyses.length} analysis objects in total.`)
+  expect(container.querySelector('pre.debug-json')).toMatchSnapshot()
 
   // Table snapshots are somewhat verbose, but they will allow us to see if there are any expected changes to the
```
yanirs, line 26: I'll open a separate PR to move the inline snapshots to the snapshot file. For this PR, I wanted to make it clear that what was previously in non-debug mode is now the debug view.

components/experiment-results/CondensedLatestAnalyses.tsx
```diff
@@ -0,0 +1,138 @@
...
+        }
+      }}
+      detailPanel={detailPanel}
+    />
+  )
+}
+
+function AnalysisDetailPanel({ analysis, experiment }: { analysis: Analysis; experiment: ExperimentFull }) {
+  return (
+    <dl className='analysis-detail-panel'>
```
yanirs, line 101: As noted in the PR description, the styling can be improved and MUIfied, but I think it's better done as a future iteration on the result views.

components/experiment-results/ExperimentResults.test.tsx
yanirs: Opened #214.

yanirs:
> When you say "outside the table" do you mean in a separate card (like an audience card)? I think this is what you mean, because if all metrics have the same participant counts, putting them in a separate card would avoid repeating the same participant count with each metric.
> 
> I'd imagine the participant count being in something like an audience card for another future iteration. Hehe, so fun to dream about future iterations. :)
> 
>     * percentage of assigned variations over time
>     * total analyzed participants
>     * percentage of participants that were crossovers and excluded from analysis
>     * percentage of participants that were spammers and excluded from analysis
> 
> I think we're in agreement with what to do for now. :) To summarize:
> 
>     * Analyze all the metrics until the maximum of the last dates
>     * Keep the results metrics view as you have it (no participant total counts and no difference interval as columns to the table)

Thanks @aaronyan. This sounds good. I added #215 to cover the participant count move and a backend issue for the metric analysis dates. The PR is now ready for review (CC @jessie-ross).

components/experiment-results/__snapshots__/ExperimentResults.test.tsx.snap
jessie-ross: Ideally we wouldn't be testing MaterialTable but trusting it and mocking the component to test its inputs, i.e. that  `columns` and `data` props are receiving what we expect. 

Unfortunately MaterialTable doesn't do any testing of their own so as far as our testing standards go we may want to test it.
(Linking this to #205 for the future.)

components/experiment-results/CondensedLatestAnalyses.tsx
```diff
@@ -0,0 +1,138 @@
...
+        }
+        return <RecommendationString recommendation={analysis.recommendation} experiment={experiment} />
+      },
+    },
+  ]
+  const detailPanel = [
+    ({ analysis, recommendationConflict }: { analysis?: Analysis; recommendationConflict?: boolean }) => {
+      return {
+        render: () => analysis && <AnalysisDetailPanel analysis={analysis} experiment={experiment} />,
+        disabled: !analysis || recommendationConflict,
```
jessie-ross, line 76: It is a bit confusing when they are disabled, especially if it is due to a "conflict". 
In the future it might be good to open up the details panel and then explain why Analysis details aren't showing up, and what they can do if a manual analysis is required i.e. who to ping.

components/experiment-results/CondensedLatestAnalyses.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const resultSummaries = experiment.getSortedMetricAssignments().map((metricAssignment) => {
+    const latestAnalyses = metricAssignmentIdToLatestAnalyses[metricAssignment.metricAssignmentId as number] || []
+    const uniqueRecommendations = _.uniq(latestAnalyses.map(({ recommendation }) => JSON.stringify(recommendation)))
+    return {
+      metricAssignment,
+      metric: metricsById[metricAssignment.metricId],
+      analysis: latestAnalyses.find((analysis) => analysis.analysisStrategy === defaultAnalysisStrategy),
+      recommendationConflict: uniqueRecommendations.length > 1,
+    }
+  })
```
jessie-ross, line 44: FYI: These will be changed soon as part of #204

jessie-ross:
Minor comments, looks fine for now.

components/experiment-results/CondensedLatestAnalyses.tsx
yanirs: Sure :slightly_smiling_face:

components/experiment-results/CondensedLatestAnalyses.tsx
yanirs: Yeah, I didn't have it disabled in the first iteration, but found having to click through to a mostly-empty panel a bit confusing. Can definitely iterate on it. :slightly_smiling_face:

PR #203, opened 2020-07-07
Split AnalysisSummary file by component

yanirs:
Split the `AnalysisSummary` file as part of the supporting work for #199:

* Move all the components to separate files under `/components/experiment-results`.
* Rename `AnalysisSummary` to `ExperimentResults`. This is in line with the page path where this component is used (`/experiments/<id>/results`). It also reduces ambiguity as an experiment's results include multiple analyses.
* Rename `LatestResults` to `FullLatestAnalyses` in preparation for #199, which will render this component only in debug mode and show a condensed version of the latest analyses outside debug mode.

**Note**: While the diff may look big, it's just moving around existing code, so let's keep the review focused on component names and the directory structure. :slightly_smiling_face: 

## How has this been tested?

* Manually checked that the result page and stories render as before.
* Automated snapshot tests for `AnalysisSummary` were moved and renamed, and are still passing without any adjustments to the snapshots


yanirs:
> I like this change, definitely approve of organising the components folder further, ([not sure if I showed you this](https://react-file-structure.surge.sh/)), and the Analysis/Experiment naming was bugging me too :)

Thanks @jessie-ross! _"Move files around until it feels right"_ definitely captures the spirit of this PR, and the `/components` folder can definitely use further moves now that there are many components. :slightly_smiling_face:

PR #204, opened 2020-07-07
Schema Refactor

### LINKED ISSUE #149: Metric detail page should show a full metric object
See: https://github.com/Automattic/abacus/pull/136#discussion_r443393110. It looks like `MetricsApi.findById()` returns a `MetricBare` object rather than a `MetricFull` object. From a quick search, this is likely to be due to the way TypeScript handles static method inheritance.

jessie-ross:
### This PR proposes to :
- Add all entities to `schemas.ts`
  - Schema Docs: Include only a link to the swagger documentation, so there is no need to keep up to date.
  - Replaces all typescript interfaces using yup.InferSchema
- Seperate code from data by moving from model centric to lib centric:
   - Removes the need for boxing/unboxing with classes
   - Removes the need for testing boxing/unboxing
   - Removes the coupling in our fetchApi lib (We should be using a well made and tested lib for this)
   - Fixes problems we have of taxonomy:
      We can group related functions but not have them necessarily work on the same data structure
      E.g. `latestAnalyses: Analysis[] -> Analysis[]` right now we don't have a place for it but we would put it in `/lib/analyses.ts`.
      E.g. working with normalized forms instead of denormalized 
      `Experiments.getPrimaryMetric: normalizedExperiment -> Metric`
   - Tends to be idiomatic with modern javascript and works well with various libs
- Provides an alternate solution to bundling functions together:
   `import * as Experiments from '@/lib/experiments'`
   `Experiments.getPrimaryMetric(experiment)`
    
   `import * as ResultProcessing from '@/lib/result-processing`
    `getResultDataByMetric(normalizedExperiment, normalizedMetrics, normalizedAnalyses)`

### For a future PR:

- Fixture data testing
   - Fixture data should be tested against the schema.
   - Fixture data should be server data which we coerce with our schema

--- 
### Previous WIP Pull Request:

_This is the second of two controversial topics I want to broach (the first being testing)_

_As usual I am just presenting an argument and I am happy to let it fall where it may, but I would feel amiss if I didn't bring it up at this formative stage in the project._

**This PR demos a refactor that seems large but is mostly deletion:**
- Add all entities to `schemas.ts`
  - Schema Docs: Include only a link to the swagger documentation, so there is no need to keep up to date.
  - (More optional) Replaces all typescript interfaces using yup.InferSchema
     This is working very well, but I have never used it before so I don't know.
     A downside is not having local docs.
- Seperate code from data by moving from model centric to lib centric:
   - Removes the need for boxing/unboxing with classes
   - Removes the need for testing boxing/unboxing
   - Removes the coupling in our fetchApi lib (We should be using a well made and tested lib for this)
   - Fixes problems we have of taxonomy:
      We can group related functions but not have them necessarily work on the same data structure
      E.g. `latestAnalyses: Analysis[] -> Analysis[]` right now we don't have a place for it but we would put it in `/lib/analyses.ts`.
      E.g. working with normalized forms instead of denormalized 
      `Experiments.getPrimaryMetric: normalizedExperiment -> Metric`
   - Tends to be idiomatic with modern javascript and works well with various libs
- Provides an alternate solution to bundling functions together:
   `import * as Experiments from '@/lib/experiments'`
   `Experiments.getPrimaryMetric(experiment)`
    
   `import * as ResultProcessing from '@/lib/result-processing`
    `getResultDataByMetric(normalizedExperiment, normalizedMetrics, normalizedAnalyses)`
- (Not shown) Fixture data testing
   - Fixture data should be tested against the schema.

**The end result will be:**
- No more `/models` 1000+ lines 
- A '/lib' directory, this is what I normally use but happy to rename, these aren't utils but core business logic.
  - `/lib/schemas.ts`
     200-300 lines at an estimate
     Significantly better testing by both integration tests and verification of fixture data
     More functionality than before - before most model files did very little other than boxing/unboxing
  - `/lib/experiments.ts` ~40 lines
  - `/lib/experiments.test.ts` ~50 lines
  - Thats it, once you remove the boxing/unboxing methods there are no other methods I can see at the moment

**General Discussion**

At the core of this change is also the idea that we should be using functions by default, rather than classes.
This is not out of disregard for OOP, I think OOP is useful in the right situations, but rather because classes are a step up in complexity then functions: They add an extra layer of modelling. 

If we stick to data and transforms of that data we can avoid the risks and decisions we will have to make with regards to modelling.

We already have a well modelled system with the data structures, classes would be another layer on top of that. In this application I don't see us needing anything more than fairly simple data-transformations.

If we just want to group functions together this PR demonstrates one out of a few methods of doing so.

My motivation behind this PR is a want to keep things simple.

(Additionally I did discuss this with another JavaScript Engineer and have heard that on JavaScript A8C the current push is towards functional - but it seems like we don't necessarily want to be matching outside culture.)

Fixes #149.


jessie-ross:
@yanirs this is now ready for review.

The PR ballooned a bit after the last merge but most changes are just changing the import references.

The only tests not passing are the integration tests:

```
  ● ExperimentsApi.ts module › findById › should return an experiment with the expected experiment shape

    expect(received).toEqual(expected) // deep equality

    - Expected  -  1
    + Received  + 10

    - Array []
    + Array [
    +   "p2Url must be a valid URL",
    +   "conclusionUrl must be a valid URL",
    +   "deployedVariationId must be a positive number",
    +   "segmentAssignments[0].segmentAssignmentId must be a positive number",
    +   "segmentAssignments[0].experimentId must be a positive number",
    +   "segmentAssignments[0].segmentId must be a positive number",
    +   "variations[0].variationId must be a positive number",
    +   "variations[0].allocatedPercentage must be greater than or equal to 1",
    + ]

  ● SegmentsApi.ts module › findAll › should return a set of segments with the expected segment shape

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 3

    - Array []
    + Array [
    +   "[0].segmentId must be a positive number",
    + ]
```

I will PR to Experimentation Platform to fix them.

components/AudiencePanel.tsx
```diff
@@ -61,7 +62,7 @@ function AudiencePanel({ experiment, segments }: { experiment: ExperimentFull; s
     {
       label: 'Variations',
       padding: 'none' as TableCellProps['padding'],
-      value: <VariationsTable variations={experiment.getSortedVariations()} />,
+      value: <VariationsTable variations={Variations.getSortedVariations(experiment.variations)} />,
```
yanirs, line 65: `Variations.getSortedVariations(experiment.variations)` is a bit awkward and repetitive. What do you think of renaming to consider the new namespace? `Variations.sort(experiment.variations)` may be a good option.

The same applied to the other `getSortedField()` methods.

api/utils.ts
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable */
+// This is temporary for the WIP,
+
```
yanirs, line 3: Can this be removed now? May need to change the type of `body` to `unknown` to make eslint happy.

api/ExperimentsApi.ts
```diff
@@ -20,18 +29,17 @@ async function create(experiment: ExperimentFull) {
  * @throws UnauthorizedError
  */
 async function findAll(): Promise<ExperimentBare[]> {
-  return (await fetchApi('GET', '/experiments')).experiments.map((apiData: ApiData) =>
-    ExperimentBare.fromApiData(apiData),
-  )
+  const { experiments } = await fetchApi('GET', '/experiments')
```
yanirs, line 32: Can we make things a bit more consistent? Now, some API functions have a temporary variable, some don't, some use destructuring and some refer to specific fields (`.metrics` in `MetricsApi).

lib/schemas.ts
```diff
@@ -0,0 +1,252 @@
...
+export type Variation = yup.InferType<typeof variationSchema>
+
+/**
+ * The platform where the experiment is running.
+ * - `calypso`: The experiment is being run on the front-end Calypso interface,
+ *   WordPress.com. Account sign-up and site management design experiments are
+ *   likely run here.
+ * - `wpcom`: The experiment is being run on the back-end, like APIs which are
+ *   usually written in PHP. Email and landing pages experiments are likely run here.
+ */
```
yanirs, line 144: Can drop this and the docstring for `Status` since the other enums are undocumented.

lib/schemas.ts
```diff
@@ -0,0 +1,252 @@
...
+    refundDays: yup.number().integer().positive().defined(),
+    productSlugs: yup.array(yup.string().defined()).defined(),
+    transactionTypes: yup.array(yup.string().oneOf(Object.values(TransactionTypes)).defined()),
+  })
+  .defined()
+  .camelCase()
+
+export type MetricRevenueParams = yup.InferType<typeof metricRevenueParamsSchema>
+
+export enum MetricParameterTypes {
```
yanirs, line 43: This should be singular like other enums.

lib/experiments.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+ * Return this experiment's default analysis strategy, which depends on the existence of exposureEvents.
+ */
+export function getDefaultAnalysisStrategy(experiment: ExperimentFull) {
+  return experiment.exposureEvents ? AnalysisStrategy.PpNaive : AnalysisStrategy.MittNoSpammersNoCrossovers
+}
+
+export function createNewExperiment(): Partial<ExperimentFull> {
```
yanirs, line 48: Is this still needed? The implementation looks a bit odd now.

lib/schemas.ts
```diff
@@ -0,0 +1,252 @@
...
+  ThreeDays = 259200,
+  OneWeek = 604800,
+  TwoWeeks = 1209600,
+  ThreeWeeks = 1814400,
+  FourWeeks = 2419200,
+}
+
+export const metricAssignmentSchema = yup
+  .object({
+    metricAssignmentId: idSchema.defined(),
```
yanirs, line 83: General question, probably not for this PR: Previously we made ID fields like this one nullable because assignment fields don't have IDs on the client side before they're created. How will we handle this now?

api/AnalysesApi.ts
```diff
@@ -11,9 +12,8 @@ import { fetchApi } from './utils'
  * @throws UnauthorizedError
  */
 async function findByExperimentId(experimentId: number): Promise<Analysis[]> {
-  return (await fetchApi('GET', '/analyses/' + experimentId)).analyses.map((apiData: ApiData) =>
-    Analysis.fromApiData(apiData),
-  )
+  const { analyses } = await fetchApi('GET', `/analyses/${experimentId}`)
+  return await yup.array(analysisSchema).defined().validate(analyses, { abortEarly: false })
```
yanirs, line 16: Is there a way to change the global yup defaults so we don't have to repeat `{ abortEarly: false }`?

__tests__/api/AnalysesApi.test.ts
```diff
@@ -1,38 +1,18 @@
...
-          if (analysis.recommendation.chosenVariationId !== null) {
-            expect(typeof analysis.recommendation.chosenVariationId).toBe('number')
-            expect(Object.values(RecommendationReason).includes(analysis.recommendation.reason)).toBe(true)
-            analysis.recommendation.warnings.forEach((warning) => {
-              expect(Object.values(RecommendationWarning).includes(warning)).toBe(true)
-            })
-          }
+      try {
+        await AnalysesApi.findByExperimentId(123)
+      } catch (e) {
```
yanirs, line 10: Is the API now expected to throw errors as part of normal operation?

For all these tests, it'd be good to keep at least one assertion, as they can now pass even if the tested functions do nothing.

Also, `err` or something similar would be better than `e`. :slightly_smiling_face:

lib/schemas.ts
jessie-ross: The current method is to extend the schema, overwriting the ID field and creating a seperate entity creation schema:
https://github.com/Automattic/abacus/blob/c7ac3858c052e45eb5c9f3264d73aa8d327583ca/lib/schemas.ts#L192-L195

api/AnalysesApi.ts
jessie-ross: The `abortEarly` is useful for the integration tests but it doesn't hurt to keep them around for run-time on the API calls. We don't want it for the form validation.

I did look into this but yup doesn't seem to have global defaults that can change, and we would need them to change just for some calls.

I'm pining a little for dynamic binding which is a nice solution to the general problem, alas it is not idiomatic in JavaScript.

__tests__/api/AnalysesApi.test.ts
jessie-ross: > Is the API now expected to throw errors as part of normal operation?

It isn't necessary we use it but, yep, they now will throw ValidationErrors if there is anything not right.

It gives us guarantees about the data these functions output, which is very useful for run-time correctness.

jessie-ross: To expand on run-time correctness: Input validation and type-systems go very well together - we now know our types will be correct at run-time as we are sure of the data coming through the system.

A good example of this is the test above. `AnalysesApi.findByExperimentId`'s return type is matched against the schema validation type, so we have a very high guarantee that it will either return the promised `Analysis[]` type or throw an error - which will fail the test.

yanirs: > It isn't necessary we use it but, yep, they now will throw ValidationErrors if there is anything not right.

So why are we catching errors in the tests?

> A good example of this is the test above. `AnalysesApi.findByExperimentId`'s return type is matched against the schema validation type, so we have a very high guarantee that it will either return the promised Analysis[] type or throw an error - which will fail the test.

Technically, the test itself doesn't check anything other than that the function is callable, which is what's bothering me. :slightly_smiling_face:

lib/experiments.ts
jessie-ross: It is being used for Experiment Creation, it creates a "default" experiment. E.g. with the audience form-part it makes the default platform WPCom.

I say at least let's move this decision to the #9 PRs.

jessie-ross:
See https://github.com/Automattic/experimentation-platform/pull/384 for the swagger changes

withinboredom:
Can we update the initial description since it's no longer a demo?

__tests__/api/AnalysesApi.test.ts
yanirs: @jessie-ross I still don't get the point of catching exceptions that just get rethrown and fail the tests anyway...

__tests__/api/AnalysesApi.test.ts
```diff
@@ -1,38 +1,19 @@
...
-          if (analysis.recommendation.chosenVariationId !== null) {
-            expect(typeof analysis.recommendation.chosenVariationId).toBe('number')
-            expect(Object.values(RecommendationReason).includes(analysis.recommendation.reason)).toBe(true)
-            analysis.recommendation.warnings.forEach((warning) => {
-              expect(Object.values(RecommendationWarning).includes(warning)).toBe(true)
-            })
-          }
+      try {
+        const analyses = await AnalysesApi.findByExperimentId(123)
+        expect(analyses.length).toBeGreaterThan(0)
```
yanirs, line 10: Can we maybe use yup to run the validations here as well?

__tests__/api/AnalysesApi.test.ts
jessie-ross: The reason for catching the exception is for better Test DevEx.

If they weren't cause we would just see an exception which might say: "8 Validation Errors".
I wanted to expose the errors inside the exception and this gave a nice output.

__tests__/api/AnalysesApi.test.ts
jessie-ross: We can but I think that would be superfluous as we are making the same validation in `AnalysesApi.findByExperimentId`.

__tests__/api/AnalysesApi.test.ts
jessie-ross: > Technically, the test itself doesn't check anything other than that the function is callable, which is what's bothering me. 🙂

We have effectively moved the same test code to run-time which is why it is missing here.

I can add an extra schema in here but I feel it is overkill:
Between E2E and integration we should be getting double coverage of "integration" tests.
We also gain IMO the most important integration test: testing against real data when it is manually tested.

yanirs: > The reason for catching the exception is for better testing DevEx.
> 
> If we didn't catch we would only see: `8 Validation Errors`.
> I wanted to expose the errors inside the exception and this gave a nice output.

Alright, I'm not a fan of adding seven duplicated lines per test for the sake of nice output, but it's a fairly small issue in the grand scheme of things and it'd be good to get this big refactor merged. It'd be nice to add a helper function for this repeated try-catch, but I'll leave it up to you if you want to do it before merging. :slightly_smiling_face:

jessie-ross: Yeah this was a good idea

Issue #200, opened 2020-07-07
Empty error boxes appearing on /experiment/[id] and /metric/[id]


jessie-ross:
Following up on Yanir's comment at pbmo2S-oh-p2#comment-912.

Replicating:
1. Visit https:/.../experiments/1 or https://.../metrics/1 directly

There are two problems going on here:
1. It looks like pages are being rendered with null path properties initally:
    <img width="1416" alt="Screen Shot 2020-07-07 at 11 46 13 am" src="https://user-images.githubusercontent.com/971886/86717900-2b558700-c055-11ea-8078-aa12f45ed05e.png">
    This might be a bug with the router loading the page before fully initialising.    
2. Error messages aren't being reset.
    I think the solution to this is to [use Snackbars](https://material-ui.com/components/snackbars/)

yanirs:
This was fixed by @jessie-ross in #206 and #209.

PR #206, opened 2020-07-08
Use notistack for Errors

jessie-ross:
This PR implements notistack for non-fatal user error messages:
- Saves on code relating to snackbars/alerts
- Provides message variants like 'error'
- Provides a better interface for code-reuse
- Decouples errors from the "layout"

Notistack maintainability case:
- Minimal, single purpose
- Small surface area `<SnackbarProvider>`, `useSnackbar`, `enqueueSnackbar`, `closeSnackbar`
- Two years old, (young but not too young)
- Active maintainer
- Recommended by MUI (a good sign).

Resolves: #183
Replaces: #193 
Fixe half of the problem of: #200 

## General thoughts

Something to think of is if we want to elevate more errors to fatal level.
In fact it looks like we don't have fatal errors: disabling any interaction and only showing a message telling a user to refresh with a refresh button as well. 

This is a good option if we want to ensure valid states on the front-end but comes at the cost of UX.

## How has this been tested?
- AT - Getting the hang of in-depth testing
- MT


jessie-ross:
Ah I need to remove the existing system, will push that shortly.

jessie-ross:
And oops I was wrong about not having fatal errors

utils/data-loading.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function useDataLoadingError<E extends Error | null>(error: E, dataName?: string) {
+  const { enqueueSnackbar } = useSnackbar()
+
+  useEffect(() => {
+    if (error) {
+      const userErrorMessage = dataName
+        ? `Oops! There was a problem loading some data of type: ${dataName}.`
+        : 'Oops! There was a problem loading some data.'
+      enqueueSnackbar(userErrorMessage, { variant: 'error' })
```
yanirs, line 12: What do you think about using `persist: true`? I played with it a bit, and it feels like the errors disappear too quickly.

utils/data-loading.ts
```diff
@@ -0,0 +1,15 @@
+import { useSnackbar } from 'notistack'
+import { useEffect } from 'react'
+
+export function useDataLoadingError<E extends Error | null>(error: E, dataName?: string) {
```
yanirs, line 4: Can you please add a short docstring?

Also, it looks like `dataName` isn't being used. If it's not useful, let's remove it for now. If it is useful, let's use it. :slightly_smiling_face:

jessie-ross: I plan to use it very soon with #208

utils/data-loading.ts
jessie-ross: Yeah I am ok with that.

jessie-ross:
Ok made it persist, should be good to go.

PR #207, opened 2020-07-08
Clean out Error systems

jessie-ross:
This commit branches off from #206 to clean up the existing error infrastructure.
- Removes layout error system
- Removes duplicate error boundary in layout
- Remove unnecessary "pass up" props of ReactErrorBoundary as it provides and uses "pass down"

## How has this been tested?
- MT & AT


pages/metrics/[id].tsx
```diff
@@ -31,9 +31,7 @@ const MetricsDetailPage = () => {
   useDataLoadingError(error)
 
   return (
-    <Layout title='Metrics' error={error}>
-      {isLoading ? <LinearProgress /> : <pre> {JSON.stringify(metric, null, 2)} </pre>}
-    </Layout>
+    <Layout title='Metrics'>{isLoading ? <LinearProgress /> : <pre> {JSON.stringify(metric, null, 2)} </pre>}</Layout>
```
Comment (yanirs) [inline pages/metrics/[id].tsx:34]: Unrelated to this PR but since you're already changing this line, can you please change the title to `Metric: <metric_name>`, similarly to what we have for experiment pages?

Comment (jessie-ross) [reply pages/metrics/[id].tsx:34]: Can we delete this file for now? This page is currently unused.

Comment (yanirs) [reply pages/metrics/[id].tsx:34]: Sure, makes sense.

PR #208, opened 2020-07-08
Add use data source

jessie-ross:
This PR implements an abstraction for data-loading, `useDataSource`:
- Declarative way of loading data.
- Single purpose
- Inherently async, a better pattern than needing Promise.all
- Intended to be combined with #206 to provide more detailed errors
- Cuts a lot of data loading code
- Tests the process of data loading

## How has this been tested?
- AT (using advanced testing techniques 🤓) and MT


pages/experiments/[id]/results.tsx
```diff
@@ -19,28 +19,22 @@ export default function ResultsPage() {
...
+    () => ExperimentsApi.findById(experimentId),
+    [experimentId],
+  )
+  const { isLoading: metricsIsLoading, data: metrics, error: metricsError } = useDataSource(
+    () => MetricsApi.findAll(),
+    [],
+  )
+  const { isLoading: analysesIsLoading, data: analyses, error: analysesError } = useDataSource(
+    () => AnalysesApi.findByExperimentId(experimentId),
+    [],
```
Comment (yanirs) [inline pages/experiments/[id]/results.tsx:32]: Shouldn't this be `[experimentId]`?

Also, can we easily extend the linting rules or do something to catch such cases? We get an error if the deps are missing for `useMemo()`, for example. Don't worry about it if it's too hard.

pages/experiments/[id].tsx
```diff
@@ -29,24 +29,22 @@ export default function ExperimentPage() {
...
-  const [error, setError] = useState<Error | null>(null)
-  const [experiment, setExperiment] = useState<ExperimentFull | null>(null)
-  const [metrics, setMetrics] = useState<MetricBare[] | null>(null)
-  const [segments, setSegments] = useState<Segment[] | null>(null)
+  const { isLoading: experimentIsLoading, data: experiment, error: experimentError } = useDataSource(
+    () => ExperimentsApi.findById(experimentId),
+    [experimentId],
+  )
+  const { isLoading: metricsIsLoading, data: metrics, error: metricsError } = useDataSource(
+    () => MetricsApi.findAll(),
```
Comment (yanirs) [inline pages/experiments/[id].tsx:37]: Can't we change this to `MetricsApi.findAll` (also applies elsewhere)?

pages/experiments/[id].tsx
```diff
@@ -29,24 +29,22 @@ export default function ExperimentPage() {
...
+    [],
```
Comment (yanirs) [inline pages/experiments/[id].tsx:38]: Maybe make `deps` default to an empty array to shorten these calls?

pages/experiments/[id]/results.tsx
```diff
@@ -19,28 +19,22 @@ export default function ResultsPage() {
...
-      .then(([analyses, experiment, metrics]) => {
-        setAnalyses(analyses)
-        setExperiment(experiment)
-        setMetrics(metrics)
-        return
-      })
-      .catch(setError)
-      .finally(() => setIsLoading(false))
-  }, [experimentId])
+  const error = [experimentError, metricsError, analysesError].filter((x) => !!x)[0]
```
Comment (yanirs) [inline pages/experiments/[id]/results.tsx:37]: I guess this will no longer be necessary with the changes that are now on `master`. Should update this PR accordingly.

utils/data-loading.ts
```diff
@@ -0,0 +1,27 @@
+import { DependencyList, useEffect, useState } from 'react'
+
+export function useDataSource<Data, Deps extends DependencyList | undefined, E extends Error>(
```
yanirs, line 3: Should document the exported functions.

utils/data-loading.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  const p = new Promise((resolve, reject) => {
+    resOuter = resolve
+    rejOuter = reject
+  })
+
+  return {
+    res: (resOuter as unknown) as (...x: unknown[]) => void,
+    rej: (rejOuter as unknown) as (...x: unknown[]) => void,
+    p,
+  }
```
yanirs, line 18: It'd be good to use more informative names. Any reason not to go with `resolve`, `reject`, and `promise`? It's hard to know what `p` stands for, and `res` is often used as a shorthand for `result`.

utils/data-loading.test.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+describe('utils/data-loading.ts module', () => {
```
yanirs, line 21: Nice advanced testing work! :slightly_smiling_face:

pages/experiments/[id]/results.tsx
Comment (jessie-ross) [reply pages/experiments/[id]/results.tsx:32]: > Shouldn't this be [experimentId]?

Yep that right, nice catch

> Also, can we easily extend the linting rules or do something to catch such cases? We get an error if the deps are missing for useMemo(), for example. Don't worry about it if it's too hard.

Yeah this is a good point, I had to do a bit of digging but there is an undocumented, but semi-supported option for it:
https://github.com/facebook/react/pull/16912

pages/experiments/[id].tsx
Comment (jessie-ross) [reply pages/experiments/[id].tsx:38]: The standard semantics for a missing `deps` array is to fire on every render:
https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

Although it doesn't make sense to not include the `[]` for data loading I think it is good to keep with the convention.

utils/data-loading.test.ts
jessie-ross: Yeah I can change the names around

pages/experiments/[id].tsx
Comment (jessie-ross) [reply pages/experiments/[id].tsx:37]: Ah this conflicts with adding dependency checking, I think I would rather have the dep checking as it is so nice.

```14:47  error  React Hook useDataSource received a function whose dependencies are unknown. Pass an inline function instead  react-hooks/exhaustive-deps```

pages/experiments/[id].tsx
Comment (yanirs) [reply pages/experiments/[id].tsx:38]: Ah, makes sense.

pages/experiments/[id].tsx
Comment (yanirs) [reply pages/experiments/[id].tsx:37]: Yeah, agreed.

jessie-ross:
@yanirs I have made the updates as per your resolved comments, see the unresolved comments for more info.

Additionally I've generalised `combineIsLoading` as just an `or` function - in the future we should consider a more substantial base lib like [`rambda`](ramdajs.com) instead of `lodash`.

PR #209, opened 2020-07-08
Fix null requests due to router

### LINKED ISSUE #200: Empty error boxes appearing on /experiment/[id] and /metric/[id]

jessie-ross:
Branches off of use-data-source.

Our current routing library provides a null for path parameters
initially. This was causing errors when requests were being made with
null ids.

This PR fixes the immediate problem:
- Prevents those requests being made by providing a data loading
   promise that doesn't get fulfilled in those occasions.

For the router problem a longer look is needed, might be due to pre-rendering:
> It will be an empty object during pre-rendering if the page doesn't have data fetching requirements. 
>https://nextjs.org/docs/api-reference/next/router

On that topic my general preference is to drop NextJS and not do any pre-rendering. 

Fixes #200 

## How has this been tested?
- MT 
- Existing AT


pages/experiments/[id].tsx
```diff
@@ -30,7 +31,10 @@ export default function ExperimentPage() {
   debug(`ExperimentPage#render ${experimentId}`)
 
   const { isLoading: experimentIsLoading, data: experiment, error: experimentError } = useDataSource(
-    () => ExperimentsApi.findById(experimentId),
+    () =>
+      experimentId
+        ? ExperimentsApi.findById(experimentId)
+        : ((new Promise(() => null) as unknown) as Promise<ExperimentFull>),
```
Comment (yanirs) [inline pages/experiments/[id].tsx:37]: This gets a bit repetitive and hard to read. What do you think of adding a helper to at least create the promise for the null case? Can probably also replace lines 34-37, but I'm not sure how messy it'd get without playing with it.

Another thing that comes to mind is that we always call `useDataSource()` with an API function wrapper as the first argument and the arguments to the API function as the second argument, so there may be an opportunity to simplify things even further. For example, this call would become `useDataSource(ExperimentsApi.findById, [experimentId])`, and `useDataSource()` would handle calling `ExperimentsApi.findById` with `[experimentId]` and the case where any arguments are null. Again, I'm not sure how messy the implementation would get, but it'd definitely make all the calls cleaner so it may be worth the effort. To keep things simple for now, we can only handle the cases where the number of arguments is zero or one.

yanirs:
> On that topic my general preference is to drop NextJS and not do any pre-rendering.

Yeah, can consider it, but it's not a major priority for now. I added #212 to remind us to look into it later.

pages/experiments/[id].tsx
Comment (jessie-ross) [reply pages/experiments/[id].tsx:37]: > This gets a bit repetitive and hard to read. What do you think of adding a helper to at least create the promise for the null case? Can probably also replace lines 34-37, but I'm not sure how messy it'd get without playing with it.

Sure I can do that, and it would make it more readable:
```createUnresolvingPromise= () => new Promise(() => null) as unknown```

Interestingly, the `as unknown` part is only necessary as without it prettier strips out the `as Promise<ExperimentFull>` even though typescript doesn't complain.

> Another thing that comes to mind is that we always call useDataSource() with an API function wrapper as the first argument and the arguments to the API function as the second argument, so there may be an opportunity to simplify things even further. For example, this call would become useDataSource(ExperimentsApi.findById, [experimentId]), and useDataSource() would handle calling ExperimentsApi.findById with [experimentId] and the case where any arguments are null. Again, I'm not sure how messy the implementation would get, but it'd definitely make all the calls cleaner so it may be worth the effort. To keep things simple for now, we can only handle the cases where the number of arguments is zero or one.

I don't think `ExperimentsApi.findById` should be responsible for that case and it should result in an argument error - it would be worth throwing an error in fact.

The problem here is that the router is calling a component with a null argument, we don't want to pass that problem down.

Comment (yanirs) [reply pages/experiments/[id].tsx:37]: > I don't think `ExperimentsApi.findById` should be responsible for that case and it should result in an argument error - it would be worth throwing an error in fact.

I didn't mean that `ExperimentsApi.findById()` would be responsible for that case, I meant that the hook would handle it, but I don't mind just making it more readable with `createUnresolvingPromise` for now.

Comment (jessie-ross) [reply pages/experiments/[id].tsx]: Ahh I think I see what you mean.

Similarly, I am not a fan of extending useDataSource either and don't think it should have the responsibility of checking arguments, as well as the `deps` parameter being funky enough. In the future we may have more complicated data sources, multiple parameters etc.

I am going to go with `createUnresolvingPromise`

File: utils/general.ts [hunk 2]
```diff
@@ -1,9 +1,12 @@
...
-
 /**
  * Coerces values to booleans and performs an `or` operation across them.
  * @param xs Any values
  */
 export function or(...xs: unknown[]) {
   return xs.reduce((acc, x) => acc || !!x, false) as boolean
 }
+
+export function createUnresolvingPromise() {
```
yanirs, line 10: Maybe add a type parameter to this function to make it return Promise<Type> to obviate the need for `as Promise<Type>` when it's called? Also, a documentation that includes a quick explanation why this is needed would be nice. :slightly_smiling_face:

jessie-ross: The `as Promise<Type>` is necessary because otherwise the data that is returned will be typed as `any` instead of `Type`.

I've added the docstring.

jessie-ross: Ah I think I misread you, yeah I will add a type parameter.

yanirs: Yeah, I meant something like this:

```typescript:
function createUnresolvingPromise<T>() {
  return new Promise<T>(() => null)
}
```

PR #210, opened 2020-07-08
Add BasicInfo form part

jessie-ross:
Current Look:

Screen Shot 2020-07-08 at 10 34 08 pm

(It will eventually be embedded in a `Page` component.)

**Thoughts:**
- What about using monospace all round for data input and output?
- Using a text field instead of a dropdown for owner, at least at first?
- Ah I need to capitalise the D in "date"

Part of #9 

## How has this been tested?
- MT & AT


jessie-ross:
Ah also note that I am using the native date-picker for simplicity: 
https://material-ui.com/components/pickers/#native-pickers
https://caniuse.com/#feat=input-datetime

Notably it doesn't work on Safari but I think this should be fine for the MVP.

yanirs:
Looks good! Some quick thoughts:

* It'd be nice to make the fields wider so that they're aligned on the right. Looks like a bit of a pyramid right now. :slightly_smiling_face: 
* We should provide more space in the description text area and/or make it expandable. This will be partly addressed by making it wider to match the end of the `End date` input.
* In my opinion, `Sentence case` would look better than `Title Case` for the labels (i.e., rather than capitalising the `D`s for the date fields we should downcase the `N` and `D` in `Experiment Name` and `Experiment Description`).
* We don't need `Please` in helper texts. It's nice to be polite, but consistency and brevity would be better in this case.
* How easy would it be to add basic constraints to the date picker? The start date shouldn't be in the past and the end date should be after the start date.
* Agreed on no dropdown for the owner. We will need to add an API endpoint to provide the auto-completion values and the default value, so it can definitely be kept for later. Also, with over a thousand potential owners, I think it will never be a dropdown. :slightly_smiling_face: 
* To simplify things, let's make the placeholder for the owner `scjr` rather than `@scjr` so we don't have to strip the `@` when making the API call. The helper text for that field can be `WordPress.com account name`, or something along these lines.
* I don't feel strongly about monospace everywhere, but I do feel strongly about repeating it everywhere if we go down this path. As noted in a previous PR, we should be able to style components on the theme level.

Should get opinions from @aaronyan and @withinboredom as well.

jessie-ross:
> It'd be nice to make the fields wider so that they're aligned on the right. Looks like a bit of a pyramid right now. 🙂

Sure I will try that out.

> We should provide more space in the description text area and/or make it expandable. This will be partly addressed by making it wider to match the end of the End date input.

I'll play around with it.


> In my opinion, Sentence case would look better than Title Case for the labels (i.e., rather than capitalising the Ds for the date fields we should downcase the N and D in Experiment Name and Experiment Description).

👍 

> We don't need Please in helper texts. It's nice to be polite, but consistency and brevity would be better in this case.

👍 

> How easy would it be to add basic constraints to the date picker? The start date shouldn't be in the past and the end date should be after the start date.

It is not hard: https://github.com/jquense/yup/issues/287#issuecomment-501481042

> Agreed on no dropdown for the owner. We will need to add an API endpoint to provide the auto-completion values and the default value, so it can definitely be kept for later. Also, with over a thousand potential owners, I think it will never be a dropdown. 🙂
> To simplify things, let's make the placeholder for the owner scjr rather than @scjr so we don't have to strip the @ when making the API call. The helper text for that field can be WordPress.com account name, or something along these lines.

I will add an adornment for the `@`: https://material-ui.com/components/text-fields/#input-adornments

> I don't feel strongly about monospace everywhere, but I do feel strongly about repeating it everywhere if we go down this path. As noted in a previous PR, we should be able to style components on the theme level.

Yep I will be changing this on a high level.
(MaterialTable happens to be annoying about it.)

withinboredom:
> In my opinion, Sentence case would look better than Title Case for the labels (i.e., rather than capitalising the Ds for the date fields we should downcase the N and D in Experiment Name and Experiment Description).

I'd rather us use CSS to control label casing and not get too worried about this. We can capitalize the first letter (`::first-letter` selector) and force everything to be lower case (`text-transform: lowercase`), or have each letter be upper case (`text-transform: capitalize`). This gets us consistency for free and we can always change it later without a huge diff. Same with monospace.

> We should provide more space in the description text area and/or make it expandable.

Instead of using a textarea, try experimenting with a `div` with `contenteditable` which should auto-expand the edit area as you type (for free).

jessie-ross:
I have made @yanirs suggested changes.
This is what it looks like now:
Screen Shot 2020-07-09 at 5 02 56 pm

jessie-ross:
In response to @withinboredom:
 
> I'd rather us use CSS to control label casing and not get too worried about this. We can capitalize the first letter (::first-letter selector) and force everything to be lower case (text-transform: lowercase), or have each letter be upper case (text-transform: capitalize). This gets us consistency for free and we can always change it later without a huge diff. Same with monospace.

I am mostly going for design at this point and will refactor later.

> Instead of using a textarea, try experimenting with a div with contenteditable which should auto-expand the edit area as you type (for free).

How do you feel about the last solution?

withinboredom:
> How do you feel about the last solution?

It looks fine, but may be annoying for editing larger form descriptions. It's also annoying when trying to scroll a page with an embedded scrollbar and you don't realize your mouse is on the inner part so the page doesn't scroll. I think it's fine though as I don't imagine most descriptions will be too long (there's a p2 url for that).

aaronyan:
Very nice, this looks great! I really like the overall look. A few suggestions:
- For the owner hint, let's add "Use WordPress.com username" since we don't autocomplete atm
- Pretty minor detail: In the second revision, there are gaps between the field name and the line at the top of the field. It would be good to close that gap up :)
Screen Shot 2020-07-09 at 4 37 35 PM

jessie-ross:
Following @aaronyan's comments

> For the owner hint, let's add "Use WordPress.com username" since we don't autocomplete atm

Good point, I've just added it in. And I think as a future feature we should pre-fill this with the current user.

> Pretty minor detail: In the second revision, there are gaps between the field name and the line at the top of the field. It would be good to close that gap up :)

Ah I can't un-see this now 😱 . Nice attention to detail!

This is likely a bug in MUI as I am not doing anything funky with this component and it doesn't look like there are any style conflicts going on. 
I will post an issue to MUI, since this is a public project I will even reference this issue.

**Latest Look:**

Screen Shot 2020-07-10 at 1 27 27 pm

jessie-ross:
I have added it to the `/experiments/new` page and you can play with it here:
https://abacus-git-add-experiment-creation-form-basic-info-part.a8c.vercel.app/experiments/new

Nice validation and the data output will be fixed at a later point.

jessie-ross:
While I was putting in @withinboredom' s suggestion, found that all lower-case labels looks quite nice:

Screen Shot 2020-07-14 at 12 32 21 am

Other than that suggestion I think this is ready to go.

components/experiment-creation/BasicInfo.tsx
```diff
@@ -0,0 +1,129 @@
...
+        color: theme.palette.text.hint,
+      },
+    },
+  }),
+)
+
+const BasicInfo = () => {
+  const classes = useStyles()
+
+  // TODO: COMPONENT
```
yanirs, line 37: Can we remove this?

PR #211, opened 2020-07-08
Tweak experiment results and improve fixtures

yanirs:
Another PR carved out of #199 with the following changes:

* Adjust `FullLatestAnalyses` to handle the case where some metric assignments don't have analyses yet.
* Change the wording in `RecommendationString` to be independent of the experiment's status (e.g., saying "end experiment" when the experiment has already ended isn't helpful).
* Minor tweaks to the `ExperimentResults` component.
* Extend the test fixtures to cover more cases and reflect the schema better.

## How has this been tested?

* Inspected the rendered changes in Storybook.
* Verified and adjusted the snapshots and automated tests to match the fixture and component changes.

### DIFF
--- a/components/MetricAssignmentsPanel.test.tsx
+++ b/components/MetricAssignmentsPanel.test.tsx
@@ -119,6 +119,58 @@ test('renders as expected with all metrics resolvable', () => {
                 No
               </td>
             </tr>
+            <tr
+              class="MuiTableRow-root"
+            >
+              <td
+                class="MuiTableCell-root MuiTableCell-body"
+              >
+                metric_2
+              </td>
+              <td
+                class="MuiTableCell-root MuiTableCell-body"
+              >
+                <span>
+                  $0.50
+                </span>
+              </td>
+              <td
+                class="MuiTableCell-root MuiTableCell-body"
+              >
+                1 hour
+              </td>
+              <td
+                class="MuiTableCell-root MuiTableCell-body"
+              >
+                Yes
+              </td>
+            </tr>
+            <tr
+              class="MuiTableRow-root"
+            >
+              <td
+                class="MuiTableCell-root MuiTableCell-body"
+              >
+                metric_3
+              </td>
+              <td
+                class="MuiTableCell-root MuiTableCell-body"
+              >
+                <span>
+                  12 pp
+                </span>
+              </td>
+              <td
+                class="MuiTableCell-root MuiTableCell-body"
+              >
+                6 hours
+              </td>
+              <td
+                class="MuiTableCell-root MuiTableCell-body"
+              >
+                Yes
+              </td>
+            </tr>
           </tbody>
         </table>
       </div>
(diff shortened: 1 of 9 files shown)


components/MetricAssignmentsPanel.test.tsx
```diff
@@ -119,6 +119,58 @@ test('renders as expected with all metrics resolvable', () => {
                 No
               </td>
             </tr>
+            <tr
```
yanirs, line 122: Added more metric assignments to the default `ExperimentFull` fixture.

components/MetricsTable.test.tsx
```diff
@@ -51,12 +51,12 @@ test('with some metrics, loads and opens metric details', async () => {
   const tBodyElmt = container.querySelector('tbody') as HTMLTableSectionElement
   expect(tBodyElmt).not.toBeNull()
 
-  for (let i = 0; i < 6; i++) {
+  for (let i = 1; i < 7; i++) {
```
yanirs, line 54: 0 isn't a valid metric ID. :slightly_smiling_face:

components/__snapshots__/ExperimentDetails.test.tsx.snap
```diff
@@ -228,6 +228,58 @@ exports[`renders as expected at large width 1`] = `
                     No
                   </td>
                 </tr>
+                <tr
```
yanirs, line 231: This is a bit annoying because the `ExperimentDetails` and the individual panel tests seem to cover the same cases and snapshots. We should look into reducing this testing duplication as part of moving components to subdirectories.

components/experiment-results/ExperimentResults.tsx
```diff
@@ -34,14 +34,11 @@ export default function ExperimentResults({
   )
 
   if (analyses.length === 0) {
-    return <h2>No analyses yet for {experiment.name}.</h2>
+    return <p>No analyses yet for {experiment.name}.</p>
```
yanirs, line 37: We should MUIfy all the `/experiment-results` components at a later stage. For now, I just removed the main headings because they seem unnecessary with all the other headings on the page.

helpers/fixtures.ts
```diff
@@ -283,7 +351,7 @@ function createMetricFull(id: number) {
     name: `metric_${id}`,
     description: `This is metric ${id}`,
     parameterType,
-    higherIsBetter: id % 3 === 0 ? true : false,
+    higherIsBetter: id % 3 === 0,
```
yanirs, line 354: WebStorm noted that this was already a boolean. :slightly_smiling_face:

helpers/fixtures.ts
```diff
@@ -265,7 +333,7 @@ function createMetricBare(id: number) {
 }
 
 function createMetricBares(numMetrics = 3) {
-  return _.range(numMetrics).map(createMetricBare)
+  return _.range(1, numMetrics + 1).map(createMetricBare)
```
yanirs, line 336: Using legal metric IDs again.

helpers/fixtures.ts
jessie-ross: Ha nice

PR #214, opened 2020-07-12
Move inline snapshots out of ExperimentResults test

yanirs:
Branching out of #199, this PR moves the inline snapshots for the `ExperimentResults` component to the snapshot file.

## How has this been tested?

Automated tests pass as before.


yanirs:
> I am still not sure what the benefits of inline snapshots are?
> On the other hand I feel non-inline makes the test files cleaner and easier to read.

I'm not too fussed, but I think they're nice for very small snapshots like `ExperimentStatus` as there's less jumping around. This PR is a clear a case where inlining wasn't the best idea. :slightly_smiling_face:

PR #216, opened 2020-07-14
Add experiment creation form ui audience

jessie-ross:
**This PR implements the Audience form-part for Experiment Creation:**

Screen Shot 2020-07-14 at 11 47 47 pm

- Video: https://d.pr/v/sJ4Zat
- Implemented the basic fields
- Implemented Segment using [MUI (lab) Autocomplete](https://material-ui.com/components/autocomplete/)

**Still to come:**
- Overall form look and feel
- Validation
- I will likely unnest the field names later on

Part of a series on #9 

## How has this been tested?
- MT
- Having trouble with AT
   - Not sure how to access the React Portal for the Segments Select
   - MUI Id's are changing


jessie-ross:
By the way, the Segmentation field was from @dhurlburtusa's suggestion, I think it works quite well

aaronyan:
Looking good so far!

> By the way, the Segmentation field was from @dhurlburtusa's suggestion, I think it works quite well

I like the use of Chips! Let's change the hint on the Segments form to be something like "Search and select a country or locale". And maybe we remove the triangle on the right so we don't indicate it's a list, but rather a search box to type in? Search and select seems like a good move to accommodate the many countries and locales we offer.

yanirs:
Only looked at the video and skimmed the code. I'll wait until the schema is in to have a closer look, but generally it looks good. There are a couple of things missing, though:

* How do we set segments as excluded? Should have two separate lists?
* We ended up agreeing that setting the variations would be under the audience part: pbmo2S-kT-p2#comment-698. This is consistent with [what we show under Audience on the details view](https://abacus-git-add-experiment-creation-form-ui-audience.a8c.vercel.app/experiments/112). Can you please add it to this PR?

jessie-ross:
> How do we set segments as excluded? Should have two separate lists?

Good point, I missed that.
I think two lists would work.

> We ended up agreeing that setting the variations would be under the audience part: pbmo2S-kT-p2#comment-698. This is consistent with what we show under Audience on the details view. Can you please add it to this PR?

👍

jessie-ross:
Ok, I have updated the PR:
- Created "New" schemas for various types.
- Integrated this PR with the changes from the Schema refactor.
- Added Variations.

As I don't want this PR to become too big, I plan to create some further PRs from this one:
- Turn on UI-level validation
- Populate segments from the real data (I plan to use `normalizr` for this).
- Allow segments to be excluded, I am still thinking of how best to do this, I feel like there should be a better way. There is 

Things to note:
- There is a feature switch `ALLOW_ADDITIONAL_VARIATIONS` in `Audience.tsx` to allow adding and editing variations (except for `control`). In the future feature switches will be centralised.
- See the use of `_key`

@aaronyan 

> I like the use of Chips! Let's change the hint on the Segments form to be something like "All visitors by default. Search and select to customize". And maybe we remove the triangle on the right so we don't indicate it's a list, but rather a search box to type in? Search and select seems like a good move to accommodate the many countries and locales we offer.

I have changed the placeholder text, I think that is what you meant. I couldn't fit it all in, not sure if I am happy with it. I will be revisiting this after populating the segments and after adding the Exclude Segment feature.


The current look:

Screen Shot 2020-07-16 at 3 32 54 pm

aaronyan:
Looking good, thanks @jessie-ross! If you're looking for suggestions on how to add the exclude segment feature, I thought about it some and created some mocks! Feel free to use or implement what you had in mind.
## State when `include` is selected
Screen Shot 2020-07-16 at 1 44 07 PM


## State when `exclude` is selected
Screen Shot 2020-07-16 at 1 44 03 PM


I think it would be good to rename this section to `Targeting`, but happy to discuss. And once a country or locale is added, we should remove the placeholder. I think the placeholder currently remains, even after a segment is added.

jessie-ross:
@aaronyan, that looks like a great improvement!

I love its simplicity - having only include or exclude.
If I am understanding correctly, it would make sense we would either be whitelisting or blacklisting, and doing both at once wouldn't make sense.

> I think it would be good to rename this section to Targeting

I think this works well!

> And once a country or locale is added, we should remove the placeholder. I think the placeholder currently remains, even after a segment is added.

Yeah this was annoying me as well. It is a bug in MUI and I can look into it further.

jessie-ross:
@yanirs do you have any further feedback for this?

yanirs:
> @yanirs do you have any further feedback for this?

I was waiting to see where the discussion with @aaronyan lands. :slightly_smiling_face: I can think of cases where both segment exclusions and inclusions would be needed (e.g., all EN users except for those in the US), but I'm happy to allow only inclusions or exclusions at this stage because supporting sophisticated targeting may be a case of YAGNI. We can always add it later as two lists if there's demand for it. I'll add some quick inline notes and do a full review once it's fully implemented.

.eslintrc.js
```diff
@@ -151,6 +151,7 @@ module.exports = {
       {
         selector: 'property',
         format: ['snake_case', 'strictCamelCase', 'StrictPascalCase', 'UPPER_CASE'],
+        leadingUnderscore: 'allow',
```
yanirs, line 154: What do you think of using a _trailing_ underscore for client-side attributes? Accessing `obj._attribute` still feels wrong to me given the prevalence of leading underscore use to denote limited visibility. This might be my Python bias, though. :slightly_smiling_face: The idea of a trailing underscore also came to me from a Python package: In `scikit-learn`, model attributes that are calculated after construction are denoted with a trailing underscore (e.g., `model.feature_importances_`).

jessie-ross: Yeah that works for me :)

components/experiment-creation/Audience.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+      }
+    >
+      {(formikProps: FormikProps<{ experiment: Partial<ExperimentFullNew> }>) => <Audience formikProps={formikProps} />}
+    </Formik>,
+  )
+  expect(container).toMatchSnapshot()
+
+  fireEvent.click(screen.getByRole('combobox'))
+
+  await waitFor(() => screen.getByText(/en-AU/))
```
jessie-ross, line 27: This line doesn't seem to work.

jessie-ross:
**PR Update**
- New radio exclusion field for the segments, take a look
- Using trailing underscores for object data

Planned future PRs:
- Turn on UI-level validation
- Populate segments from the real data (I plan to use normalizr for this).

What is missing:
- The testing, I am struggling with testing the segments, can't seem to match on a React Portal
   @yanirs can you take a look, see my inline comment on `Audience.test.tsx`

lib/schemas.ts
```diff
@@ -112,8 +119,17 @@ export const segmentSchema = yup
...
+export const segmentAssignmentNewSchema = yup
   .object({
+    segmentId: idSchema.defined(),
+    isExcluded: yup.bool().defined(),
+  })
+  .defined()
+  .camelCase()
+export type SegmentAssignmentNew = yup.InferType<typeof segmentAssignmentNewSchema>
+
+export const segmentAssignmentSchema = segmentAssignmentNewSchema
```
yanirs, line 131: Let's keep things consistent and drop the fields that are already defined (as done for `metricAssignmentSchema`).

lib/schemas.ts
```diff
@@ -78,23 +78,30 @@ export enum AttributionWindowSeconds {
...
     minDifference: yup.number().defined(),
   })
   .defined()
   .camelCase()
+export type MetricAssignmentNew = yup.InferType<typeof metricAssignmentNewSchema>
+
+export const metricAssignmentSchema = metricAssignmentNewSchema
+  .shape({
+    metricAssignmentId: idSchema.defined(),
+    experimentId: idSchema.defined(),
```
yanirs, line 101: Should this be dropped now (below as well)?

lib/schemas.ts
```diff
@@ -123,15 +139,27 @@ export const segmentAssignmentSchema = yup
...
+export const variationNewSchema = yup
   .object({
-    variationId: idSchema.defined(),
     name: nameSchema.defined(),
     isDefault: yup.bool().defined(),
     allocatedPercentage: yup.number().integer().min(1).max(99).defined(),
   })
   .defined()
   .camelCase()
+export type VariationNew = yup.InferType<typeof variationNewSchema> & { key_?: DefaultVariationKey | number } // For client use only
```
yanirs, line 155: I don't understand what the `key_` is used for. Can you please explain? If it's only needed to support more than two variations, let's omit it for now.

components/experiment-creation/Audience.tsx
```diff
@@ -0,0 +1,317 @@
...
+  Platform,
+  Segment,
+  SegmentAssignmentNew,
+  SegmentType,
+  VariationNew,
+} from '@/lib/schemas'
+import { SegmentTypeToHuman } from '@/lib/segments'
+
+// TODO: Add to feature flag object
+const ALLOW_ADDITIONAL_VARIATIONS = false
```
yanirs, line 40: There's already a lot going on. I think we should drop this, as we're pretty far away from supporting multiple variations &ndash; it'd be a major end-to-end change.

components/experiment-creation/Audience.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      maxWidth: '36rem',
+      // TODO: Remove, this is just for the storybook.
```
yanirs, line 46: This is getting a bit repetitive. :slightly_smiling_face: Can we add it to the styles injected in storybook?

components/experiment-creation/Audience.tsx
```diff
@@ -0,0 +1,317 @@
...
+    },
+  }),
+)
+
+enum SegmentExclusionState {
+  Exclude = 'exclude',
+  Include = 'include',
+}
+
+const segments: Record<number, Segment> = {
```
yanirs, line 74: Can you please add an issue for populating the segments from real data and reference it here?

components/experiment-creation/Audience.test.tsx
yanirs: :man_shrugging: I wonder if there's an easy way to see what the test is seeing. [`screen.debug()`](https://testing-library.com/docs/dom-testing-library/api-queries#screendebug) might be helpful here.

lib/schemas.ts
jessie-ross: Yep that's right but it will show in the merge and I usually deal with this sort of thing there.

components/experiment-creation/Audience.tsx
jessie-ross: I thought I would keep it in since I had already written the front-end for it?

lib/schemas.ts
jessie-ross: Ah oops

components/experiment-creation/Audience.tsx
yanirs: I'd rather we focus on implementing what we need right now rather than add code that's inaccessible and will remain inaccessible for the foreseeable future. We can keep the code for later by adding it to a separate branch that we can revive if and when we'll get to it. Adding this right now just adds to the review and maintenance overheads.

lib/schemas.ts
yanirs: I'm not sure it'll show up in the merge given that there are no conflicts and that the `experimentId` issue only popped up in production.

jessie-ross: Shouldn't there be a conflict as experimentId was deleted in master?

lib/schemas.ts
jessie-ross: Yeah it is necessary for more-than-two-variations version.

components/experiment-creation/Audience.tsx
jessie-ross: I have created and pushed a new branch for this:
https://github.com/Automattic/abacus/tree/add/experiment-creation-form-ui-audience-with-many-variations

lib/schemas.ts
yanirs: :man_shrugging: I was just going by what GitHub says &ndash; usually it notes potentially merge conflicts. Anyway, it's just something to remember so we don't reintroduce the bug. Should probably just change Swagger to match production so we can catch this sort of stuff earlier.

jessie-ross: I have created an issue with the ExPlatform repo

components/experiment-creation/Audience.tsx
jessie-ross: https://github.com/Automattic/abacus/issues/230

components/experiment-creation/Audience.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  // The segmentExclusion code is currently split between here and SegmentAutocomplete
+  // An improvement might be to have SegmentAutocomplete only handle Segment[] and for code here
+  // to translate Segment <-> SegmentAssignment
+  const [segmentAssignmentsField, _segmentAssignmentsFieldMeta, segmentAssignmentsFieldHelper] = useField(
+    'experiment.segmentAssignments',
+  )
+  const [segmentExclusionState, setSegmentExclusionState] = useState<SegmentExclusionState>(
+    SegmentExclusionState.Include,
+  )
```
yanirs, line 120: Were the trailing commas added by Prettier?

components/experiment-creation/Audience.tsx
```diff
@@ -0,0 +1,270 @@
...
+            />
+          </Field>
+        </FormControl>
+      </div>
+      <div className={classes.row}>
+        <FormControl component='fieldset' className={classes.segmentationFieldSet}>
+          <FormLabel htmlFor='segments-select'>Targeting</FormLabel>
+          <FormHelperText className={classes.segmentationHelperText}>
+            Who should see this experiment?
+            <br />
```
yanirs, line 185: There are some lint warnings on this file. Can you please fix them?

In general, we should probably fail on warnings, as accumulating warnings just means that they're all effectively ignored.

components/experiment-creation/Audience.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+  setStart: () => undefined,
+  setEnd: () => undefined,
+  // @ts-ignore; This is just for mocking
+  commonAncestorContainer: {
+    nodeName: 'BODY',
+    ownerDocument: document,
+  },
+})
+
+test('renders as expected', async () => {
```
yanirs, line 22: This test generates many console warnings. We should address it for existing tests as part of #229, but let's not add new warnings in the meantime.

components/experiment-creation/Audience.test.tsx
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable no-irregular-whitespace,@typescript-eslint/ban-ts-ignore */
```
yanirs, line 1: Why is `no-irregular-whitespace` disabled? Also, please make the `ban-ts-ignore` suppression more specific by removing this and adding `// eslint-disable-next-line @typescript-eslint/ban-ts-ignore` before line 15.

components/experiment-creation/Audience.tsx
jessie-ross: Yep as Prettier split these into multiline.

I am for trailing commas for multiline function arguments.

components/experiment-creation/Audience.test.tsx
jessie-ross: `no-irregular-whitespace` was needed for the inline snapshots so I'll remove it now.

components/experiment-creation/Audience.tsx
jessie-ross: I've fixed these resulting in less readable code but the problem is in the rule.
This is not worth spending any amount of time on though.

components/experiment-creation/Audience.test.tsx
jessie-ross: Most of these are a problem within a lib we use - either `material-ui` or `formik-material-ui`.
Should I be making a fork of their repo, fixing it, PRing and in the meantime switch this repo to use my fork?

components/experiment-creation/Audience.tsx
yanirs: In general, we can disable rules that are not helpful. My point is that output that's full of warnings becomes useless over time.

components/experiment-creation/Audience.test.tsx
yanirs: For problems with third-party libraries, it'd be good to add issues on their repos and note them in #229. I don't think it's worth using a private fork for the console warnings, as they're not blocking the tests.

components/experiment-creation/Audience.tsx
jessie-ross: Well warnings are warnings for a reason - they are not necessarily a problem.
As I mentioned in #229 there is a middle ground between fixing every warning and ignoring them all which I think would be more efficient.

In this case the rule is useful but the rule itself would need to be PR'd to add the `br` element.

yanirs: > Well warnings are warnings for a reason - they are not necessarily a problem.

If we ignore warnings and allow them to accumulate, we're not going to notice new warnings, which is a problem. You can also argue that running ESLint and using TypeScript results in a bunch of "warnings", but we choose to use them because they do increase quality and maintainability over time.

> As I mentioned in #229 there is a middle ground between fixing every warning and ignoring them all which I think would be more efficient.

I disagree with it being more efficient to ignore warnings in code that was just added. The time to address them is when it's fresh in our minds. See https://google.github.io/eng-practices/review/reviewer/pushback.html#later.

PR #221, opened 2020-07-17
Add experiment creation form validation

jessie-ross:
Turn PR implements validation:
- Use validation schema in the form
- Improve validation messages
  Tried a few options for this, but this seemed to be the best.
  Unfortunately this makes the run-time server-validation errors less informative, I have included a comment on how to fix this if we do want it.
- Adds improved validation in the schemas
  - `name` fields now use `snake_case`
  - Date fields have to be in the future and end must be after start.

## How has this been tested?
- MT & AT
- More thorough testing will happen with e2e tests


jessie-ross:
Please take a look, the tests don't pass but it is the mock server that needs to be fixed.

lib/validation.ts
```diff
@@ -0,0 +1,64 @@
+/* istanbul ignore file; this is just config, will be tested through integration */
+/* (istanbul wouldn't let me just ignore the function) */
```
yanirs, line 2: If we're ignoring the full file, let's add it to the `jest.config.js` ignore list.

lib/schemas.ts
```diff
@@ -146,12 +149,23 @@ export enum Status {
...
     experimentId: idSchema.defined(),
     name: nameSchema.defined(),
     startDatetime: yup.date().defined(),
-    endDatetime: yup.date().defined(),
+    endDatetime: yup
+      .date()
+      .defined()
+      .when(
+        'startDatetime',
+        /* istanbul ignore next; should be e2e tested */
```
yanirs, line 165: Maybe add the test to the PR?

lib/schemas.ts
```diff
@@ -146,12 +149,23 @@ export enum Status {
...
+        (startDatetime: Date, schema: yup.DateSchema) =>
+          startDatetime && schema.min(startDatetime, 'End date must be after Start date.'),
```
yanirs, line 167: Capitalising `Start` in the middle of the sentence is a bit awkward. :slightly_smiling_face:

jessie-ross: This particular line actually prevents end-dates that are before the start date, and it happens before submission on field change. You can test it out in the Vercel Deployment.

lib/validation.ts
jessie-ross: Well my intention here is not to ignore the whole file. I wanted to ignore just the function but I couldn't get `istanbul` to ignore. (Ah and I meant e2e instead of integration).

I think it is useful to keep it here as documentation although I might change the lines around a bit:

```
/* istanbul ignore file; istanbul wouldn't ignore just the setupYupUiValidationMessage */

[...]

// This is mostly config so will be tested through e2e.
export function setupYupUiValidationMessages() {
```

lib/schemas.ts
jessie-ross: I did mean e2e over the whole form, so I think it makes sense to do so after the form is a bit more together.

lib/schemas.ts
```diff
@@ -169,17 +183,18 @@ export const experimentFullSchema = experimentBareSchema
...
   })
   .defined()
   .camelCase()
 export type ExperimentFull = yup.InferType<typeof experimentFullSchema>
 
-// Just a stub for now
 export const experimentCreateSchema = experimentFullSchema.shape({
   experimentId: idSchema.nullable(),
+  // Using yesterday here to avoid timezone issues
+  startDatetime: yup.date().defined().min(yesterday, 'Start date must be in the future.'),
```
jessie-ross, line 197: Here we ensure startDates are mostly in the future.
I have made a bit of a compromise here for simplicity, might not be the right choice here actually.

A proper solution would be using UTC strictly and informing the user of that. I think it would be good to do so throughout the App.

jessie-ross:
Just to bring my individual comments together:

> It'd be good to add tests as we go.

I am planning to do form testing after it is in one piece, right now it is scattered through a few PRs

> We should probably keep client-side validation fairly light at this stage, as we do it all on the server side. 

It looks like I need to change the schema a little as it doesn't match the production data but I wouldn't say we need to remove any other than that.

> In any case, not all the validations can be done on the client side. I'm happy to wait with additional validations and other polishing until we can create and edit experiments from the UI.

I am not sure which validations I have missed? They are quite simple to put in. 

> Taking the dates as an example, the UX is still not great since we allow users to select past dates and end dates that are before the start date. It'd be better to make it impossible to make wrong choices rather than rely on validations that happen when the form is submitted. What do you think?

So currently we do ensure that start dates are at least "yesterday" (see the comment), and that end dates are after start dates.

And validation is currently happening after each field.

Eventually it will happen on three user levels:
1. Field Level
2. Form-Part Level - might block the next button/next button might trigger it - possibly [using this MUI Stepper error state feature](https://material-ui.com/components/steppers/#non-linear-error-step) depending on how the Stepper PR goes. 
3. Form Level - When submit is clicked, it will just explain there is an error and to look in form-parts. Form-part error state and field error state would already be in place before the submit error occurs.

lib/schemas.ts
yanirs: Makes sense. Can you please start an issue to track all these cases that need to be tested? It's easy to lose inline comments.

lib/schemas.ts
yanirs: Yeah, I think it should be made clear that we're working in UTC. In any case, the server will reject invalid dates that somehow make it through the validation.

lib/validation.ts
yanirs: If this file is going to contain any other functionality, then we shouldn't suppress coverage. Thinking about it more, as this is a single-use function, why not inline it in `_app.tsx`? I just tried it locally and it seems cleaner. We can also avoid  the casting to `unknown` with a `const`:

```typescript
const yupLocale = {
  mixed: {
    default: 'This field is invalid',
    required: 'This field is a required field',
    oneOf: 'This field must be one of the following values: ${values}',
    notOneOf: 'This field must not be one of the following values: ${values}',
    defined: 'This field must be defined',
  },
  string: {
    length: 'This field must be exactly ${length} characters',
    min: 'This field must be at least ${min} characters',
    max: 'This field must be at most ${max} characters',
    matches: 'This field must match the following: "${regex}"',
    email: 'This field must be a valid email',
    url: 'This field must be a valid URL',
    uuid: 'This field must be a valid UUID',
    trim: 'This field must be a trimmed string',
    lowercase: 'This field must be a lowercase string',
    uppercase: 'This field must be a upper case string',
  },
  number: {
    min: 'This field must be greater than or equal to ${min}',
    max: 'This field must be less than or equal to ${max}',
    lessThan: 'This field must be less than ${less}',
    moreThan: 'This field must be greater than ${more}',
    notEqual: 'This field must be not equal to ${notEqual}',
    positive: 'This field must be a positive number',
    negative: 'This field must be a negative number',
    integer: 'This field must be an integer',
  },
  date: {
    min: 'This field must be later than ${min}',
    max: 'This field must be at earlier than ${max}',
  },
  boolean: {},
  object: {
    noUnknown: 'This field has unspecified keys: ${unknown}',
  },
  array: {
    min: 'This field must have at least ${min} items',
    max: 'This field must have less than or equal to ${max} items',
  },
}
yup.setLocale(yupLocale as yup.LocaleObject)
```

yanirs:
> > It'd be good to add tests as we go.
> 
> I am planning to do form testing after it is in one piece, right now it is scattered through a few PRs

Sounds good, as long we keep track of what needs to be tested. :slightly_smiling_face: 

> > We should probably keep client-side validation fairly light at this stage, as we do it all on the server side.
> 
> It looks like I need to change the schema a little as it doesn't match the production data but I wouldn't say we need to remove any other than that.

Yep, no removal needed. :slightly_smiling_face: 

> > In any case, not all the validations can be done on the client side. I'm happy to wait with additional validations and other polishing until we can create and edit experiments from the UI.
> 
> I am not sure which validations I have missed? They are quite simple to put in.

A couple of examples that come to mind are checking that the owner is an Automattician and that there are no experiment name collisions (which can only have atomicity guaranteed on the database level). These are more on the nice-to-have side of things for now. My point was that we shouldn't go out of our way with sophisticated validations before addressing the higher priority tasks. What you added in this PR is good.

> > Taking the dates as an example, the UX is still not great since we allow users to select past dates and end dates that are before the start date. It'd be better to make it impossible to make wrong choices rather than rely on validations that happen when the form is submitted. What do you think?
> 
> So currently we do ensure that start dates are at least "yesterday" (see the comment), and that end dates are after start dates.
> 
> And validation is currently happening after each field.
> 
> Eventually it will happen on three user levels:
> 
>     1. Field Level
>     2. Form-Part Level - might block the next button/next button might trigger it - possibly [using this MUI Stepper error state feature](https://material-ui.com/components/steppers/#non-linear-error-step) depending on how the Stepper PR goes.
>     3. Form Level - When submit is clicked, it will just explain there is an error and to look in form-parts. Form-part error state and field error state would already be in place before the submit error occurs.

My point was that it's possible to go a step further and prevent some erroneous input before the validation runs. For example, disabling past dates from being selected in the date picker. None of this is critical at this stage, though.

jessie-ross:
> My point was that it's possible to go a step further and prevent some erroneous input before the validation runs. For example, disabling past dates from being selected in the date picker. None of this is critical at this stage, though.

Ahh I see what you mean, that is a good idea and a simple change - I'll add it in.

lib/schemas.ts
jessie-ross: Done, this is the issue: #225

lib/validation.ts
yanirs: @jessie-ross where do we stand on this?

components/experiment-creation/BasicInfo.tsx
```diff
@@ -34,6 +40,18 @@ const useStyles = makeStyles((theme: Theme) =>
...
+  const [startDateField] = useField('experiment.startDatetime')
+  const parseDateFromInput = (dateString: string) => dateFns.parse(dateString, 'yyyy-MM-dd', new Date())
+  const minStartDate = new Date()
+  const maxStartDate = dateFns.addMonths(new Date(), MAX_DISTANCE_BETWEEN_NOW_AND_START_DATE_IN_MONTHS)
+  const minEndDate = startDateField.value && parseDateFromInput(startDateField.value)
+  const maxEndDate =
+    startDateField.value &&
+    dateFns.addMonths(parseDateFromInput(startDateField.value), MAX_DISTANCE_BETWEEN_START_AND_END_DATE_IN_MONTHS)
+  const formatDateForInput = (date: Date) => (
+    console.log('formatDate', date), date && dateFns.format(date, 'yyyy-MM-dd')
```
yanirs, line 52: This is pretty hard to read as a one-liner. Can you please split it to lines? Also, should we use `debug` rather than `console.log()`?

jessie-ross:
Sorry this is WIP right now, I got bogged down with some of the testing details.

lib/validation.ts
jessie-ross: Sorry I didn't get back to you on this.

I generally like to keep our "entry-point" small but I am ok with putting it there for now.

components/experiment-creation/BasicInfo.tsx
jessie-ross: Sorry I didn't mean for this to end up in the PR

jessie-ross:
Only waiting on https://github.com/Automattic/experimentation-platform/pull/391 and this should be ready to go.

Most of the new stuff probably didn't need to go in this PR, I should have split this up at the very least... 

**Updates:**
- Moved the yup locale code to `_App.tsx`
- Added further testing to Dates (Must be start within a year, end within a year of start).
- Added date-picker restrictions for min and max (You can still enter invalid dates if you aren't using the date-picker, generally this is by design).
- Added testing for the date validations
- Be more explicit about dates being in UTC
- Add minimums for metricAssignments (1) and variationAssignments (2)
- Improve createExperiment integration tests

**Some Pictures:**
 
Screen Shot 2020-07-21 at 11 54 34 pm
Screen Shot 2020-07-21 at 11 54 50 pm
Screen Shot 2020-07-22 at 12 21 54 am

yanirs:
Thanks @jessie-ross! I updated the mock, but there are a bunch of unrelated test failures on CircleCI. Can you please look into it?

package.json
```diff
@@ -23,6 +23,7 @@
     "isomorphic-unfetch": "^3.0.0",
     "lodash": "^4.17.19",
     "material-table": "^1.59.0",
+    "mockdate": "^3.0.2",
```
yanirs, line 26: Shouldn't this be a dev dependency?

helpers/fixtures.ts
```diff
@@ -255,21 +257,18 @@ function createMetricAssignment(fieldOverrides: Partial<MetricAssignment>): Metr
   }
 }
 
-function createExperimentFull(fieldOverrides: Partial<ExperimentFull> = {}): ExperimentFull {
+function createExperimentFullNew(fieldOverrides: Partial<ExperimentFullCreate>): ExperimentFullCreate {
```
yanirs, line 260: Can you please bring in the name changes from #216 (`experimentCreate` -> `experimentFullNew` and `ExperimentFullCreate` -> `ExperimentFullNew`) to make this function name more consistent?

jessie-ross: Can't we leave this to fix from the merge?

__tests__/api/ExperimentsApi.test.ts
```diff
@@ -5,7 +5,7 @@ import { validationErrorDisplayer } from '@/helpers/test-utils'
 describe('ExperimentsApi.ts module', () => {
   describe('create', () => {
     it('should create a new experiment', async () => {
-      const experiment = await validationErrorDisplayer(ExperimentsApi.create(Fixtures.createExperimentFullNew()))
+      const experiment = await validationErrorDisplayer(ExperimentsApi.create(Fixtures.createExperimentFullNew({})))
```
jessie-ross, line 8: That's a strange fix for this, looks like a bug in the test coverage tool, I would rather ignore the coverage for this specific function.

yanirs: It's not a bug because coverage is run separately for each test suite. I'd rather not ignore the entire file for now, unless we start hitting more issues like this.

helpers/fixtures.ts
yanirs: I'd rather not since #216 doesn't touch this file and it's an easy fix to bring in here.

__tests__/api/ExperimentsApi.test.ts
jessie-ross: I don't really understand this part:
> because coverage is run separately for each test suite

It seems to me that even though the default parameter code path is covered the code coverage tool is saying it is not.

yanirs: The default parameter code path isn't covered in the unit tests, but it is covered in the integration tests. The test suites have different config files that look at coverage of different areas.

I actually don't have strong objections to removing test helper files from the coverage reports, but I'd rather we do it separately in a PR that'll also move them around. Looking closer at the directory structure, everything under `helpers/` is for testing, so renaming the directory and excluding it explicitly is probably the way to go. If you feel strongly about doing it now, can you please do it in a separate PR?

helpers/fixtures.ts
jessie-ross: Ok, I will deal with it here but in general I'd like to fix this sort of thing in the merge to keep the PRs seperate until they become merged into master. It simplifies the process and prevents extra back and forth on the PRs. If I didn't deal with it here it would break the type checks after merge so it wouldn't go unnoticed.

yanirs: Sure. It's not critical or blocking (which is why I approved the PR) &ndash; it just looks a bit odd because the function name half-anticipates the name change from #216.

__tests__/api/ExperimentsApi.test.ts
jessie-ross: I have created an issue for this:
https://github.com/Automattic/abacus/issues/231

PR #222, opened 2020-07-17
Add experiment creation form stepper

jessie-ross:
**This PR adds a stepper to the Experiment Creation Form:**
- https://d.pr/v/hsRJl3
- Used a Vertical Stepper due to limited vertical space due to the large header and title, long form-parts.
  - It is easy to switch it to horizontal
- All one form
- Also slid in a DebugOutput component while I was in a styling mood.

**Mobile Considerations:**
- The stepper will just disappear.

**What I consider out of scope for this PR (to be tackled in a future PR):**
- Overall app structure
   - Content width restrictions - the forms are too wide
   - Consider Medium width restrictions instead of Large
   - Title bar needs to be fixed up, I have some ideas I will post later
- Validation 
  - Coming after the validation PR
  - Preventing from progressing if there is a validation error
  - Marking a stage as complete only if it is valid.
  - Error state on stages
- Better wording on the form-parts
  - Particularly Beginning and Submission
- Refine the colour palette 
   - I've added a green as we need to differentiate the normal Affirmative buttons from a more serious Affirmative button.

**Part of a series on #9**

## How has this been tested?
- MT & AT
- Whole form functionality to be tested in later e2e tests


styles/theme.ts
```diff
@@ -96,7 +96,9 @@ const theme = createMuiTheme({
       main: '#194661',
     },
     secondary: {
-      main: '#1e77a5',
+      main: '#2e7d32',
```
jessie-ross, line 99: As mentioned in the main content, this is a temporary suggestion of a colour.
To be properly looked at in a future general design PR/P2 along with selecting light and dark variants of the Primary blue colour.

withinboredom:
Would it make more sense to just use anchor tags and "smooth scrolling"? Then we don't need to worry about height and I think this would be much simpler.

components/experiment-creation/ExperimentForm.tsx
```diff
@@ -34,23 +94,142 @@ const ExperimentForm = ({
...
+  const markStageIncomplete = (stageId: StageId) => {
+    const index = completeStages.findIndex((stageIdCur) => stageIdCur === stageId)
+    index && setCompleteStages(completeStages.splice(index, 1))
+  }
+  const [errorStages, setErrorStages] = useState<StageId[]>([])
+  const changeStage = (stageId: StageId) => {
+    // TODO: Update current stage error and complete state
+    setActiveStageId(stageId)
+    if (stageFormPartRefs[stageId].current) {
+      // Not sure why typescript is complaining about needing the '?'
```
withinboredom, line 125: typescript seems to just have issues with refs in general. At least in my experience...

yanirs:
> After getting a tour of existing systems by a person on marketing, one of the things they didn't like was having things broken up: sometimes things need to be done out of order based on what they know and the role of the person filling out the form. It was hard to know all the information you needed from the beginning. When everything is on a single page, it's easy to get information collated and can enter information in the order they get it from whatever system they're pulling it from.

Well, that mystery person is still with us so we can just ask him. :joy: 
@jessie-ross can you please run this by @eidolonnight? It's probably worth a P2 post with the demo video and link to [the Vercel form](https://abacus-git-add-experiment-creation-form-overall-style.a8c.vercel.app/experiments/new). My general feeling is that this will be fine because we will also offer single-page editing for non-staging experiments, but it's better to get feedback from our experimenters. :slightly_smiling_face:

jessie-ross:
@withinboredom 

> Would it make more sense to just use anchor tags and "smooth scrolling"? Then we don't need to worry about height and I think this would be much simpler.

The implementation I have used should have the same characteristics.

The height modifications is not about the scrolling but about hiding the rest of the form and the scrollbar.
If we want to only show one section at a time, this sort of height constraint would be needed for either way of doing it.

jessie-ross:
There has been no response so far on the P2 so I would like to proceed with this.
It is hard to change the format at a later date.

yanirs:
> There has been no response so far on the P2 so I would like to proceed with this.
> It is not hard to change the format at a later date.

+1
I'll leave it to @withinboredom to review the code. Probably worth fixing the new merge conflicts, though.

components/experiment-creation/ExperimentForm.tsx
```diff
@@ -34,23 +94,142 @@ const ExperimentForm = ({
...
+              <div className={classes.formPart} ref={formPartBasicInfoRef}>
+                <Paper className={classes.paper}>
+                  <BasicInfo />
+                </Paper>
+                <div className={classes.formPartActions}>
+                  <Button onClick={prevStage}>Previous</Button>
+                  <Button onClick={nextStage} variant='contained' color='primary'>
+                    Next
+                  </Button>
+                </div>
```
withinboredom, line 183: This bit of code can probably be a separate component, maybe used like: 
```tsx
const buttonHandlers = {previous: prevStage, next: nextStage};
<StepperButtons hasPrev hasNext {...buttonHandlers} />
```

components/experiment-creation/ExperimentForm.tsx
```diff
@@ -34,23 +94,142 @@ const ExperimentForm = ({
...
+                <Paper className={classes.paper}>
+                  <Typography variant='body1'>Audience Form Part</Typography>
+                </Paper>
+                <div className={classes.formPartActions}>
+                  <Button onClick={prevStage}>Previous</Button>
+                  <Button onClick={nextStage} variant='contained' color='primary'>
+                    Next
+                  </Button>
+                </div>
+              </div>
```
withinboredom, line 195: Can these chunks be a part of Stage as a component property using `React.forwardRef()`? It's not necessary, but it would probably clean up this component some. I also like the flatness of the component, so it's really up to you.

components/experiment-creation/Beginning.tsx
```diff
@@ -20,6 +16,7 @@ const useStyles = makeStyles((theme: Theme) =>
     p2EntryField: {
       marginTop: theme.spacing(1),
       width: '100%',
+      background: '#fff',
```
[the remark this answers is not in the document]
jessie-ross: This is specifically fixing a text-field that is displayed off of "Paper" background. Instead of having a transparent background (the page blue) I want the text field to be white. This is a bit of a one-off situation.

If we did want to use it here, `theme.palette.background.paper` would be the theme variable you are looking for.

In general I am pretty liberal with CSS... I haven't found it hard to change things like this later if we do want to start factoring the styles more.

components/experiment-creation/ExperimentForm.tsx
jessie-ross: I'm consider this as they develop more, depending on how complex they get.

components/experiment-creation/ExperimentForm.tsx
jessie-ross: Yeah I may split some of this out as it develops.

PR #227, opened 2020-07-22
PR template: Add test instructions as a reminder

yanirs:
Update the PR template to remind us of things we should test.


yanirs:
> If we are frequently testing on Production servers it might also be a good idea to add on a switch that makes it easy for us to do so. It can be a double edged sword especially if we are editing/creating data as mistakes can be made, it would be good if it was super obvious that we are using production endpoints such as with a red banner.

Yeah, I think that would be useful, especially for things like the analysis data, which is hard to mock. Added #228.

PR #232, opened 2020-07-23
Remove coverage on helpers

### LINKED ISSUE #231: Helpers should be ignored from coverage
Because we use `/helpers` in-between three test suites problems of missing coverage can arise when used in one but not the others, eg. https://github.com/Automattic/abacus/pull/221#discussion_r459181682

Lets:
- Remove `/helpers` from our coverage checking
- Unrelatedly rename `/helpers` to `/test-helpers`

jessie-ross:
Fixes #231

- Renames `/helpers` to `/test-helpers` for clarity
- Removes `/test-helpers` from code coverage

## How has this been tested?
- Existing AT passes (this is only a tests change.)


__tests__/api/AnalysesApi.test.ts
```diff
@@ -1,5 +1,5 @@
 import AnalysesApi from '@/api/AnalysesApi'
-import { validationErrorDisplayer } from '@/helpers/test-utils'
+import { validationErrorDisplayer } from '@/test-helpers/test-utils'
```
yanirs, line 2: To make this a bit less awkward, we can probably rename `test-utils.ts` to `utils.ts` now or `index.ts` so we can import from `@/test-helpers`, but not blocking for this PR.

PR #234, opened 2020-07-26
Implement MetricAssignment form-part

jessie-ross:
Screen Shot 2020-07-26 at 11 42 59 pm

Demo Vid: https://d.pr/v/4dLjMq

- Changed the form-part's title to "Goals"
- Used a percentage for Min Difference, is this what we are doing?

Part of a series on #9

## General Thoughts 

I am not too happy about the overall form of it, due to two factors:
1. The Assign A Metric being split from the Assigned Metrics can be unintuitive. 
2. Having the Metric Assignment fields in a table isn't great, it can be hard to see what field is for what and we can only provide limited info alongside each field.

I think this is good enough for an MVP but for the future I was thinking about:
- Adding a Modal step for editing and setting the Metric Assignment configuration
   This is hinted at by the original design. I don't think this is ideal as it can be annoying having to in and out of modals.
   Solves factor 1 and 2.
- Making the table fields flow vertically within a row: there would only be two cells one for the name and the other for the MetricAssignment form.
   Solves factor 1.
- Moving the Assign a Metric to the end of the AssignedMetrics Table, as the last row. 
   I tried it out for a bit but couldn't get it to look right, I am not sure what that was.
   Solves factor 2.

Also would there be any objections to changing the "24 hours" attribution window text to "1 day" for consistency with the rest?

## How has this been tested?


yanirs:
Looks good! I'll leave the full code review to @withinboredom as the assigned reviewer, but here are some thoughts.

**On correctness:**

* A metric can be assigned multiple times with different attribution windows.
* `Change expected` is independent of the the minimum difference. Experimenters always need to specify the minimum difference.
* The minimum difference is in absolute terms. It depends on the metric parameter type. See the implementation on the experiment detail page and a recent discussion on pbmo2S-oh-p2#comment-922.

**On style/content:**

* I don't think that adding the "Goals" terminology is necessary at this point, and it may be confusing. We can stick to Metrics as the section title, which is consistent with the detail page and everywhere else.
* Changing "1 day" to "24 hours" makes things a bit more ambiguous. I'd rather change "3 days" to "72 hours".
* The text feels a bit too verbose and repetitive. I played with it a bit in the browser and got to:

  image

   Changes made:
  
    * Goals -> Metrics
    * Add -> Assign
    * Removed extra headers and repetitive text
    * Widened the metric search dropdown and changed the text to "Assign a metric to quantify the experiment's impact"

  I think this would work better, though I'm sure it can be improved. It'd be good to get feedback from @aaronyan and @withinboredom on this.

components/experiment-creation/Goals.tsx
```diff
@@ -0,0 +1,284 @@
...
+                      labelId='add-metric-label'
+                      id='add-metric-select'
+                      value={selectedMetricId}
+                      onChange={onSelectedMetricChange}
+                      displayEmpty
+                    >
+                      {avaliableMetricIds
+                        .map((metricId) => normalizedMetrics[metricId])
+                        .map((metric) => (
+                          <MenuItem value={metric.metricId} key={metric.metricId}>
```
withinboredom, line 157: Can we add the `description` as a `title` attribute so the user can learn more about the metric by hovering over the name when selecting.

components/experiment-creation/Goals.tsx
```diff
@@ -0,0 +1,284 @@
...
+                </Typography>
+                <div className={classes.addMetricControls}>
+                  <FormControl className={classes.addMetricSelect}>
+                    <InputLabel id='add-metric-label'>Select a Metric</InputLabel>
+                    <MuiSelect
+                      labelId='add-metric-label'
+                      id='add-metric-select'
+                      value={selectedMetricId}
+                      onChange={onSelectedMetricChange}
+                      displayEmpty
```
withinboredom, line 152: I didn't test what it looks like without this prop, but it looks very strange when it's empty! Can we add an easter egg here? Maybe if it's empty, we add a new unselectable metric that says something funny or even just "There's no more metrics to add."

I also wonder if it's worth adding an "Add All" button. But we should probably get some real life users constantly adding all metrics first :)

components/experiment-creation/Goals.tsx
```diff
@@ -0,0 +1,284 @@
...
+} from '@material-ui/core'
+import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
+import React, { useState } from 'react'
+import { FieldArray, useField, Field } from 'formik'
+import { MetricAssignment, MetricBare, MetricParameterType, AttributionWindowSeconds } from '@/lib/schemas'
+import { TextField, Select, Switch } from 'formik-material-ui'
+import { AttributionWindowSecondsToHuman } from '@/lib/metric-assignments'
+import MoreMenu from '../MoreMenu'
+
+const normalizedMetrics: Record<number, MetricBare> = {
```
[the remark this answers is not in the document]
jessie-ross: Yeah this will be populated from the server and we should just be showing everything, see #236. We shouldn't have test data in our production server... Any chance of having another server instance with a seperate DB?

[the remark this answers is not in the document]
withinboredom: No, probably not. There's no staging servers or anything. The best we could do is have a separate database table, and then switch to the other table somehow (in production :trollface:).

components/experiment-creation/Goals.tsx
```diff
@@ -0,0 +1,284 @@
...
+  Button,
+  InputAdornment,
+} from '@material-ui/core'
+import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
+import React, { useState } from 'react'
+import { FieldArray, useField, Field } from 'formik'
+import { MetricAssignment, MetricBare, MetricParameterType, AttributionWindowSeconds } from '@/lib/schemas'
+import { TextField, Select, Switch } from 'formik-material-ui'
+import { AttributionWindowSecondsToHuman } from '@/lib/metric-assignments'
+import MoreMenu from '../MoreMenu'
```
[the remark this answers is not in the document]
jessie-ross: I haven't given much thought to imports. 
This is the first project where I have used absolute imports and I have barely noticed it since `tsserver` does all the importing renaming etc.

What is the significance of using relative vs absolute?

components/experiment-creation/Goals.tsx
jessie-ross: This is a nice idea, I'll try it out!

components/experiment-creation/Goals.tsx
jessie-ross: As Yanir mentioned above, a user should actually be able to add metric assignments with the same metric.

We will show all metrics so it will never run out 😄

components/experiment-creation/Goals.tsx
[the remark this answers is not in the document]
withinboredom: I personally favor less magic and like relative imports. There's some conversation on p2: p4TIVU-9sM-p2 that's related. We should be consistent though.

components/experiment-creation/Goals.tsx
withinboredom: Easter egg suggestion: "Yanir Approved!" with the :yanir-approved: emoji :)

aaronyan:
Exciting stuff! Looking pretty good so far @jessie-ross. Thanks for including some of your thoughts and different approaches.

Great use of the button for "change expected" by the way! I think it looks really good. I'm also a fan of the "Set as Primary" option in the ellipses.

And I think @yanirs adjustments look good! I would only adjust this part:

> Widened the metric search dropdown and changed the text to "Assign a metric to quantify the experiment's impact"

I think it looks nice to keep it simple with only the text "Assign a metric" on the dropdown form.

I think our tool can benefit from more supportive text to describe each section and do a bit more hand-holding. I think for now, we can keep it super clean, and we can figure out that extra layer of description later.

I actually kind of miss the extra words ("assigned metrics"), but I think it's just cause I like the overall look and the headers gave the card more structure. That overall look can likely be fixed by adjusting the header styling in #121 😄

yanirs:
> I actually like calling it "goals" instead of "metrics" and think we should change it everywhere :D it's much more intuitive for someone who doesn't need to know how everything works, where using metrics and metric assignments feels more like a leaky abstraction of the implementation.

We can discuss this separately as the consensus is to keep it out of this PR, but I don't think that avoiding the word _metric_ would make it more intuitive given the prevalent use of the word at Automattic.

> I think our tool can benefit from more supportive text to describe each section and do a bit more hand-holding. I think for now, we can keep it super clean, and we can figure out that extra layer of description later.

Yeah, in our case we're not launching to millions of users, so I think we can figure out what supportive text is needed through "manual" user testing, i.e., Zoom sessions with experimenters. It's easy to make a UI too busy with extra text that users end up ignoring anyway, and we're a bit too close to the problem now to be good judges of what people would find intuitive. I'm very much looking forward to iterating on the entire platform based on user feedback. :slightly_smiling_face:

components/experiment-creation/Goals.tsx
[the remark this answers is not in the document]
jessie-ross: Yeah it looks like we overwhelmingly use absolutes.

(That P2 seems to be about package-relative rather than relative to this repo.)

jessie-ross:
## PR Update

- Changed "Goal" back to "Metrics"
- Changed `minDifference` units from `%` to either `USD` or `conversions`
- Allow more than one metricAssignment with the same metric
- Moved the assign metric below table.
   This has strange behaviour with the vertical scrolling, see #238 for a fix that generally makes it easier to deal with long and extending form-parts.
- Moved `attributionWindow` behind `name` as they together "determine" a metric assignment.
   In the future might put a schema constraint on having multiple `metricAssignments` with the same `metric`, `attributionWindow` combo.  

Demo Vid: https://d.pr/v/L3hWZa

yanirs:
Looks good! 

> Changed `minDifference` units from `%` to either `USD` or `conversions`

Can we please keep this consistent with the experiment detail page? It's `$` for revenue metrics and `pp` for conversion metrics (`conversions` is incorrect): https://github.com/Automattic/abacus/blob/ae9d37b77e2a13ac0d4608dab47a126cd2250672/components/MetricAssignmentsPanel.tsx#L99-L101

It'd also be good to keep the column order and headings consistent, and use the `Label` component for the primary label. I don't mind if we change the info page headings to achieve consistency, as I find the order in this PR better.

jessie-ross:
What does `pp` stand for?

yanirs:
> What does `pp` stand for?

Percentage points. Since you asked, it's probably worth spelling it out on hover.

See: https://en.wikipedia.org/wiki/Percentage_point

jessie-ross:
## Update

- Used `pp` and `$` as units
   I decided it didn't make sense to refactor with `MetricAssignmentsPanel.tsx` as the two cases are quite different.
- Added a tooltip for `pp`
- Reorder MetricAssignmentsPanel to match the form

yanirs:
> Used pp and $ as units

Looks like the ternary condition is the wrong way around: It's showing `pp` for revenue and `$` for conversions. 

Other than that, I think it looks good from a functionality/content perspective, so I'll leave the full code review to @withinboredom.

components/experiment-creation/Metrics.tsx
```diff
@@ -0,0 +1,290 @@
...
+                      }
+
+                      const onMakePrimary = () => {
+                        makeMetricAssignmentPrimary(index)
+                      }
+
+                      return (
+                        <TableRow key={index}>
+                          <TableCell>
+                            {' '}
```
withinboredom, line 164: It seems like this should be unnecessary.

components/experiment-creation/Metrics.tsx
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  const makeMetricAssignmentPrimary = (indexToSet: number) => {
+    metricAssignmentsFieldHelperProps.setValue(
+      metricAssignmentsField.value.map((metricAssignment, index) => ({
+        ...metricAssignment,
+        isPrimary: index === indexToSet,
+      })),
+    )
+  }
```
withinboredom, line 114: Would it be better to pass the `metricAssignment` instead of the index?

```suggestion
  const makeMetricAssignmentPrimary = (metricToSet) => {
    metricAssignmentsFieldHelperProps.setValue(
      metricAssignmentsField.value.map((metricAssignment) => ({
        ...metricAssignment,
        isPrimary: metricAssignment === metricToSet,
      })),
    )
  }
```

components/experiment-creation/Metrics.tsx
```diff
@@ -0,0 +1,290 @@
...
+                              variant='outlined'
+                              autoWidth
+                              displayEmpty
+                            >
+                              <MenuItem value=''>-</MenuItem>
+                              {Object.keys(AttributionWindowSecondsToHuman).map((attributionWindowSeconds) => (
+                                <MenuItem value={attributionWindowSeconds} key={attributionWindowSeconds}>
+                                  {
+                                    AttributionWindowSecondsToHuman[
+                                      (attributionWindowSeconds as unknown) as AttributionWindowSeconds
```
withinboredom, line 189: It seems like we're abandoning some type safety here. It might be worth adding a comment as to why it's necessary and handling the case when `AttributionWindowSecondsToHuman` has an invalid index (someone may modify the html in their browser to create a custom attribution window). I don't think the API will accept a custom window, but we should handle the error gracefully in the client.

components/experiment-creation/Metrics.tsx
```diff
@@ -0,0 +1,290 @@
...
+                            </MoreMenu>
+                          </TableCell>
+                        </TableRow>
+                      )
+                    })}
+                    {metricAssignmentsField.value.length === 0 && (
+                      <TableRow>
+                        <TableCell colSpan={5}>
+                          <Typography variant='body1' align='center'>
+                            {' '}
```
withinboredom, line 248: This  space also seems unnecessary.

components/experiment-creation/Metrics.tsx
```diff
@@ -0,0 +1,290 @@
...
+                        </TableCell>
+                      </TableRow>
+                    )}
+                  </TableBody>
+                </Table>
+              </TableContainer>
+              <div className={classes.addMetric}>
+                <Add className={classes.addMetricAddSymbol} />
+                <FormControl className={classes.addMetricSelect}>
+                  {/* <InputLabel id='add-metric-label'>Select a Metric</InputLabel> */}
```
withinboredom, line 260: This can be deleted.

components/experiment-creation/Metrics.tsx
jessie-ross: Weird, I have no clue why these are here...

components/experiment-creation/Metrics.tsx
jessie-ross: Ah interesting, using the object itself as the "id".

I don't see any particular advantages to doing so and it seems to me a bit less direct and less idiomatic.

And for a disadvantage: What if someone in the future immutably changes a metricAssignment before calling `makeMetricAssignmentPrimary`

More generally, in JS I would prefer to work only at the "value" level.
- Using an object like this assigns an identity to a value that is seperate from the value, the identity should be within the value so that values are comparable. Now you would have two kinds of identity - object and value, why not have one.
- It is hidden so it can't be saved, pushed to the server, or logged.

Exceptions to this might be for performance sensitive code.

components/experiment-creation/Metrics.tsx
jessie-ross: I've fixed the typing.

> someone may modify the html in their browser to create a custom attribution window

That user should know they are own their own.

(And we do pretty extensive validation on the client before submit):
https://github.com/Automattic/abacus/blob/416ff1ce77e56d6cbe63ae8c81e97665574b3a5e/lib/schemas.ts#L94-L108

PR #236, opened 2020-07-27
WIP: Normalized Data

### LINKED ISSUE #201: Apply consistent object normalisation across the codebase
Mapping nested objects (e.g., metrics to an experiment's metric assignments) is currently handled inconsistently throughout the code. @jessie-ross suggested we use [normalizr](https://github.com/paularmstrong/normalizr) for this.

See: https://github.com/Automattic/abacus/pull/198#discussion_r449975708

jessie-ross:
_For an example of normalisation [see the `normalizr` docs](https://github.com/paularmstrong/normalizr#quick-start)_

**We use denormalised data up top and normalise below, this PR flips that around.**
- Uses [`normalizr`](https://github.com/paularmstrong/normalizr) a solid, minimal and reliable lib.
- Uses schemas to index the metrics/segments data for experiment creation.
   Not exactly necessary, we could be just using an `index<T>(data: T, id) -> Record<number, T>` function for simplicity but I like the centralisation.
- Populates the Segments in experiment creation
- Adds a normalizr schema and types for ExperimentFull
- Shows the data-loading and normalisation of experiment full

Not too much to go on this:
- Pass the normalised data down for experiments, replacing indexing code
- Clean up some of the code using denormalised data
- Remove the references to denormalised `experiment`
- Fix tests

Fixes #201

---

**Rationale:** 
- Some of our code is doing cartwheels to get normalised data, ends up being tied to specific parts when it is a global concern.
- Having data readily available in a normalised shape allows us to use normalised data without a second thought. 
   We have no consistent way of doing that currently.
- It is easier (and cheaper) to denormalise data than to normalise it, so normalised is better to start off with.
- If we do change to a global state management system (like Redux) we would likely be working with normalised data.

**Design decisions:**
- Normalize as needed using an opt-in approach, *but* use standard normalisation shapes (`normalizr` schema and normalized types). 
- If entities come from a normalisation we keep those together when passing them around.
   - It is a snapshot of a single entity, we don't want to pollute it (merge with other data) or lose any parts (split it).
- Be explicit about whether the data is normalized.

**This PR introduces some new naming conventions:**
   - `indexedX` - for indexed data, eg. `indexedMetrics`
   - `normalizedX` - for a single normalized entity without children, eg. `normalizedExperiment`
   - `normalizedXData` - for a complete normalized entity 
   - See also the typescript name conventions, may change those yet to be more consistent with the above
 
 

## How has this been tested?

- It hasn't yet.

### DIFF
--- a/components/AudiencePanel.test.tsx
+++ b/components/AudiencePanel.test.tsx
@@ -1,6 +1,14 @@
+import { normalize } from 'normalizr'
 import React from 'react'
 
 import RenderErrorBoundary from '@/components/RenderErrorBoundary'
+import {
+  ExperimentFull,
+  ExperimentFullNormalizedEntities,
+  experimentFullNormalizrSchema,
+  Segment,
+  segmentNormalizrSchema,
+} from '@/lib/schemas'
 import Fixtures from '@/test-helpers/fixtures'
 import { render } from '@/test-helpers/test-utils'
 
@@ -8,7 +16,13 @@ import AudiencePanel from './AudiencePanel'
 
 test('renders as expected with no segment assignments', () => {
   const experiment = Fixtures.createExperimentFull()
-  const { container } = render(<AudiencePanel experiment={experiment} segments={[]} />)
+  const normalizedExperimentData = normalize<ExperimentFull, ExperimentFullNormalizedEntities>(
+    experiment,
+    experimentFullNormalizrSchema,
+  )
+  const { container } = render(
+    <AudiencePanel normalizedExperimentData={normalizedExperimentData} indexedSegments={{}} />,
+  )
 
   expect(container).toMatchSnapshot()
 })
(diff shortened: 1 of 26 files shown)


yanirs:
I'll hold off on more specific feedback until this is non-WIP, but I like where this is going! :slightly_smiling_face:

jessie-ross:
I am going to split this PR into three to make it a bit easier, so hold your horses.

jessie-ross:
Actually I might get you to take a look to see how you feel it is going @yanirs 

## Update

I now have normalised data flowing down through `/experiment/[id]`

This looks like a huge change with 26 files touched so I might break it down a bit:
- 10 of which are tests where I am just normalising the data before input
- Populating Metrics and Segments in the experiment creation form.
- Most of the rest are just replacements of `experiment` for `normalisedExperiment`
- Refactored a bit more substantially:
   - `getDeployedVariation`
   - `AudiencePanel`
   - `MetricAssignmentPanel`

What's left:
- Using normalised data in `ExperimentResults`
- Fixing the storybooks

I may split it into these three:
- Populating the Metrics and Segments in experiment creation form
- Using normalised data in `ExperimentDetails`
- Using normalised data in `ExperimentResults`

yanirs:
Thanks @jessie-ross! Splitting into three PRs to make this more manageable sounds good. A couple of thoughts from skimming the current PR:

* Normalisation calls in the tests can get quite long and repetitive. Maybe we can introduce normalised fixtures instead?
* Even outside the tests, it'd be nice to reduce duplication and improve clarity. I may need to play with it in the context of one of the three PRs to see if it makes sense, but one idea that comes to mind is moving the normalisation schemas to their own file and adding some helpers. For example, a block like:

    ```typescript
    const metrics = await MetricsApi.findAll()
    const {
      entities: { metrics: indexedMetrics },
    } = normalize<MetricBare, { metrics: Record<number, MetricBare> }>(metrics, [metricBareNormalizrSchema])
    return indexedMetrics
    ```
    may become `return Normalizr.indexMetrics(await MetricsApi.findAll())` where `Normalizr` (or a different name) is our namespace for `normalizr` helpers.

jessie-ross:
Waiting on #258 to continue with this

PR #238, opened 2020-07-28
Experiment Creation Form: Polish the Scrolling

jessie-ross:
**This PR makes the scrolling horizontal:**
- Allows a better experience for scrolling the long forms as you don't need an internal scroll bar
- Works well with the prev-next buttons

Demo Vid: https://d.pr/v/DeN4hm

In doing so I have had to dynamically set the width on the form-parts.

I have used the npm library `@rehooks/component-size`:
- 2+ years old.
- Micro library, does only one thing.
- Well maintained.
- Very easily replaceable. 

**Responsive considerations:**

For small screen sizes this will all get disabled and we will just have a long scrolling form with no stepper.

**Should ExperimentForm be refactored?**

ExperimentForm is getting quite big.

The component currently has three tasks:
1. Manages the steps and the stepper.
2. Initialises the form.
3. Manages the scrolling of the form.

These tasks are interconnected:
- `1 <-> 2` Syncing the validation and completion state with the steps.
- `1 <-> 3` Syncing the scroll with the step

Right now my feeling is that I want to see them change a bit further until separating them out, and it is likely that `3` will be separated out first.

## How has this been tested?

- More tests for Experiment Creation Form coming once it is stable.

### DIFF
--- a/components/experiment-creation/ExperimentForm.tsx
+++ b/components/experiment-creation/ExperimentForm.tsx
@@ -3,8 +3,10 @@
 /* istanbul ignore file */
 import { Button, Paper, Step, StepButton, Stepper, Typography } from '@material-ui/core'
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
+import useComponentSize from '@rehooks/component-size'
 import { Formik } from 'formik'
-import React, { useRef, useState } from 'react'
+import _ from 'lodash'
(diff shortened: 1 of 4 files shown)


yanirs:
> Should ExperimentForm be refactored?
> ...
> Right now my feeling is that I want to see them change a bit further until separating them out, and it is likely that 3 will be separated out first.

Sounds good. :slightly_smiling_face:

aaronyan:
Part of a series on #9

PR #240, opened 2020-07-30
Add normalisation to populate creation form

jessie-ross:
_Split from #236, part 1/3. See the original PR for design discussion._

This PR:
- Adds `normalizer`
- Uses schemas to index the metrics/segments data for experiment creation.
- Populates the Segments in experiment creation
- Adds `/lib/normalizr` to hold normalisation functions.

**Should we have normalised fixtures?**
I don't particularly like the idea of normalised fixtures and I would want to go the opposite direction if possible: have fixtures of the server data rather than our internal format. 

With the added refactoring since the original PR I think it is quite light to normalise data in the tests.

## How has this been tested?

### DIFF
--- a/components/experiment-creation/ExperimentForm.test.tsx
+++ b/components/experiment-creation/ExperimentForm.test.tsx
@@ -4,6 +4,7 @@ import MockDate from 'mockdate'
 import React from 'react'
 
 import { createNewExperiment } from '@/lib/experiments'
+import * as Normalizr from '@/lib/normalizr'
 import Fixtures from '@/test-helpers/fixtures'
 
 import ExperimentForm from './ExperimentForm'
@@ -12,8 +13,8 @@ test('renders as expected', () => {
   MockDate.set('2020-07-21')
   const { container } = render(
     <ExperimentForm
-      metrics={Fixtures.createMetricBares(20)}
-      segments={Fixtures.createSegments(20)}
+      indexedMetrics={Normalizr.indexMetrics(Fixtures.createMetricBares(20))}
+      indexedSegments={Normalizr.indexSegments(Fixtures.createSegments(20))}
       initialExperiment={createNewExperiment()}
     />,
   )
--- a/components/experiment-creation/ExperimentForm.tsx
+++ b/components/experiment-creation/ExperimentForm.tsx
@@ -86,12 +86,12 @@ const useStyles = makeStyles((theme: Theme) =>
 )
 
 const ExperimentForm = ({
-  metrics,
-  segments,
+  normalizedMetrics,
+  normalizedSegments,
   initialExperiment,
 }: {
-  metrics: MetricBare[]
-  segments: Segment[]
+  normalizedMetrics: Record<string, MetricBare>
+  normalizedSegments: Record<string, Segment>
   initialExperiment: Partial<ExperimentFullNew>
 }) => {
   const classes = useStyles()
--- a/components/experiment-creation/ExperimentForm.tsx
+++ b/components/experiment-creation/ExperimentForm.tsx
@@ -190,7 +190,7 @@ const ExperimentForm = ({
               </div>
               <div className={classes.formPart} ref={formPartAudienceRef}>
                 <Paper className={classes.paper}>
-                  <Audience formikProps={formikProps} />
+                  <Audience formikProps={formikProps} normalizedSegments={normalizedSegments} />
                 </Paper>
                 <div className={classes.formPartActions}>
                   <Button onClick={prevStage}>Previous</Button>
--- a/components/experiment-creation/ExperimentForm.tsx
+++ b/components/experiment-creation/ExperimentForm.tsx
@@ -86,12 +86,12 @@ const useStyles = makeStyles((theme: Theme) =>
 )
 
 const ExperimentForm = ({
-  normalizedMetrics,
-  normalizedSegments,
+  indexedMetrics,
+  indexedSegments,
   initialExperiment,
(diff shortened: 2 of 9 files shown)


components/experiment-creation/ExperimentForm.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 
 import { createNewExperiment } from '@/lib/experiments'
+import * as Normalize from '@/lib/normalize'
```
yanirs, line 4: For consistency, I think we should make this `import * as Normalizers from '@/lib/normalizers'`:

* Using a verb for a namespace looks off &ndash; verb phrases are commonly used for actions (functions).
* All the files under `lib/` are plural noun phrases.

lib/schemas.ts
```diff
@@ -59,6 +60,11 @@ export const metricBareSchema = yup
   .defined()
   .camelCase()
 export type MetricBare = yup.InferType<typeof metricBareSchema>
+export const metricBareNormalizrSchema = new normalizr.schema.Entity<MetricBare>(
```
yanirs, line 63: I think these schemas should live in `normalizers.ts` and not be exported.

pages/experiments/new.tsx
```diff
@@ -17,29 +18,33 @@ const ExperimentsNewPage = function () {
...
   const initialExperiment = createNewExperiment()
 
-  const { isLoading: metricsIsLoading, data: metrics, error: metricsError } = useDataSource(
-    () => MetricsApi.findAll(),
-    [],
-  )
+  const { isLoading: metricsIsLoading, data: indexedMetrics, error: metricsError } = useDataSource(async () => {
+    const metrics = await MetricsApi.findAll()
+    return Normalize.indexMetrics(metrics)
+  }, [])
```
yanirs, line 24: I think it'd be cleaner without the temporary variable, which would keep this as a one-liner (same for segments).

```suggestion
  const { isLoading: metricsIsLoading, data: indexedMetrics, error: metricsError } = useDataSource(
    async () => Normalizers.indexMetrics(await MetricsApi.findAll()),
    [],
  )
```

lib/normalize.ts
```diff
@@ -0,0 +1,25 @@
+import { normalize } from 'normalizr'
+
+import { MetricBare, metricBareNormalizrSchema, MetricFull, Segment, segmentNormalizrSchema } from './schemas'
+
+export function indexMetrics<Metric extends MetricBare | MetricFull>(metrics: Metric[]) {
```
yanirs, line 5: Can you please add docstrings and declare the return type? It's not obvious in these cases.

lib/normalize.ts
```diff
@@ -0,0 +1,25 @@
...
+  const {
+    entities: { metrics: indexedMetrics },
+  } = normalize<Metric>(metrics, [metricBareNormalizrSchema])
+  /* istanbul ignore next */
```
yanirs, line 9: I think it'd be better to unit test these functions.

components/experiment-creation/ExperimentForm.tsx
```diff
@@ -86,12 +86,12 @@ const useStyles = makeStyles((theme: Theme) =>
...
 const ExperimentForm = ({
-  metrics,
-  segments,
+  indexedMetrics,
+  indexedSegments,
   initialExperiment,
 }: {
-  metrics: MetricBare[]
-  segments: Segment[]
+  indexedMetrics: Record<string, MetricBare>
```
yanirs, line 93: Why is this `Record<string, MetricBare>` rather than `Record<number, MetricBare>` (same for segment)?

components/experiment-creation/ExperimentForm.test.tsx
```diff
@@ -12,8 +13,8 @@ test('renders as expected', () => {
   MockDate.set('2020-07-21')
   const { container } = render(
     <ExperimentForm
-      metrics={Fixtures.createMetricBares(20)}
-      segments={Fixtures.createSegments(20)}
+      indexedMetrics={Normalize.indexMetrics(Fixtures.createMetricBares(20))}
```
yanirs, line 16: I agree that with the helpers we don't need normalised fixtures. :slightly_smiling_face:

components/experiment-creation/ExperimentForm.stories.tsx
jessie-ross: Nice!

lib/normalize.ts
jessie-ross: I have added an explicit typescript return type, is that ok?

yanirs: Yeah, that's what I meant. I still think it's worth adding a short description as well, as it won't necessarily be obvious what this function does from the name in a few months. :slightly_smiling_face:

yanirs:
@jessie-ross I just pushed https://github.com/Automattic/abacus/pull/240/commits/acb808344cbead482ad94531b9d08777b3d0bab2 since it was quicker to play with it than to describe the changes. :slightly_smiling_face: Summary:

* Remove the warning supressions from the tests by using fixture objects.
* Document and simplify the indexing functions.
* Not export the normalizr schemas (can start exporting them if we do need them later).
* Remove unused normalizr schema from `schemas.ts`.
* Use one-liner callbacks in `new.tsx`.

Should be good to go if you don't see any issues with the changes.

jessie-ross:
Thanks! Looks good!

aaronyan:
Part of a series on #9

PR #245, opened 2020-07-30
Add experiment creation stage-level validation

### LINKED ISSUE #242: Remove scroll code from Experiment Form
See #241

We have disabled the scrolling so we should remove the code related to it.

jessie-ross:
**This PR adds stage level validation to the form:**
- Validate's the last stage after moving away
  - Shows either completion or error state
- Validates all stages when you get to the submit stage
- Disables the submit button until submit
- Fixes Validation

And unrelatedly:
- Changes "3 Days" attribution window to "72 hours"
- Removes the scrolling code (To make this all simpler)

Demo Vid: https://d.pr/v/2gwkGn
- There is an unadressed bug with the metricAssignment table width.
- Since the vid I have added always showing current step

Part of a series on #9 
Fixes #242 

## How has this been tested?

- Manual Testing on local data
- Basic Automated testing
- More extensive e2e and unit testing of ExperimentForm to come in a seperate PR


components/experiment-creation/ExperimentForm.tsx
```diff
@@ -97,179 +110,204 @@ const ExperimentForm = ({
...
+
   useEffect(() => {
-    const onResize = _.debounce(() => {
-      changeStage(currentStageId)
-    }, 100)
+    rootRef.current?.scrollIntoView({ behavior: 'smooth', block: 'start', inline: 'start' })
+  }, [currentStageId])
 
-    window.addEventListener('resize', onResize)
+  return (
```
jessie-ross, line 150: Best to hide whitespace for this one

yanirs: It actually gets more confusing. :disappointed:

components/experiment-creation/ExperimentForm.tsx
```diff
@@ -97,179 +110,204 @@ const ExperimentForm = ({
...
+          changeStage(stages[prevStageIndex].id)
+        }
+        const nextStage = () => {
+          if (stages.length <= currentStageIndex) {
+            return
+          }
+
+          const nextStageIndex = currentStageIndex + 1
+          changeStage(stages[nextStageIndex].id)
+        }
```
yanirs, line 200: Looks like some of these functions can be simplified, e.g., 

```suggestion
        const prevStage = () => {
          if (currentStageIndex > 0) {
            changeStage(stages[currentStageIndex - 1].id)
          }
        }
        const nextStage = () => {
          if (currentStageIndex < stages.length) {
            changeStage(stages[currentStageIndex + 1].id)
          }
        }
```

There's also a bug in the current `nextStage()` because if `currentStageIndex` is `stages.length - 1`, it tries to increment to `stages.length`.

jessie-ross: Nice catching that bug!

**I am happy to go through with these changes but I do code like this purposefully:**

1. Using short-circuit return instead of a wrapped if.

    There is a subtle but important difference between these two - the way they return:
    - A short-circuit explicitly returns under explicit conditions.
    - The wrapped if implicitly returns under implicit conditions.

    I believe short-circuit returns greatly improve readability - both by making control-flow more linear, by being a pattern that is familiar. I also think it adds an important layer of semantics (We skip this function if X vs we run this code if X). I believe it also improves later code changes - what if things get expanded, what if the if statement gets changed.

    I tend to use this pattern a lot, it is more generally known as the 'return early, return often' pattern, it is not always the best choice but I think it should be the default.
    Plenty of discussion around the web about it.
    And you will see it in my Auth code refactoring soon to be split into a seperate PR.

2. Using intermediate variables

     I do like to seperate out even small operations, I think it helps readability.

yanirs: I think it really depends on the code in question. In my opinion, here both practices reduce readability:

* The short-circuit return is often useful for long functions. But here I find that it adds more mental work, e.g., reading the body of the original `prevStage()` after the `if`, I need to remember that I'm now in the `currentStageIndex !== 0` case, i.e., invert the condition in my mind.
* Single-use intermediate variables are useful when the evaluated expression is complicated. However, each new variable introduces another thing to keep in mind. Taking `prevStage()` as an example again, having to remember what `prevStageIndex` is for one line can be seen as reducing readability.

With the proposed changes, I think that it's much clearer that the functions are called for their side-effects (no `return`s) and that the `currentStageIndex` is checked correctly (which wasn't the case previously).

>  Plenty of discussion around the web about it.

Yeah, plenty of opinions either way. Here a couple from Spartan programming on [vertical complexity](https://github.com/SpartanRefactoring/Main/wiki/Vertical-complexity) and [variable elimination](https://github.com/SpartanRefactoring/Main/wiki/Variable-elimination). I think that the best answer is: It depends. :slightly_smiling_face:

lib/experiments.ts
```diff
@@ -45,8 +45,15 @@ export function getDefaultAnalysisStrategy(experiment: ExperimentFull) {
   return experiment.exposureEvents ? AnalysisStrategy.PpNaive : AnalysisStrategy.MittNoSpammersNoCrossovers
 }
 
-export function createNewExperiment(): Partial<ExperimentFullNew> {
+export function createNewExperiment() {
```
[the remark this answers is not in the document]
jessie-ross: > Why is this change even needed?

There are a few changes here:
- Now have empty strings instead of `undefined`, this fixes any 'uncontrolled input becoming controlled' warnings.
- Empty strings for datetimes, this is a quirk of our date input controls wanting text.
- Strings for booleans, a quirk of our radio controls.

The last two mean that this deviates from the `ExperimentFullNew` type.

In the future it will probably be worth bringing back the original `createNewExperiment(): Partial<ExperimentFullNew>` function and bringing in a new type that is needed for the Experiment form and a function for converting to it. Likely when I get to experiment editing.

> Given the new return type, I don't think this belongs here anymore. It's not really an experiment object – it's a mapping to a bunch of strings.

I do still think this is an alright place for it to be. 

Firstly it needs to be somewhere a bit more common than a component - it is needed for a lot of testing code, that leaves the `util` and the `lib` code, I think it fits more in with the `lib` code. I didn't think it warranted a new file, and it fits closest with `/lib/experiments.ts`. 

Secondly `experiments.ts` isn't experiment-object-centric, rather it is code relating to experiment in general, what we have here is in fact the third experiment type in `/lib/experiments.ts`.

Thirdly, as we are using lib files of related functions, it is incredibly easy to move them around as different patterns emerge. Even if it is not ideal now, later down the track we can find a home for it.

jessie-ross:
## Update

I have rebased this PR on top of the new PR #250, which has the "Scroll code removal" split from this PR.
Once that PR is merged this PR will be simpler.

I have also changed the prev/next functions as per your recommendations.

lib/experiments.ts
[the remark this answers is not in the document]
yanirs: OK, these are good reasons to keep it here _for now_, but I find the name and lack of documentation confusing. This function doesn't return an experiment, it returns an object that maps some experiment fields to strings. It's very specific to form quirks. I think it should be documented and have a name that matches its functionality.

components/experiment-creation/ExperimentForm.tsx
```diff
@@ -94,155 +111,198 @@ const ExperimentForm = ({
...
 }) => {
   const classes = useStyles()
 
   const rootRef = useRef<HTMLDivElement>(null)
 
   const [currentStageId, setActiveStageId] = useState<StageId>(StageId.Beginning)
   const currentStageIndex = stages.findIndex((stage) => stage.id === currentStageId)
+
   const [completeStages, setCompleteStages] = useState<StageId[]>([])
   const markStageComplete = (stageId: StageId) => {
```
yanirs, line 124: This gets a bit repetitive and hard to follow. I'll push a simplification proposal in a separate PR. :slightly_smiling_face:

yanirs: See #251

lib/experiments.test.ts
```diff
@@ -68,15 +68,32 @@ describe('lib/experiments.ts module', () => {
...
-      expect(Experiments.createNewExperiment()).toEqual({
-        platform: Platform.Wpcom,
-        metricAssignments: [],
-        segmentAssignments: [],
-        variations: [
-          { name: 'control', isDefault: true, allocatedPercentage: 50 },
-          { name: 'treatment', isDefault: false, allocatedPercentage: 50 },
-        ],
-      })
+      expect(Experiments.createNewExperiment()).toMatchInlineSnapshot(`
```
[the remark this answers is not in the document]
jessie-ross: Just out of efficiency, so we don't have to update it by hand any time it changes.

PR #246, opened 2020-07-30
Add Experiment Creation submit

jessie-ross:
**This PR covers submitting:**
- Sending data to our API
- Submit progress
- Submit success/failure notifications
- Navigate to `/experiment/[id]/snippets?freshly_created` after success
   (Can change later.)

Demo Vid: https://d.pr/v/71v7q9

Part of a series on #9

## How has this been tested?

- Manually tested
- Existing automated tests
- Full e2e tests for ExperimentForm to come
- I want to test this on production if possible before I go further with it.


lib/schemas.ts
```diff
@@ -214,6 +214,7 @@ export type ExperimentFull = yup.InferType<typeof experimentFullSchema>
 const now = new Date()
 export const experimentFullNewSchema = experimentFullSchema.shape({
   experimentId: idSchema.nullable(),
+  status: yup.string().oneOf(Object.values(Status)),
```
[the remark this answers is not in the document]
jessie-ross: So status isn't being sent, this is just because I extend experimentFullSchema and I wanted to make it not required.

It might be better if I change to specifically being undefined.

jessie-ross:
I have it tested with my local setup with the prod APIs, it just needed some conversion to snake_case.

It should be ready to go.

api/ExperimentsApi.ts
```diff
@@ -18,7 +19,8 @@ import { fetchApi } from './utils'
  */
 async function create(newExperiment: ExperimentFullNew) {
   const validatedNewExperiment = await experimentFullNewSchema.validate(newExperiment, { abortEarly: false })
-  const experiment = await fetchApi('POST', '/experiments', validatedNewExperiment)
+  const outboundNewExperiment = experimentFullNewOutboundSchema.cast(validatedNewExperiment)
+  const experiment = await fetchApi('POST', '/experiments', outboundNewExperiment)
```
yanirs, line 23: Can call this `returnedExperiment` or something similar to differentiate it from the two other variables.

lib/schemas.ts
```diff
@@ -243,6 +246,24 @@ export const experimentFullNewSchema = experimentFullSchema.shape({
   variations: yup.array<VariationNew>(variationNewSchema).defined().min(2),
 })
 export type ExperimentFullNew = yup.InferType<typeof experimentFullNewSchema>
+export const experimentFullNewOutboundSchema = experimentFullNewSchema
+  .shape({
+    metricAssignments: yup.array(metricAssignmentNewOutboundSchema).defined(),
```
yanirs, line 251: Should this have `.min(1)`? We generally also want to ensure that there's a primary metric, as not having one will stop the experiment from launching.

lib/schemas.ts
```diff
@@ -214,7 +217,7 @@ export type ExperimentFull = yup.InferType<typeof experimentFullSchema>
 const now = new Date()
 export const experimentFullNewSchema = experimentFullSchema.shape({
   experimentId: idSchema.nullable(),
-  // Using yesterday here to avoid timezone issues
+  status: yup.mixed().oneOf([]).notRequired(),
```
yanirs, line 220: Is there a more explicit way of expressing the status being `undefined`? Maybe add an inline comment if not?

jessie-ross: Yeah was the best I could find, I'll add a comment

lib/schemas.ts
jessie-ross: The outbound schema is just for casting (making the data snake_case) not for verification, I'll add a comment to that effect.

jessie-ross: > Should this have .min(1)? 

We do currently check this.

PR #249, opened 2020-07-31
Add config and refactor auth code

### LINKED ISSUE #228: Make it easy to run locally with the production API
As noted by @scjr in https://github.com/Automattic/abacus/pull/227#pullrequestreview-453020865, it'd be good to add a switch that allows us to run locally against the production API, probably with a strong warning that this is what we're doing:

> If we are frequently testing on Production servers it might also be a good idea to add on a switch that makes it easy for us to do so. It can be a double edged sword especially if we are editing/creating data as mistakes can be made, it would be good if it was super obvious that we are using production endpoints such as with a red banner.

jessie-ross:
**This PR refactors the code so we get a nice config object out of it:**
- Move config variables to `/config.ts`
- Add switch to change to Production config in development mode 
- Add warning banner for working with production.

While I was there I refactored the auth code:
- Centralise auth code in `/utils/auth.ts`
- Better error checking (It would miss anything other than `access_denied` before.)
- Styled the Auth page

Fixes #228 

I've seen some discussion about e2e testing auth flow with pbmo2S-kn-p2, it looks like it was never in use.
In the future we should add a third profile to config for testing auth.

## How has this been tested?

- Manual testing with mock server
- Manual testing with production server

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -20,4 +20,5 @@ export const isTestingProductionConfigInDevelopment = false
 
 const isProduction = false
 
+// istanbul ignore next; Development only
 export const config = isProduction || isTestingProductionConfigInDevelopment ? productionConfig : developmentConfig
--- a/config.ts
+++ b/config.ts
@@ -1,0 +1,23 @@
+const productionConfig = {
+  experimentApi: {
+    needsAuth: true,
+    authPath: 'https://public-api.wordpress.com/oauth2/authorize',
+    authClientId: 68795,
+    rootUrl: 'https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0',
+  },
+}
+
+const developmentConfig = {
+  experimentApi: {
+    needsAuth: false,
+    authPath: null,
+    authClientId: 68797,
+    rootUrl: 'https://virtserver.swaggerhub.com/yanir/experiments/0.1.0',
+  },
+}
+
+export const isTestingProductionConfigInDevelopment = false
+
+const isProduction = false
+
+export const config = isProduction || isTestingProductionConfigInDevelopment ? productionConfig : developmentConfig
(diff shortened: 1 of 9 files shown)


config.ts
```diff
@@ -0,0 +1,24 @@
...
+    needsAuth: false,
+    authPath: null,
+    authClientId: 68797,
+    rootUrl: 'https://virtserver.swaggerhub.com/yanir/experiments/0.1.0',
+  },
+}
+
+export const isTestingProductionConfigInDevelopment = false
+
+const isProduction = false
```
jessie-ross, line 21: TODO: This needs to work with our deployment system.

config.ts
```diff
@@ -0,0 +1,24 @@
...
+const developmentConfig = {
+  experimentApi: {
+    needsAuth: false,
+    authPath: null,
+    authClientId: 68797,
+    rootUrl: 'https://virtserver.swaggerhub.com/yanir/experiments/0.1.0',
+  },
+}
+
+export const isTestingProductionConfigInDevelopment = false
```
yanirs, line 19: I'm not sure if this is ready, but how is this supposed to be used?

jessie-ross: This is ready, you just set this value to true to use the production config.

yanirs: Can we get this from the CLI?

jessie-ross: We can but I am not quite sure what the advantages of that are?

This allows you to change it on the fly as you are developing instead of having to start and stop the dev server.

yanirs: The main advantage is not needing to change a file that doesn't get committed, which is the main issue with the current solution.

jessie-ross:
This PR has been split into #252 and #253

PR #251, opened 2020-08-04
Suggestion for #245: Simplify stage state tracking

yanirs:
Suggestion following https://github.com/Automattic/abacus/pull/245#discussion_r464856594: Use self-documenting lodash calls instead of single-use functions.

## How has this been tested?


components/experiment-creation/ExperimentForm.tsx
```diff
@@ -121,28 +121,7 @@ const ExperimentForm = ({
   const currentStageIndex = stages.findIndex((stage) => stage.id === currentStageId)
 
   const [completeStages, setCompleteStages] = useState<StageId[]>([])
```
yanirs, line 123: Another idea: we can replace these arrays with a mapping from each stage to its validations status (undefined/false/true), but it can get a bit messy.

PR #252, opened 2020-08-04
Add config profiles

jessie-ross:
_Split from #249_

This PR adds config and config profiles:
- Consolidates static configuration throughout the app into one file
- Allows having different profiles
- Add switch to change to Production config in development mode, add environment variable to set the switch 
- Add warning banner for working with production.
- Ensures development profile and `NODE_ENV=test` for e2e testing

## How has this been tested?

   Try out prod-config-in-development:
   1. `NEXT_PUBLIC_PRODUCTION_CONFIG_IN_DEVELOPMENT=true npm run dev`
   2. Go into the dev tools and delete our auth token in local storage - you should only have to do this once.


jessie-ross:
The smoke test is failing but I feel that it is because of the redirect race condition, I don't see why it is failing here but not on #253

yanirs:
> The smoke test is failing but I feel that it is because of the redirect race condition, I don't see why it is failing here but not on #253

Looks like it's not a race condition, as it's failing consistently. I also can't run it locally: When I clear the local storage and try to go to http://a8c-abacus-local:3000, it gets into a redirect loop.

config.ts
```diff
@@ -0,0 +1,44 @@
+const productionConfig = {
+  experimentApi: {
+    needsAuth: true,
+    authPath: 'https://public-api.wordpress.com/oauth2/authorize',
+    authClientId: 68795,
+    rootUrl: 'https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0',
+  },
+}
+
+const developmentConfig = {
```
yanirs, line 10: Maybe call this `mockConfig`? We don't only use it in the dev environment.

config.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Simply set `NEXT_PUBLIC_PRODUCTION_CONFIG_IN_DEVELOPMENT=true` before spinning up the app.
+ *
+ * You may need to clear your localstorage to get the app to reauth.
+ */
+export const isTestingProductionConfigInDevelopment =
+  // istanbul ignore next; Development only
+  process.env.NEXT_PUBLIC_PRODUCTION_CONFIG_IN_DEVELOPMENT === 'true'
+
+export /* istanbul ignore next; Development only */ const config =
```
yanirs, line 43: Why are the `istanbul ignore` comments different in each case? I find the format on line 29 to be the cleanest.

pages/_app.tsx
```diff
@@ -93,29 +95,17 @@ const App = React.memo(function App(props: AppProps) {
   }, [])
 
   if (typeof window !== 'undefined') {
-    // Inject a fake auth token to skip authentication in non-production contexts.
-    // This is a temporary solution. Ideally, we should test the full authentication flow in every environment.
-    if (window.location.host !== 'experiments.a8c.com') {
```
yanirs, line 98: This is probably the cause of the test failures and the redirect loop. I haven't tested it, but I suspect that adding `&& config.experimentApi.needsAuth` to the condition on line 97 would fix it.

pages/_app.tsx
```diff
@@ -9,7 +9,9 @@ import * as yup from 'yup'
 import RenderErrorBoundary from '@/components/RenderErrorBoundary'
 import RenderErrorView from '@/components/RenderErrorView'
 import ThemeProvider from '@/styles/ThemeProvider'
-import { getAuthClientId, getExperimentsAuthInfo, saveExperimentsAuthInfo } from '@/utils/auth'
+import { getExperimentsAuthInfo } from '@/utils/auth'
+
+import { config } from '../config'
```
yanirs, line 14: Any reason not to import from `@/config`?

pages/_app.tsx
jessie-ross: Ahhh thanks, I see.

jessie-ross: Yeah I forgot I had fixed this during the auth refactor

config.ts
jessie-ross: Firstly the ignore on line 40 isn't needed, I had forgotten about it. The ones on line 30 and 43 are necessary.

I do prefer them like 29 as well but the one on line 43 only works when it is in that position. It is a weird bug in istanbul to do with exports. For consistency I can use the line 41 version throughout from now on.

jessie-ross:
It looks like there is a seperate problem with the redirect tests sometimes breaking, this must be what confused me.

Screen Shot 2020-08-05 at 9 45 38 am

yanirs:
@jessie-ross I'm trying to follow the instructions by running `NEXT_PUBLIC_PRODUCTION_CONFIG_IN_DEVELOPMENT=true npm run dev`. I cleared the local storage but it's still loading the mock server. Any ideas?

Also, can you please document the new flag in the readme?

jessie-ross:
Have you run `npm install`?

yanirs:
> Have you run `npm install`?

:man_facepalming: yep, that fixed it. :slightly_smiling_face:

jessie-ross:
> Also, can you please document the new flag in the readme?

I am going to do this in a seperate PR (right now) so I can clean up the README a bit.

yanirs:
> I am going to do this in a seperate PR (right now) so I can clean up the README a bit.

Sounds good, thanks! :slightly_smiling_face:

PR #253, opened 2020-08-04
Refactor Auth code

jessie-ross:
_Split from #249_

This PR refactors the auth code:
- Centralised the auth code in `/utils/auth.ts`
- Better error checking (It would miss anything other than access_denied before.)
- Styled the Auth page

## How has this been tested?


pages/auth.tsx
```diff
@@ -1,68 +1,70 @@
...
-          }
-          saveExperimentsAuthInfo(experimentsAuthInfo)
-
-          window.location.replace(window.location.origin)
-        } else if (authInfo.error === 'access_denied') {
-          setError('Please log into WordPress.com and authorize Abacus - Testing to have access.')
-
-          saveExperimentsAuthInfo(null)
-        }
+    const error = onExperimentAuthCallbackUrl()
```
jessie-ross, line 40: Here we use an old fashioned error enum to seperate out the display side-effects.
This has the advantage of making our `auth.ts` code extremely reusable (I am surprised a pre-existing solution for this doesn't exist already.)

utils/auth.ts
```diff
@@ -30,12 +35,91 @@ const getExperimentsAuthInfo = (): ExperimentsAuthInfo | null => {
...
+
+  if (authInfo.error) {
+    console.error('Authentication Error:', authInfo.error, authInfo.error_description)
+    saveExperimentsAuthInfo(null)
+    switch (authInfo.error) {
+      case 'access_denied': {
+        return AuthError.AccessDenied
+      }
+      default: {
+        return AuthError.UnknownError
```
jessie-ross, line 108: We didn't handle the default case before and any other errors would have slipped by.

utils/auth.ts
```diff
@@ -30,12 +35,91 @@ const getExperimentsAuthInfo = (): ExperimentsAuthInfo | null => {
...
+      }
+    }
+  }
+
+  if (authInfo.access_token && authInfo.scope === 'global' && authInfo.token_type === 'bearer') {
```
jessie-ross, line 113: Ideally I would be short-circuiting and returning an UnknownError if this didn't hold but I didn't here and that is fine for now.

jessie-ross:
I've been having with the redirect test here as well:
Screen Shot 2020-08-05 at 9 47 41 am

I am quite skeptical about the net benefits of the redirect test, can we remove it?

jessie-ross:
And it seems like the required reviews rule has been turned off?
Screen Shot 2020-08-05 at 9 18 20 am

yanirs:
> I am quite skeptical about the net benefits of the redirect test, can we remove it?

As you know, I'm not a fan of removing tests just because of testing issues. :slightly_smiling_face: Each removal in isolation may make sense at the time, but it can reduce the overall codebase health over time.

jessie-ross:
> As you know, I'm not a fan of removing tests just because of testing issues. 🙂 Each removal in isolation may make sense at the time, but it can reduce the overall codebase health over time.

I feel like this is a bit too idealistic a way to look at it. In practice I think we should be weighing the benefit vs the costs (considering the long-term) as we go. This is ultimately the "tests cost in the short run, save in the long run" argument which I think is true to some extent but it certainly should have exceptions.

Here is why I think it is not worth it in this case:
- We have to look for another solution for it, it is not simple to do, may take an hour or more.
- It is very unlikely to break.
- It is not an architectural concern (it is very isolated).
- If it breaks it is not a serious problem.
- If it breaks there is a workaround for the user `/experiments`.
- It is easy to know when it is broken.
- It is cheap to fix.
- It is going to be replaced in the future by the Dashboard.

yanirs:
> And it seems like the required reviews rule has been turned off?

Nope, this was because the base branch wasn't `master`. Now that the base branch changed, it's required again. :slightly_smiling_face: 

> Here is why I think it is not worth it in this case

These are good arguments for this case. Let's disable the test if it keeps causing trouble, as I still think it's worth knowing how to fix it in the general case.

yanirs:
> Nope, this was because the base branch wasn't `master`. Now that the base branch changed, it's required again. :slightly_smiling_face:

Speaking of which, the deletion of the base branch (which happened automatically on merge) messed up the diff view. Can you please fix it? Then I'll have another look.

utils/auth.ts
jessie-ross: Actually I may as well fix this

jessie-ross:
It is probably best to look at these four commits individually:

Screen Shot 2020-08-05 at 12 04 50 pm

They are pretty well isolated except for "Extract static config out" which actually adds better error checking to auth initialization.

I have been getting a bit slack on my commits (this PR in particular is a mess) but I usually curate my PRs through the commits and commit messages, GitHub has a nice feature to review PR's by commit (when you click "Start Review" I think). 

I try to make my changes atomic across commits and do one thing per commit, it helps if we are rebasing as well.

@yanirs I think you previously made an argument against rebasing but it seems like Github handles it quite well by keeping comments and marking them as outdated.

pages/auth.tsx
```diff
@@ -1,68 +1,70 @@
...
+      marginTop: theme.spacing(3),
+    },
+  }),
+)
+
 /**
  * The authorization page.
  *
+ * TODO: It is better to have this at /auth-callback
+ *
```
yanirs, line 30: Not sure if this helps as an inline TODO. We're better off opening an issue or just changing it, though I'm not sure I agree with calling the file `auth-callback` &ndash; it makes it sound more like a function than a page.

pages/_app.tsx
```diff
@@ -93,21 +91,7 @@ const App = React.memo(function App(props: AppProps) {
...
-        redirect_uri: `${window.location.origin}/auth`,
-        response_type: 'token',
-        scope: 'global',
-      }
-
-      const authUrl = `${config.experimentApi.authPath}?${qs.stringify(authQuery)}`
-      window.location.replace(authUrl)
-    }
-  }
+  initializeExperimentsAuth()
```
yanirs, line 94: This moves the ` window.location.replace(authUrl)` side effect from the page to a utility function, which I don't think is a good idea.

pages/auth.tsx
yanirs: I'm not sure why we need extreme reusability, and as above, I disagree with moving redirection side effects to utility functions.

This is also quite confusing since the `error` variable isn't passed to `setError()`, and it's not even an error in some cases. Returning `void | AuthError` from any function makes it hard to follow &ndash; a nullable error makes more sense to me since then we can check for `if (error) ... ` rather than use a switch.

utils/auth.ts
```diff
@@ -30,12 +35,94 @@ const getExperimentsAuthInfo = (): ExperimentsAuthInfo | null => {
...
+export const saveExperimentsAuthInfo = (experimentsAuthInfo: ExperimentsAuthInfo | null) => {
   if (experimentsAuthInfo === null) {
     localStorage.removeItem('experiments_auth_info')
   } else {
     localStorage.setItem('experiments_auth_info', JSON.stringify(experimentsAuthInfo))
   }
 }
 
-export { getExperimentsAuthInfo, saveExperimentsAuthInfo }
+export /* istanbul ignore next; TODO: e2e test authorization */ function initializeExperimentsAuth() {
```
yanirs, line 46: I find that this `export /* istanbul ignore next; TODO: e2e test authorization */` pattern greatly decreases readability. This can easily be avoided by keeping things as they were, with a single `export` call.

utils/auth.ts
yanirs: Good catch, but this could have easily been handled in `AuthPage` with greater readability. No need for a big refactor and shuffling things around, which now requires more back-and-forth to see what's going on.

utils/auth.ts
```diff
@@ -30,12 +35,94 @@ const getExperimentsAuthInfo = (): ExperimentsAuthInfo | null => {
...
+  // The hash should look something like the following upon success:
+  // #access_token=...&expires_in=#######&scope=global&site_id=0&token_type=bearer
+  // The hash should look something like the following upon failure:
+  // #error=access_denied&error_description=You+need+to+log+in+to+WordPress.com&state=
+  const authInfo = qs.parse(window.location.hash.substring(1))
+
+  if (authInfo.error) {
+    console.error('Authentication Error:', authInfo.error, authInfo.error_description)
+    saveExperimentsAuthInfo(null)
+    switch (authInfo.error) {
```
yanirs, line 103: Why do we need a switch for two cases?

jessie-ross:
> It's still hard to see what exactly changed and why because this PR combines moving code with changing it.

How did you go with those commits I mentioned? 
By going commit by commit you should be able to see the movement and the changed code separately.

> In my opinion, the move is unnecessary and potentially harmful because we now have redirection logic in undocumented utility functions.

Let's break this down:
1. Unnecessary
2. Potentially harmful because we now have redirection logic in undocumented utility functions

**On Being Necessary**

I think it is important to have the auth code in one place so it can all be looked at at once. This is speaking from general experience fixing hand-rolled OAuth libs as well as seeing in third-party libs like Auth0 as well as particular experience in this code-base when looking at the Token code - it was spread amongst three files and mixed with presentation.

The code in the `/auth.ts` file relates only to code in the `/auth.ts` file rather than the code it was situated in before, for the best context it should all be viewed together. 

This is also (for the most part) business code rather than presentation which I believe should be kept seperate from the React view tree.

**The Redirection**

Firstly I should say the way "redirection logic" is done isn't inherent to collecting the code into this one file.
Eg. we could split `initializeExperimentsAuth` into `needsAuth()` and `redirectForLogin()` and use it like so in `_App.tsx`:
```
if (needsAuth()) {
    redirectForLogin()
}
```

Another solution that I would prefer might be a simple name change to indicate the redirection, from `initializeExperimentsAuth` to `ensureAuthWithPossibleLoginRedirection`.

> undocumented utility functions.

Also I am not opposed to documenting them.

utils/auth.ts
jessie-ross: Yeah I am used to dealing with more error types but in this case we should only need to check access denied.

utils/auth.ts
jessie-ross: Yeah this isn't ideal and it greatly bugs me as well.

I think having exports at the end is bad for readability as well as you can't immediately see which functions are export and which are internal.

There are a few in-between solutions, and I might try upgrading babel/jest (as this is a bug), and will take a look later.

pages/auth.tsx
jessie-ross: Yeah I might open up an issue, very low priority though.

The page itself is a callback with parameters set, the AuthProvider is effectively remotely calling a "function" on the users browser.
I did find it a bit awkward for it to be called Auth since it is doing a very specific thing.

pages/auth.tsx
jessie-ross: > I'm not sure why we need extreme reusability, and as above, I disagree with moving redirection side effects to utility functions.

Reusability is just a bonus and I have mentioned the redirection elsewhere.

> This is also quite confusing since the error variable isn't passed to setError(), and it's not even an error in some cases. Returning void | AuthError from any function makes it hard to follow – a nullable error makes more sense to me since then we can check for if (error) ... rather than use a switch.

Yeah it makes sense to setError the AuthError, and have the function return `null | AuthError`.

pages/_app.tsx
jessie-ross: See my PR comment.

yanirs:
@jessie-ross nothing has really changed on this PR, including on things you agreed to change, so there's not much to re-review here. I still don't think that this PR is needed or an improvement for the same reasons from my previous comment. Also, looking at it commit-by-commit isn't helpful since moving the code is still contentious. In general, commit-by-commit reviews require much more work from reviewers &ndash; stacked PRs are the way to go. Finally, this isn't a good way to spend our time &ndash; we have the milestone prioritisation process for a reason and I highly doubt we would see this as a priority when weighed against all the other things we have to do. 

In my opinion, this PR _still_ obfuscates the auth code and makes it more brittle (especially as it hasn't changed since the last round). The packaging of three different changes (styling, improving error handling, and moving code) makes it hard to ship the useful parts of it that have nothing to do with the contentious bit. You can ask @aaronyan or @withinboredom if you want another opinion, but this would just add to the workload on a low priority refactor, which isn't helpful in moving us towards our near-term goals.

And to be clear, I agree that it's possible to refactor the auth logic so that it's not in the pages (where routing, redirection, etc. still occur by calling the new functions), but it's not just the redirection. There's a lot of reliance on global state in the new functions that can and should be refactored in a cleaner way (which would also make it testable). But again, discussing it isn't the best use of our time given our other priorities.

jessie-ross:
> @jessie-ross nothing has really changed on this PR, including on things you agreed to change, so there's not much to re-review here.

Sorry, I just meant for you to review my last comment before I went ahead with further changes to this. I will use the "Request Review" button more purposefully in the future.

---

Seems like there are a few seperate issues going on here, some are specific to this PR and some are more general about the way we deal with code. I think it is useful to work through these issues as far as coming to better understand each others values around code and development practices. I will focus mostly on the general since you seem to be opposed to this change.

---

> this isn't a good way to spend our time – we have the milestone prioritisation process for a reason and I highly doubt we would see this as a priority when weighed against all the other things we have to do.

There are two parts to this, namely should I have made these changes in the first place and should we continue it right now.

**The first place**

The change itself was very quick. The decision was quick and easy as I have implemented and seen this practice a lot in both my code, others and third-party libraries. The coding was very quick, I had just read through all the auth code to fix the related issue and I this was a response to problems I had seen in the code. 

I judged that it would be worth it to make this improvement to the code while I was there and decided to include it with some other code. At the time I didn't think it would raise any eyebrows so I went ahead with it.

This might have been a misjudgement on my part and I will be sure to adjust for the future. But in general I do like to improve the codebase where it seems efficient to do so and that was the main motivation for this.

**Right Now**

But now we are here, and I was thinking we already have had discussion about it, we have it in our heads and are thinking about it. I was seeing this as a sunk cost and at this point thought the value of going a little further with it to get it over the line would out-weigh the cost.

---

> looking at it commit-by-commit isn't helpful since moving the code is still contentious

I mentioned the commit-by-commit because you were requesting the ability to see the movement seperate from the change of code. For every movement to be put in a seperate PR to code change seems a bit much? Is there anyway to twist your leg on that?

> The packaging of three different changes (styling, improving error handling, and moving code) makes it hard to ship the useful parts of it that have nothing to do with the contentious bit

I am happy to split it up and I will try to be a bit more fine grained in the future but I was thinking while we were both here if it could be dealt with.

---

As far as the specifics I will just mention that this is a very common pattern for auth code, [even scorecard uses it](https://github.com/Automattic/scorecard-next/blob/master/scorecard/helpers/api.js) (we should look into the way they are doing it, it seems like we are missing out on some handy wpcom npm libs, if not copy and pasting their auth code) and I don't see any obvious negative impact on testing, they should be easier to test now as they are unit-testable, just need to mock window.

jessie-ross:
I will leave this PR for now and split it up as mentioned.

yanirs:
> Sorry, I just meant for you to review my last comment before I went ahead with further changes to this. I will use the "Request Review" button more purposefully in the future.

No worries. I was going to reply anyway, just took a while to get to it as I was digging my way out of the pile of post-AFK notifications. :slightly_smiling_face: 

> The first place
>
> The change itself was very quick. The decision was quick and easy as I have implemented and seen this practice a lot in both my code, others and third-party libraries. The coding was very quick, I had just read through all the auth code to fix the related issue and I this was a response to problems I had seen in the code.
>
> I judged that it would be worth it to make this improvement to the code while I was there and decided to include it with some other code. At the time I didn't think it would raise any eyebrows so I went ahead with it.
>
> This might have been a misjudgement on my part and I will be sure to adjust for the future. But in general I do like to improve the codebase where it seems efficient to do so and that was the main motivation for this.

I completely agree with trying to improve the codebase wherever possible. I also see a lot of places where we can improve things throughout the code, but I often try to hold myself back from refactoring things I don't like unless I can justify the immediate need. For example, switching to `yup` was a great change as it simplified a bunch of downstream work, so it was worthwhile even though we could have made it work without `yup`. I made my point more succinctly in https://github.com/Automattic/abacus/pull/260#issuecomment-673876759 &ndash; I think we should try to focus on our agreed priorities and refactoring that supports these priorities when during the three milestone week, and do the more discretionary work in unplanned weeks. And of course, sometimes refactoring and paying down tech debt can become a higher priority for our planned work, which was the case recently for the PHP backend.

> Right Now
>
> But now we are here, and I was thinking we already have had discussion about it, we have it in our heads and are thinking about it. I was seeing this as a sunk cost and at this point thought the value of going a little further with it to get it over the line would out-weigh the cost.

Yeah, the sunk cost is sort of what I was getting at when referring to priority inversion in https://github.com/Automattic/abacus/pull/260#issuecomment-673876759 &ndash; it's much easier to say "no" or "not right now" to an issue/idea than to a PR, as less work has been done on it.

> I mentioned the commit-by-commit because you were requesting the ability to see the movement seperate from the change of code. For every movement to be put in a seperate PR to code change seems a bit much? Is there anyway to twist your leg on that?

Yeah, I think in some cases it's reasonable. Here, the move itself was what I was questioning, so commit-by-commit wasn't helpful.

> I will leave this PR for now and split it up as mentioned.

Thank you. :slightly_smiling_face: Happy to revisit the refactoring bit at a later stage.

jessie-ross:
> Yeah, the sunk cost is sort of what I was getting at when referring to priority inversion in #260 (comment) – it's much easier to say "no" or "not right now" to an issue/idea than to a PR, as less work has been done on it.

Ah I see, this is quite a good point.

jessie-ross:
## Update

Ok I have moved the code in the two functions back to where they were, and I am pretty sure I have taken care of @yanirs other concerns.

This PR now contains the following changes:
- Better error checking (It would miss anything other than access_denied before.)
- Warnings for strange Auth states.
- Styled Auth page.

pages/auth.tsx
```diff
@@ -15,54 +43,71 @@ const debug = debugFactory('abacus:pages/auth.tsx')
...
+      console.warn('onExperimentAuthCallbackUrl: Could not find `window`')
+      return
+    }
 
-          window.location.replace(window.location.origin)
-        } else if (authInfo.error === 'access_denied') {
-          setError('Please log into WordPress.com and authorize Abacus - Testing to have access.')
+    if (!window.location.hash || window.location.hash.length === 0) {
+      console.error('Authentication Error:', 'Missing hash in auth callback url.')
+      setError(AuthError.UnknownError)
```
yanirs, line 58: String errors are more descriptive and easier to debug:

```suggestion
      setError('Authentication Error: Missing hash in auth callback url.')
      return
```

Also, we can't expect all our users to be comfortable with the console, so if we include all the information as a string, it'll be easier to debug based on screenshots.

jessie-ross:
## Update

- Reverted changes to `_app.tsx`.
- Stuck to string errors.

yanirs (approved):

PR #254, opened 2020-08-05
Clean up readme

jessie-ross:
This PR effectively rewrote the Readme. 

It was very wordy in some places, had references to scripts we don't use, had references to scripts that don't exist anymore, was missing scripts we do use, and missing mentions of options on scripts we do use. 

Changes:
- Reorganised the readme into different sections:
   - Getting Started - Greatly expanded but concise, put in a nice order.
   - Testing Tips/Notes.
   - Advanced Tools.
- Removed references to unused scripts.
- Removed references to scripts that don't exist.
- Added references to scripts we use.
- Added `--updateSnapshot` references.
- Makes mentions of scripts much more concise.
- Makes Auth Flow section more concise.
- Adds section on Testing Production data 
- Mentions that Docker and Bundle Analysis are not recommended

## How has this been tested?
- It hasn't

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ npm run build
 
 - Integration: `/__tests__`
 - E2E: `/e2e`
-- Unit tests: Rest of the codebase alongside in the same directory of the file it is testing.
+- Unit tests: Rest of the codebase, alongside file it is testing.
 
 ### E2E Testing tips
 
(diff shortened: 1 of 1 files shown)


jessie-ross:
Best to [view like this](https://github.com/Automattic/abacus/blob/282782e9b03de3b4dee2e91b548d28fb2a95fd91/README.md) and compare with our current readme.

PR #256, opened 2020-08-05
Fix outbound transformations for experiment creation

jessie-ross:
**Currently our experiment structure isn't being camel-cased correctly, this PR fixes that.**

## How has this been tested?

### DIFF
--- a/lib/schemas.ts
+++ b/lib/schemas.ts
@@ -252,9 +252,8 @@ export type ExperimentFullNew = yup.InferType<typeof experimentFullNewSchema>
  */
 export const experimentFullNewOutboundSchema = experimentFullNewSchema
   .shape({
-    metricAssignments: yup.array(metricAssignmentNewOutboundSchema).defined(),
-    segmentAssignments: yup.array(segmentAssignmentNewOutboundSchema).defined(),
-    variations: yup.array<VariationNew>(variationNewOutboundSchema).defined().min(2),
+    // Seems to work here but not below?
+    variations: yup.array<VariationNew>(variationNewOutboundSchema).defined(),
   })
   .snakeCase()
   .transform(
@@ -264,8 +263,12 @@ export const experimentFullNewOutboundSchema = experimentFullNewSchema
       // The P2 field gets incorrectly snake_cased so we fix it here
       p_2_url: undefined,
       p2_url: currentValue.p_2_url,
-      // Not sure why but metric_assignments isn't working either so we fix it here.
+      // These two only seem to work down here rather then above?
       metric_assignments: yup.array(metricAssignmentNewOutboundSchema).defined().cast(currentValue.metric_assignments),
+      segment_assignments: yup
+        .array(segmentAssignmentNewOutboundSchema)
+        .defined()
+        .cast(currentValue.segment_assignments),
     }),
   )
 


jessie-ross:
@aaronyan this was the segments issue you found in the demo.

lib/schemas.ts
```diff
@@ -252,9 +252,8 @@ export type ExperimentFullNew = yup.InferType<typeof experimentFullNewSchema>
  */
 export const experimentFullNewOutboundSchema = experimentFullNewSchema
   .shape({
-    metricAssignments: yup.array(metricAssignmentNewOutboundSchema).defined(),
-    segmentAssignments: yup.array(segmentAssignmentNewOutboundSchema).defined(),
-    variations: yup.array<VariationNew>(variationNewOutboundSchema).defined().min(2),
+    // Seems to work here but not below?
```
yanirs, line 255: Maybe because this is the same in snake and camel case?

lib/schemas.ts
```diff
@@ -264,8 +263,12 @@ export const experimentFullNewOutboundSchema = experimentFullNewSchema
       // The P2 field gets incorrectly snake_cased so we fix it here
       p_2_url: undefined,
       p2_url: currentValue.p_2_url,
-      // Not sure why but metric_assignments isn't working either so we fix it here.
+      // These two only seem to work down here rather then above?
```
yanirs, line 266: Maybe the problem is that the outbound schemas are snake cased so the shaping shouldn't happen early?

yanirs: Also, looking at the file as a whole, why are we exporting `metricAssignmentNewOutboundSchema`, `segmentAssignmentNewOutboundSchema`, and `variationNewOutboundSchema`? Do we even need those single-use variables?

yanirs:
@jessie-ross can you please share more specific testing notes, i.e., what was broken and how it's fixed? It's hard to judge this change otherwise, especially given the inline comments that make it seem like it's unclear why the fix is even working. :slightly_smiling_face:

jessie-ross:
Sure, I should have included a bit more context to this in the first place, I think I was a bit slack with it as I was fixing code that doesn't have AT yet (But is coming soon).

Currently the experiment form isn't submitting because the segment objects aren't being converted to using snake_case for their property names.

There is some strange behaviour in snakeCase relating to whether you put child schemas in the shape or the transform.

All three of `variations`, `segmentAssignments`, and `metricAssignments` have camelCased properties that need to be converted to snake case.

I found this solution experimentally just by playing around with it, I tested it using production APIs by submitting an experiment with segments. Automated tests will be added along with further ExperimentForm tests (over and above #263).

**Testing instructions:**
1. Use a production config profile in development
2. Create a new experiment with

yanirs:
Thanks for the extra testing details! This is clearer now. I'll play with it next week unless @aaronyan beats me to it. :slightly_smiling_face:

aaronyan:
> Use a production config profile in development

lib/schemas.ts
```diff
@@ -264,8 +263,12 @@ export const experimentFullNewOutboundSchema = experimentFullNewSchema
...
       metric_assignments: yup.array(metricAssignmentNewOutboundSchema).defined().cast(currentValue.metric_assignments),
```
aaronyan, line 267: This seems to be the same transformation, but formatted differently. Consider making this consistent to be more pleasant on the eyes :)

yanirs:
> It looks good to me, but I can't quite make out why these code changes seem to work. :)

Yeah, this is so strange. I tried playing with it to get a better intuition for it but couldn't find a better solution. Might be worth reporting to `yup`, but for now this should be good enough to merge.

lib/schemas.ts
jessie-ross: > Also, looking at the file as a whole, why are we exporting metricAssignmentNewOutboundSchema, segmentAssignmentNewOutboundSchema, and variationNewOutboundSchema? Do we even need those single-use variables?

Ah sorry I didn't respond to this earlier.

I can see these being used for patching experiments.

lib/schemas.ts
jessie-ross: This is our automated formatter doing this unfortunately and the way we have it set up it doesn't allow us to make any exceptions.

PR #257, opened 2020-08-05
Allow a wider max width for long metric names

jessie-ross:
**In production there are some very long metric names, this PR allows the form to widen further.**

## How has this been tested?


jessie-ross:
> No objections to increasing the max width, but should we also use word-break: break-all where applicable to break up long snake_case words?

I was thinking of using ellipsis truncation and adding the full name to the tooltip, but I am leaving that for a second pass at the whole form.

PR #258, opened 2020-08-11
Refactor experiment pages

jessie-ross:
**We are duplicating a lot of code between the two experiment pages (soon to be 3 with snippets), this PR refactors them together.**

- Create ExperimentPageView
- Use ExperimentPageView in `/experiments/[id]` and `/experiments/[id]/results`

That the data needs to reload on every tab is a limitation of NextJS's router #212 

## How has this been tested?

- Exisiting Automated tests

### DIFF
--- a/pages/experiments/[id].tsx
+++ b/pages/experiments/[id].tsx
@@ -1,70 +1,21 @@
-import { LinearProgress } from '@material-ui/core'
-import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
 import debugFactory from 'debug'
 import { useRouter } from 'next/router'
 import { toIntOrNull } from 'qc-to_int'
 import React from 'react'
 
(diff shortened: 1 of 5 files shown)


pages/experiments/[id].tsx
```diff
@@ -1,70 +1,21 @@
...
-      marginBottom: theme.spacing(2),
-    },
-  }),
-)
-
 export default function ExperimentPage() {
-  const classes = useStyles()
   const router = useRouter()
   const experimentId = toIntOrNull(router.query.id)
+  const debugMode = router.query.debug === 'true'
```
Comment (withinboredom) [inline pages/experiments/[id].tsx:13]: nit: should we make this `router.query.debug.toLowerCase()` here?

Comment (jessie-ross) [reply pages/experiments/[id].tsx:13]: Not too fussed about this, it is not for end-user consumption and I don't think we would have problems with it.

In this future this may turn into a role-based check anyway.

PR #259, opened 2020-08-11
Add experiment code setup page

jessie-ross:
**This PR adds the code-setup AKA snippets page.**
Screen Shot 2020-08-11 at 9 27 48 pm

- Renamed Snippets to Code Setup
   I think it is an improvement over Snippets as otherwise you might now know why you would want snippets.
   Think it can be improved, let me know if you think of anything.
- Didn't implement the freshly created message.
   Our snackbar provides enough context for now. See
 https://d.pr/v/ubyma7

## How has this been tested?

- Manual testing with production data

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ npm run test:integration -- --updateSnapshot
 npm run test:unit
 npm run test:unit -- -- --watch
 npm run test:unit -- -- --updateSnapshot
+DEBUG_PRINT_LIMIT=30000 npm run test:unit # For when react-test-library truncates your debug output
 
 
 # Code formatting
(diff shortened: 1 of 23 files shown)


jessie-ross:
> I think we'll want to add some more context to help with the transition from the page redirect. I tried mocking something up!

That looks great! The extra text really helps out.


> We can also only show one dependent on their choice in platform "platform": "calypso". If we do that, we can cut down on a lot of the text in my mockup and scrap the additional tabs.

This is very easy to do 😃 


> And for the actual code snippet, we can include them in another PR if we need more definition.

Nice, also I think I can actually link the PHP file mentioned as the repo is public so I will do that, and I might add a bit more text surrounding each one.

jessie-ross:
@aaronyan I might merge this and leave the snippet part for later - the page doesn't look like much with only either the Calypso or WPCOM links but no snippets.

PR #260, opened 2020-08-11
Rename experiment details to experiment overview

jessie-ross:
I was finding 'Details' and 'Results' a bit too similar, both imply 'zoomed-in' versions of an experiment.
**This PR suggests 'Overview' instead, best seen in pictures:**

Before:

Screen Shot 2020-08-11 at 10 05 31 pm

After:

Screen Shot 2020-08-11 at 10 03 42 pm

## How has this been tested?

### DIFF
--- a/components/ExperimentPageView.tsx
+++ b/components/ExperimentPageView.tsx
@@ -26,7 +26,7 @@ const useStyles = makeStyles((theme: Theme) =>
 )
 
 export enum ExperimentView {
-  Details = 'details',
+  Overview = 'overview',
   Results = 'results',
   CodeSetup = 'code-setup',
 }
@@ -81,7 +81,7 @@ export default function ExperimentPageView({
           segments &&
           analyses && (
             <>
-              {view === ExperimentView.Details && <ExperimentDetails {...{ experiment, metrics, segments }} />}
+              {view === ExperimentView.Overview && <ExperimentDetails {...{ experiment, metrics, segments }} />}
               {view === ExperimentView.Results && (
                 <ExperimentResults {...{ experiment, metrics, analyses, debugMode }} />
               )}
--- a/components/ExperimentTabs.test.tsx
+++ b/components/ExperimentTabs.test.tsx
@@ -11,9 +11,9 @@ test('renders expected links', () => {
     metricAssignments: [],
     segmentAssignments: [],
   })
-  const { getByText } = render(<ExperimentTabs experimentId={experiment.experimentId} tab={ExperimentView.Details} />)
+  const { getByText } = render(<ExperimentTabs experimentId={experiment.experimentId} tab={ExperimentView.Overview} />)
 
-  expect(getByText('Details', { selector: '.MuiTab-wrapper' })).toBeInTheDocument()
+  expect(getByText('Overview', { selector: '.MuiTab-wrapper' })).toBeInTheDocument()
   expect(getByText('Results', { selector: '.MuiTab-wrapper' })).toBeInTheDocument()
   expect(getByText('Code Setup', { selector: '.MuiTab-wrapper' })).toBeInTheDocument()
 })
--- a/components/ExperimentTabs.tsx
+++ b/components/ExperimentTabs.tsx
@@ -48,8 +48,8 @@ export default function ExperimentTabs({
   return (
     <Tabs className={className} value={tab}>
       <LinkTab
-        label='Details'
-        value={ExperimentView.Details}
+        label='Overview'
+        value={ExperimentView.Overview}
         url='/experiments/[id]'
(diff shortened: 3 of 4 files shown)


yanirs:
> I'm not too sure what is the best way to go about these types of design suggestions and where we want to vote and discuss. As long as the suggestion is small and we can all agree quickly, I think we can use GitHub approvals as our votes. If there is a disagreement, we can iterate/close depending on the suggested iterations? What do y'all think?

While I have no objections to this particular change, I strongly object to doing it like this moving forward (PRs) for the following reasons:

* It usually takes longer to implement and review code than discuss an idea for a change. In this case, a quick question on Slack whether there are objections would have been good.
* Sometimes, seemingly small changes balloon into monster changes, e.g., if the PR author missed a detail about something else that needs to change. [A classic quote on this topic](https://blog.codinghorror.com/code-reviews-just-do-it/): _"In a software-maintenance organization, 55 percent of one-line maintenance changes were in error before code reviews were introduced. After reviews were introduced, only 2 percent of the changes were in error."_
* We have our prioritisation process for a reason. Pushing PRs like this (or parts of #237 and #253) essentially "hijacks" the prioritisation process, as it means that at least two people aren't working on the items we explicitly prioritised &ndash; a form of [priority inversion](https://en.wikipedia.org/wiki/Priority_inversion). 

I think we're better off sticking to our process:

* During our three sprint weeks, open issues or collect lists of such improvements and discuss them prior to implementation, i.e., try to avoid priority inversion as much as possible.
* Follow a more lenient approach during our unplanned weeks &ndash; I do agree that sometimes _showing_ a change and discussing it in context is quicker than discussing it as an abstract idea. So unplanned PRs on unplanned weeks should be fine. :slightly_smiling_face:

jessie-ross:
**I am happy to go with the group opinion but I would argue for the more discretionary approach similar to @aaronyan's suggestions as I don't think there is a one size fits all solution to this.**

By that I mean the developer should be able to make a call in the moment about what will be the most efficient/persuasive way to present a change.

I should say that this is also what I feel most comfortable with.

To respond to @yanirs comments:

---

>  Pushing PRs like this (or parts of #237 and #253) essentially "hijacks" the prioritisation process

Yep, this isn't working perfectly for me.

But I'd like to think I have examples where I have made good judgments about these sorts of things, when I have decided to post on slack, make P2s, and I have made some not-in-scope-but-nearby-change PRs which seemed to have gone down well at the time.

I'd also like to think that with a bit more time I will get better at understanding the culture of this project, this division and A8C as a whole and become better at making these kinds of decisions. Feedback is always welcome!

---

> It usually takes longer to implement and review code than discuss an idea for a change. In this case, a quick question on Slack whether there are objections would have been good.

There are two parts to this:
1. Longer to implement (tackled here).
2. Longer to review code (tackled below).

**Firstly, making a PR can be more persuasive:**
Some changes are easier to demonstrate in code, others are most persuasive as mockups - which it is easiest to just code up anyway (eg. this PR). Being persuasive up front - presenting your best argument - can save a lot of arguing about things.

**Secondly, it is not necessarily longer to implement:**
Sometimes editing code now will be faster than editing it later. 
You have your editor open, you have the code fresh in your mind, you know the tests relating to this code, there is no context switching needed. It is the idea of "syngergy" (to use a passe term).

---

> Sometimes, seemingly small changes balloon into monster changes, e.g., if the PR author missed a detail about something else that needs to change. A classic quote on this topic: "In a software-maintenance organization, 55 percent of one-line maintenance changes were in error before code reviews were introduced. After reviews were introduced, only 2 percent of the changes were in error."

I don't quite follow how the quote is related, it might just mean people were more careful in environments where other people were going to judge their code.

But to tackle your main point here I think there should be some expectations on Reviewers to redirect PRs, for example they might say the following:

> Hmm, this looks like a big change, let's move it to a P2 first.

> I think others might want to take a look at this, lets notify them.

> This may be too out of scope for this Milestone, can we hold it off for later?

We can pause problematic PRs in the cases where misjudgments are made, before they "balloon".

--- 

In general I sit more in the "individuals and interactions over processes and tools" camp as you guys may have noticed, but as always I am happy to go along with the team.

aaronyan:
Too bad we didn't bring this up during our retro! :) Let's make sure we discuss this next retro to see if there are ways to find a process that strikes a good balance for both:
- protecting each other's attention and maintaining focus on the milestone
- encouraging the freedom and trust to work on what we think is important

I'm not entirely sure what that looks like, but let's discuss together during the next retro (or unplanned week) if we like! :)

--- 

For now, we should try our best to stick to milestone work items during the planned weeks to protect each other's attention and focus. Let's continue using issues as the way to document, propose, and discuss enhancements and new ideas.

PR #263, opened 2020-08-14
Add basic tests for experiment form

jessie-ross:
**This PR starts the testing of the experiment form.**
- Moves the submit functionality out to the page level to allow unit testing (also for using for editing and forking)
- Unit test the stage sectioning.
- Unit test a basic valid experiment.

There is a bit of funkiness around the date mocking as I can't seem to mock the Date object before `/lib/schemas` gets loaded. This is needed as the state date has validations requiring it to be in the future.

## How has this been tested?

- N/A


README.md
```diff
@@ -43,6 +43,7 @@ npm run test:integration -- --updateSnapshot
 npm run test:unit
 npm run test:unit -- -- --watch
 npm run test:unit -- -- --updateSnapshot
+DEBUG_PRINT_LIMIT=30000 npm run test:unit # For when react-test-library truncates your debug output
```
withinboredom, line 46: hahahaha. Might be best to just use the max_int though.

```suggestion
DEBUG_PRINT_LIMIT=$(echo "console.log(Number.MAX_SAFE_INTEGER)" | node -) npm run test:unit # For when react-test-library truncates your debug output
```

components/experiment-creation/ExperimentForm.stories.tsx
```diff
@@ -13,5 +15,6 @@ export const Form = () => (
     indexedMetrics={Normalizers.indexMetrics(Fixtures.createMetricBares(20))}
     indexedSegments={Normalizers.indexSegments(Fixtures.createSegments(20))}
     initialExperiment={createInitialExperiment()}
+    onSubmit={async (formData: unknown) => alert(JSON.stringify(formData, null, 2))}
```
withinboredom, line 18: Just out of curiosity, why is this a promise? It should resolve immediately.

README.md
jessie-ross: Hmm, as far as the readme goes I think the original example is a bit easier to read and see what is going on.
A developer should know to add more zeros if needed.

components/experiment-creation/ExperimentForm.stories.tsx
jessie-ross: Sure:
- Form submission typically involves async tasks such as sending data to a server.
- Formik changes form state while a form is submitting - the `isSubmitting` flag gets set on a form.
   This flag disables all form fields, prevents further submission and is used for the loading indicator.
- For Formik to know whether submission is happening it relies on the resolution of a `Promise` from `onSubmit`.
- Typescript makes this necessary in all cases, even if Formik doesn't explicitly need `onSubmit` to return a promise (e.g. `await` doesn't care if it is awaiting a promise or not).

withinboredom: Thanks! I'm not intimately familiar with Formik, so this helped a lot!

PR #265, opened 2020-08-20
Use Normalized data in ExperimentDetails

jessie-ross:
**This PR adds normalized data usage in ExperimentDetails:**
- Adds `normalizr` schemas for experiment
- Adds `normalizeExperiment` function
- Removes `Experiments.getDeployedVariation` as it is no longer necessary
   - Add data test for `deployedVariationId` - this will check on outgoing data as well as incoming. 
- Removes `resolveMetricAssignments` in MetricsPanel
- Removes `resolveSegmentAssignments` from SegmentsPanel

Looks like a lot but isn't too much:
- 7 of the files are test files where we just change them to use normalized data.
- Other than the above removal we are just adding and threading the normalized data through.

This is part 2/3 of #236 and should be done before we do further work in ExperimentDetails for the CRUD.
- A departure is that I am using `normalizedXEntities` instead of `normalizedXData` as it makes things simpler.

See also part 1/3 at #240.

## How has this been tested?

### DIFF
--- a/components/AudiencePanel.test.tsx
+++ b/components/AudiencePanel.test.tsx
@@ -9,11 +9,11 @@ import { render } from '@/test-helpers/test-utils'
 import AudiencePanel from './AudiencePanel'
 
 test('renders as expected with no segment assignments', () => {
-  const [normalizedExperiment, normalizedExperimentData] = normalizeExperiment(Fixtures.createExperimentFull())
+  const [normalizedExperiment, normalizedExperimentEntities] = normalizeExperiment(Fixtures.createExperimentFull())
   const segments: Segment[] = []
   const indexedSegments = indexSegments(segments)
   const { container } = render(
-    <AudiencePanel {...{ normalizedExperiment, normalizedExperimentData, indexedSegments }} />,
+    <AudiencePanel {...{ normalizedExperiment, normalizedExperimentEntities, indexedSegments }} />,
   )
 
   expect(container).toMatchSnapshot()
@@ -23,11 +23,11 @@ test('renders as expected with existing users allowed', () => {
   const experiment = Fixtures.createExperimentFull({
     existingUsersAllowed: true,
   })
-  const [normalizedExperiment, normalizedExperimentData] = normalizeExperiment(experiment)
+  const [normalizedExperiment, normalizedExperimentEntities] = normalizeExperiment(experiment)
   const segments: Segment[] = []
   const indexedSegments = indexSegments(segments)
   const { container } = render(
-    <AudiencePanel {...{ normalizedExperiment, normalizedExperimentData, indexedSegments }} />,
+    <AudiencePanel {...{ normalizedExperiment, normalizedExperimentEntities, indexedSegments }} />,
   )
 
   expect(container).toMatchSnapshot()
@@ -44,9 +44,9 @@ test('renders as expected with all segments resolvable', () => {
       Fixtures.createSegmentAssignment({ segmentAssignmentId: 104, segmentId: 4 }),
     ],
   })
-  const [normalizedExperiment, normalizedExperimentData] = normalizeExperiment(experiment)
+  const [normalizedExperiment, normalizedExperimentEntities] = normalizeExperiment(experiment)
   const { container } = render(
-    <AudiencePanel {...{ normalizedExperiment, normalizedExperimentData, indexedSegments }} />,
+    <AudiencePanel {...{ normalizedExperiment, normalizedExperimentEntities, indexedSegments }} />,
   )
 
   expect(container).toMatchSnapshot()
@@ -64,7 +64,7 @@ test('throws an error when some segments not resolvable', () => {
       Fixtures.createSegmentAssignment({ segmentAssignmentId: 110, segmentId: 10 }),
     ],
   })
-  const [normalizedExperiment, normalizedExperimentData] = normalizeExperiment(experiment)
+  const [normalizedExperiment, normalizedExperimentEntities] = normalizeExperiment(experiment)
 
   // Note: This console.error spy is mainly used to suppress the output that the
   // `render` function outputs.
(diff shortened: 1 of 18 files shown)


lib/normalizers.test.ts
```diff
@@ -24,4 +24,21 @@ describe('lib/normalizers.ts module', () => {
       expect(Normalizers.indexSegments(segments)).toEqual({ 1: segments[0], 2: segments[1] })
     })
   })
+
+  describe('normalizeExperiment', () => {
+    it('normalizes an empty experiment', () => {
+      const experiment = { experimentId: 0 }
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore; This is just for minimal testing
```
yanirs, line 32: Can you please create a more meaningful test with an actual experiment? This also came up in https://github.com/Automattic/abacus/pull/240#issuecomment-666273443.

lib/schemas.ts
```diff
@@ -116,6 +118,11 @@ export const metricAssignmentSchema = metricAssignmentNewSchema
   .defined()
   .camelCase()
 export type MetricAssignment = yup.InferType<typeof metricAssignmentSchema>
+export const metricAssignmentNormalizrSchema = new normalizr.schema.Entity<MetricAssignment>(
```
yanirs, line 121: Why are these schemas here? I thought we agreed to keep them in `normalizers.ts` and not export them as in #240.

lib/schemas.ts
```diff
@@ -210,9 +227,42 @@ export const experimentFullSchema = experimentBareSchema
...
     variations: yup.array<Variation>(variationSchema).defined().min(2),
   })
+  .test(
+    'deployedVariationId-variation-check',
+    'Could not find variation matching deployedVariationId.',
+    /* istanbul ignore next; this is a data validation test itself */
+    (experiment) => {
+      return _.isNumber(experiment.deployedVariationId)
+        ? experiment.variations.some((variation: Variation) => variation.variationId === experiment.deployedVariationId)
+        : true
```
yanirs, line 237: I think this would be a bit clearer as
```suggestion
      return _.isNull(experiment.deployedVariation) ||
        experiment.variations.some((variation: Variation) => variation.variationId === experiment.deployedVariationId)
```

lib/schemas.ts
```diff
@@ -210,9 +227,42 @@ export const experimentFullSchema = experimentBareSchema
...
+export type ExperimentFullNormalized = Omit<
+  ExperimentFull,
+  'metricAssignments' | 'segmentAssignments' | 'variations'
+> & {
+  metricAssignments: number[]
+  segmentAssignments: number[]
+  variations: number[]
+}
+export interface ExperimentFullNormalizedEntities {
+  experiments: Record<number, ExperimentFullNormalized>
```
yanirs, line 261: Why do we return this and why is it plural? It doesn't seem to be used.

lib/normalizers.ts
```diff
@@ -19,3 +27,23 @@ const segmentNormalizrSchema = new schema.Entity<Segment>('segments', {}, { idAt
 export function indexSegments(segments: Segment[]): Record<number, Segment> {
   return normalize<Segment>(segments, [segmentNormalizrSchema]).entities.segments || {}
 }
+
+/**
+ * Returns a tuple of normalizedExperiment and normalizedExperimentEntities (all the entities in experiment including experiment itself).
+ */
+export function normalizeExperiment(
```
yanirs, line 34: This function returns a value that's too hard to use in my opinion. Why not define `ExperimentFullNormalized` to include the raw assignment arrays _and_ the indexed arrays? 

Taking `metricAssignments` as an example, it's very unintuitive that it's an array of `MetricAssignment`s in `ExperimentFull`, an array of numbers in `ExperimentFullNormalized`, and a mapping from ID to `MetricAssignment` in `ExperimentFullNormalizedEntities`. I think we should stick to keeping it as an array of `MetricAssignment`s in `ExperimentFullNormalized` and add an `indexedMetricAssignments` field for the mapping from ID to `MetricAssignment`. 

The same goes for `segmentAssignments` and `variations`. If we include the indexed versions in `ExperimentFullNormalized` we would have a single value that'd be much easier to use.

lib/schemas.ts
```diff
@@ -210,9 +227,42 @@ export const experimentFullSchema = experimentBareSchema
     segmentAssignments: yup.array(segmentAssignmentSchema).defined(),
     variations: yup.array<Variation>(variationSchema).defined().min(2),
   })
+  .test(
+    'deployedVariationId-variation-check',
+    'Could not find variation matching deployedVariationId.',
+    /* istanbul ignore next; this is a data validation test itself */
```
yanirs, line 233: It's not a test, though. I think it should still be covered.

components/AudiencePanel.test.tsx
```diff
@@ -1,14 +1,20 @@
...
 import Fixtures from '@/test-helpers/fixtures'
 import { render } from '@/test-helpers/test-utils'
 
 import AudiencePanel from './AudiencePanel'
 
 test('renders as expected with no segment assignments', () => {
-  const experiment = Fixtures.createExperimentFull()
-  const { container } = render(<AudiencePanel experiment={experiment} segments={[]} />)
+  const [normalizedExperiment, normalizedExperimentEntities] = normalizeExperiment(Fixtures.createExperimentFull())
+  const segments: Segment[] = []
```
yanirs, line 13: Can we avoid this temporary variable?

yanirs (changes requested):
See inline comments: The main issues I see are similar comments to those addressed on #240 and `normalizeExperiment()` being too unintuitive and a leaky abstraction of the `normlizr` library. I haven't looked through all the files yet since my main comments would affect everything else.

lib/schemas.ts
jessie-ross: Ah yep, I forgot about that.

jessie-ross:
I am still thinking about this, I am now not so convinced of the urgency for it and the method of execution...

yanirs:
> I am still thinking about this, I am now not so convinced of the urgency for it and the method of execution...

@jessie-ross I'll close this for now as it's very stale. We can always revisit in the future. :slightly_smiling_face:

Issue #13, opened 2020-04-14
Add basic metric creation page


yanirs:
Aaron edited:

Consider building metric details page first (#162). We can consider making the metric creation just be an empty metric details page.

## Metric creation:
- name (snake case, lowercase)
- description
- indicator for `is_higher_better` (checkbox or slider button for on/off)
- radio box for kind of metric (revenue vs conversion)

### Revenue metric:
- refund days dropdown (1, 3, 7, 30) 
- transaction types checklist (can select multiple) (new purchase, recurring, cancelation, etc). Ping @aaronyan for full list.
- product slugs dropdown (but allow additional slugs to be added/removed with a +add product button)
- all fields optional (I think)

For future iteration, make refund days and transaction types options populated by API endpoint.

### Conversion metric:
- event name (form field, with future autocomplete) - allow events to be added/removed with a +add event button
- property name and values (form field with future autocomplete) - allow properties to be added/removed with a +add property button
- properties are attached to events and events can have multiple properties
- only one event name is required

jessie-ross:
Can we cut this down a bit, especially since this won't be used by end users:

Instead of in depth revenue and conversion parameter controls, can we have a single "Params" textbox where directly type in the corresponding JSON, with live validation happening over that JSON.

aaronyan:
A simplified metric creation sounds good to me! Let's also add a skeleton of the JSON for the "params" textbox that changes depending on the type of metric (revenue vs conversion).

So metric creation will then have these fields:
- name (snake case, lowercase)
- description
- indicator for `is_higher_better` (checkbox or slider button for on/off)
- radio box for kind of metric (revenue vs conversion)
- params JSON textbox

If we go the JSON skeleton route, they could look like this:

Conversion metric
```
[
    {
        "event": "event_name",
        "props": {
                "prop_key": "prop_value"
        }
    }
]
```

Revenue metric
```
{
    "refund_days": 7,
    "transaction_types": [],
    "product_slugs": []
}
```

yanirs:
I don't think we should cut it down and use JSON editing:

* The conversion event props are the same as what we need to support adding exposure events (#262 &ndash; a higher priority issue), so we will have a component for that.
* The revenue event props aren't too nested and should be simple to implement as a proper form.

jessie-ross:
@aaronyan Yep, that is exactly right.

@yanirs It is basically instant to add it with JSON editing, can't we improve it later?

yanirs:
[...]

PR #267, opened 2020-08-20
Add metric CRUD modal state

jessie-ross:
**This PR adds modal state for Metric add/edit, all behind the scenes.**

I am trying this out as stacked PRs, more to come, let me know how you find it.

Part of a series on #13

## How has this been tested?

- Automated tests
- More testing once it is done

### DIFF
--- a/pages/metrics/index.tsx
+++ b/pages/metrics/index.tsx
@@ -1,7 +1,7 @@
 import { Button, createStyles, LinearProgress, makeStyles, Theme } from '@material-ui/core'
 import debugFactory from 'debug'
 import { useRouter } from 'next/router'
-import React from 'react'
+import React, { useState } from 'react'
 
 import MetricsApi from '@/api/MetricsApi'
 import Layout from '@/components/Layout'
@@ -30,7 +30,17 @@ const MetricsIndexPage = () => {
   const router = useRouter()
   const debugMode = router.query.debug === 'true'
 
-  const onEditMetric = (metricId: number) => alert(metricId)
+  // Edit Metric Modal
+  const [editMetricMetricId, setEditMetricMetricId] = useState<number | null>(null)
+  const isEditingMetric = editMetricMetricId !== null
+  const { isLoading: editMetricIsLoading, data: editMetricInitialMetric, error: editMetricError } = useDataSource(async () => {
+    return editMetricMetricId !== null ? await MetricsApi.findById(editMetricMetricId) : null
+  }, [editMetricMetricId])
+  useDataLoadingError(error, 'Metric to edit')
+  const onEditMetric = (metricId: number) => {
+    setEditMetricMetricId(metricId)
+  }
+
   const onAddMetric = () => alert('add metric')
 
   return (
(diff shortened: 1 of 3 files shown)


pages/metrics/index.tsx
```diff
@@ -1,11 +1,13 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */ // Temporary
```
yanirs, line 1: I'm not a fan of these file-level suppressions. They're easy to forget (seems like that's the case in `ExperimentForm.tsx`). If we must suppress warnings due to partial implementation, I think we should suppress on specific lines.

pages/metrics/index.tsx
```diff
@@ -30,8 +32,31 @@ const MetricsIndexPage = () => {
...
-  const onAddMetric = () => alert('add metric')
+  // Edit Metric Modal
+  const [editMetricMetricId, setEditMetricMetricId] = useState<number | null>(null)
+  const isEditingMetric = editMetricMetricId !== null
+  const {
+    isLoading: editMetricIsLoading,
+    data: editMetricInitialMetric,
+    error: editMetricError,
+  } = useDataSource(async () => {
+    return editMetricMetricId !== null ? await MetricsApi.findById(editMetricMetricId) : null
```
yanirs, line 43: Double negatives are harder to read (the else is `if !(editMetricMetricId !== null)`). Should reverse the condition or use`isEditingMetric`.

jessie-ross: I don't see any problem with this but why not.

pages/metrics/index.tsx
jessie-ross: Sure lets do that.

pages/metrics/index.tsx
yanirs: See https://www.refactoring.com/catalog/reverseConditional.html for example.

pages/metrics/index.tsx
yanirs: Can you please remove it then?

jessie-ross: It's fine if I remove it in the next PR I am making for this series, it makes this all simpler to reason about.

yanirs: Which PR do you mean? I think that removing it here would make it easier to reason about, as it'd make it more explicit what's used and unused. In general, most unused variables can be introduced when they're actually needed &ndash; otherwise reviewing the stack is harder. And as I said, file-level suppressions tend to get forgotten.

Is it that much of an effort to remove it here so we can move on with this PR?

jessie-ross: Yeah it is a bit more effort: having to edit it, perform a local `verify` (and again if I make a mistake), I will have to check for the verify/build on GH (sometimes they fail randomly), there will likely be a merge conflict later. The local verify takes 4 minutes and I am personally not great at context switching so the time I spend waiting for a verify is not very efficient.

It seems like we have different values over developer-time surrounding this. We agree on the general file-level suppressions topic but it is just these small exceptions where we are working things out we seem to have different approaches, I am usually a "let's change this going forward" while you are a more "let's change this now". The aggravating factor seems to be that we need to do this fairly often as we don't have a lot of common code culture already set up and we have to work through many different issues.

I'll leave this discussion for the retro and make the change here.

pages/metrics/index.tsx
jessie-ross: That example makes sense but I still don't personally feel it for this specific case.

pages/metrics/index.tsx
yanirs: Happy to discuss further at the retro, but just a few points:

* If making changes as part of a review becomes something we avoid, then reviews become pointless. 
* The "change it later" bit is directly addressed by [the Google code review guidelines we adopted](https://google.github.io/eng-practices/review/reviewer/pushback.html#later). In this specific case, the change would make the code _easier_ to reason about. In general, most suppressions can be avoided.
* There's no need for a full local `verify` when fixing a lint issue. :slightly_smiling_face:

pages/metrics/index.tsx
yanirs: I don't see what's special about this case but it doesn't matter since you changed it. :slightly_smiling_face:

pages/metrics/index.tsx
yanirs: By the way, I think that a part of the issue in this case is that this PR is a bit _too_ small and could have been a part of #268. It's hard to find the balance sometimes. :disappointed:

jessie-ross: That makes sense, I am new to using stacked PRs like this so it is a work in progress 😄

PR #268, opened 2020-08-20
Add Metric add/edit modal dialogs

jessie-ross:
**This PR adds add and edit modal dialogs**

To try it out visit `/metrics?debug=true`.

Part of a series on #13

## How has this been tested?

### DIFF
--- a/pages/metrics/index.tsx
+++ b/pages/metrics/index.tsx
@@ -1,10 +1,20 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */ // Temporary
-import { Button, createStyles, LinearProgress, makeStyles, Theme, Dialog, DialogTitle, DialogContent, DialogContentText, TextField, DialogActions } from '@material-ui/core'
+/* eslint-disable @typescript-eslint/no-unused-vars, @typescript-eslint/require-await */ // Temporary
+import {
+  Button,
+  createStyles,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+  LinearProgress,
+  makeStyles,
+  Theme,
+} from '@material-ui/core'
 import debugFactory from 'debug'
-import { useRouter } from 'next/router'
-import React, { useState } from 'react'
 import { Formik } from 'formik'
+import { useRouter } from 'next/router'
 import { useSnackbar } from 'notistack'
+import React, { useState } from 'react'
 
 import MetricsApi from '@/api/MetricsApi'
 import Layout from '@/components/Layout'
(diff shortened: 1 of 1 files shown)


pages/metrics/index.tsx
```diff
@@ -1,7 +1,19 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */ // Temporary
-import { Button, createStyles, LinearProgress, makeStyles, Theme } from '@material-ui/core'
+/* eslint-disable @typescript-eslint/no-unused-vars, @typescript-eslint/require-await */ // Temporary
```
yanirs, line 1: Same comment as on #267: Please suppress specific lines to make it clearer what's being suppressed and why.

pages/metrics/index.tsx
```diff
@@ -72,6 +102,45 @@ const MetricsIndexPage = () => {
...
+                  <Button color='primary' type='submit'>
+                    Save
+                  </Button>
+                </DialogActions>
+              </form>
+            )}
+          </Formik>
+        )}
+      </Dialog>
+      <Dialog open={isAddingMetric} aria-labelledby='add-metric-form-dialog-title'>
```
yanirs, line 126: I see that you separated the form fields into a component in #269, but I think we can take it a step further and share more code between the dialogs with a common local component, which would make it a bit easier to see the similarities and keep them aligned.

jessie-ross: I think it may be a bit early to do this sort of refactoring, maybe once we see the other dialog forms we will know more what an abstraction over this should look like.

pages/metrics/index.tsx
jessie-ross: See my reply there

pages/metrics/index.tsx
yanirs: What other dialog forms? These are two things that are practically identical and already share functionality.

jessie-ross: The dialog forms for:
- Editing the general panel
- Adding/Editing the Conclusion
- Adding a MetricAssignment

yanirs: I don't see how they're related to this PR. :slightly_smiling_face: Here we have two dialogs for editing or adding the same thing. As #269 shows, it means that we need to remember to make changes in two places. It seems to me like we can share the code on a higher level than the `MetricFormFields`, e.g., if we just had `Save` as the the button text in both cases, the `form` level would be more natural.

Anyway, not a blocker. I'm happy to revisit later if you prefer.

jessie-ross: Yeah lets revisit this later.

PR #269, opened 2020-08-20
Add metric CRUD basic form fields

jessie-ross:
**This PR adds basic form fields to the metric CRUD dialogs.**

Screen Shot 2020-08-20 at 7 56 48 pm

To try it out visit `/metrics?debug=true`

Part of a series on #13

## How has this been tested?


jessie-ross:
> Woohoo! This is coming together nicely. Tested and the form validation works as expected.

Ah the actual validation is yet to come, but it is a very quick to add in.
Once we get the conventions around these done they should take something like half a day to implement each, they are all basically the same.

yanirs (approved):

Issue #196, opened 2020-07-04
Support concluding an experiment


yanirs:
See #179 for buttons/toolbar we can use. The general idea is to support populating the experiment's conclusion fields if they're empty and the experiment's status is `completed` or `disabled`.

jessie-ross:
With the current UI in Figma the conclusion controls and the conclusion form are strangely disconnected:

![Markup 2020-08-18 at 13 42 14](https://user-images.githubusercontent.com/971886/90474800-b3e93c00-e158-11ea-8aaf-3143d86d8601.png)
 

For UI I would like to suggest something like GitHub's review button:
<img width="719" alt="Screen Shot 2020-08-18 at 1 39 28 pm" src="https://user-images.githubusercontent.com/971886/90474629-489f6a00-e158-11ea-8f74-e306404d35f5.png">

We can use [MUI's popover](https://material-ui.com/components/popover/)

This should be more intuitive for the user and result in simpler code.

yanirs:
I agree with improving the connection between the two components, but I'm not sure I follow how it'd work. Currently, we don't show the conclusion panel if the experiment hasn't been concluded, so do you mean we'd have the `Add Conclusion` button with popover in this case? If so, it sounds good to me.

How will we support editing the conclusions? With a separate edit button for the conclusion panel, as suggested in #195? If so, I like it. :slightly_smiling_face:

jessie-ross:
> How will we support editing the conclusions? With a separate edit button for the conclusion panel, as suggested in #195? If so, I like it. 🙂

Yep I think that would be good.

> Currently, we don't show the conclusion panel if the experiment hasn't been concluded, so do you mean we'd have the Add Conclusion button with popover in this case?

Yeah I was thinking of having this at the top in the "toolbar" opposite the tabs, but now that I think about it further we could have an empty "Conclusion" panel with an green Add conclusion button in it.

jessie-ross:
I might include @withinboredom and @aaronyan for this and #195

Issue #195, opened 2020-07-04
Support updating an experiment's details


yanirs:
See #179 for buttons/toolbar we can use. The idea is to only allow editing specific fields depending on the status. When the time comes to implementation, I think it's worth discussing again whether we want to do it via inline editing.

jessie-ross:
Following on from my suggestions on #196, I find having the edit button on top a bit unintuitive so I would like to suggest an alternative approach here.

**For a staging experiment,** for the MVP, I see no problem with reusing the experiment creation wizard and I have built it for that purpose. The edit button will take them to the same wizard with all their fields pre-filled. The text would just need to be changed slightly.

**For a non-staging experiment**:

<img width="796" alt="Screen Shot 2020-08-18 at 1 56 15 pm" src="https://user-images.githubusercontent.com/971886/90475868-ef850580-e15a-11ea-92c5-bad0765e4447.png">


- Having Seperate edit buttons for each section (general and adding metrics).
- Use modal forms for these changes.

My reasoning:
- There is no disconnect between the overall form controls and the 
- It is clear what is editable.
- It is simple to disable the add-metrics button after completion.
- We can include a tooltip as well indicating why different things are disabled, teaching our Experimenter.
- Modals provide much better code separation and will be cleaner to implement.
- Modals forms are a very well known pattern.


To go further with this:
- Again I would love to see "General" and "Audience" split into "Metadata", "Experiment Parameters", and "Audience", I think it will make the UI behaviour make even more sense
- Keep the top-level "Edit Experiment", "Conclude Experiment" and the general-panel "Edit" and metric-panel "Add Metric" in there all the time, but show them as disabled when they are not available (with tooltips saying why). This can be nice UI as the buttons are always in the same spot and you learn what is available and when.

jessie-ross:
Ah it also makes a lot of sense with the tab design.

yanirs:
I like the idea of separate buttons and modals. 

Not entirely sure about "going further":

> Again I would love to see "General" and "Audience" split into "Metadata", "Experiment Parameters", and "Audience", I think it will make the UI behaviour make even more sense

I don't see how this separation would help, as the description and owner can always be edited, and the end date can be edited while the experiment is running.

[...]

PR #270, opened 2020-08-21
Cut down LabelValuePanel abstraction to LabelValueTable

jessie-ross:
**The `LabelValuePanel` abstraction is not suitable for the upcoming CRUD behaviours, we cut it down to just being the table.**

`LabelValuePanel` does two things: The panel, the table. This PR makes it just do the table.

Why does this need to happen:
- Need to be able to add buttons to the panel header for Conclusion + General
- Need to be able to change the ConclusionPanel content to have the "Add Conclusion" button
- The MetricAssignmentPanel is already an exception to it.

Part of #195, #196 

**Also I didn't carry forward the strange padding "feature" as it was a rather leaky part of the abstraction and didn't seem necessary:**

Before:

Screen Shot 2020-08-22 at 1 16 13 am

After:

Screen Shot 2020-08-22 at 1 16 28 am

- Best to stick with defaults
- It was confusing seeing the outer table rows in line with the inner table
- Simplifies the code, it was a messy/leaky part of the abstraction - are we going to include future attributes like this, it would be much better just to have something like a `rowProps` property but even that is leaky as we would need to know about the row. 

## How has this been tested?


components/AudiencePanel.tsx
```diff
@@ -81,7 +95,14 @@ function AudiencePanel({ experiment, segments }: { experiment: ExperimentFull; s
       ),
     },
   ]
-  return <LabelValuePanel data={data} title='Audience' />
+  return (
+    <Paper>
+      <Typography className={classes.title} color='textPrimary' variant='h3'>
```
withinboredom, line 100: non sequitur: I really wish these kinds of libraries would just generate a bunch of variants so that you could also just do `<H3 className={classes.title} color='textPrimary'>` instead of this "variant" bs.

`</rant>`

withinboredom (approved):

components/AudiencePanel.tsx
jessie-ross: I think it has its benefits:
- It won't get confused with the lowercase `<h3>` tag.
- It allows for more variants in a consistent way e.g. `variant="body1"` instead of `<Body1`.
- It allows the code behind it to be simpler - only one component.

And the difference in practice is negligible.

PR #272, opened 2020-08-21
Add general edit modal fields

jessie-ross:
**This PR adds fields to the General Edit Modal.**

This one is pretty straight forward:

Screen Shot 2020-08-22 at 3 41 21 am

## How has this been tested?

- E2E testing for all of these to come at the end.


components/GeneralPanel.tsx
```diff
@@ -66,7 +83,10 @@ function GeneralPanel({ experiment }: { experiment: ExperimentFull }) {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [isEditing, setIsEditing] = useState<boolean>(false)
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
yanirs, line 85: Can some of these be removed now?

components/GeneralPanel.tsx
```diff
@@ -91,12 +111,85 @@ function GeneralPanel({ experiment }: { experiment: ExperimentFull }) {
...
         <DialogTitle id='edit-experiment-general-form-dialog-title'>Edit Experiment: General</DialogTitle>
         <Formik initialValues={{ experiment: generalEditInitialExperiment }} onSubmit={onSubmitEdit}>
           {(formikProps) => (
             <form onSubmit={formikProps.handleSubmit}>
-              <DialogContent></DialogContent>
+              <DialogContent>
+                <div className={classes.row}>
+                  <Field
+                    component={TextField}
+                    id='experiment.p2Url'
```
yanirs, line 123: We don't allow edits to the P2 URL for non-staging experiments.

components/GeneralPanel.tsx
```diff
@@ -91,12 +111,85 @@ function GeneralPanel({ experiment }: { experiment: ExperimentFull }) {
...
+                      shrink: true,
+                    }}
+                  />
+                </div>
+
+                <div className={classes.row}>
+                  <Field
+                    component={TextField}
+                    className={classes.datePicker}
+                    name='experiment.endDatetime'
```
yanirs, line 159: The experiment end datetime is only editable if the experiment is still running. Shall we disable/hide this field if it's not running?

jessie-ross: Hmm, knowing what's editable when is going to be a bit tricky for users.

jessie-ross: This is what I've done to it:

Screen Shot 2020-08-25 at 11 18 08 pm

yanirs: Looks good! 

This is a bit out of scope of this PR, but we should hide the `Edit` button on this panel if the experiment status is `staging`. If we don't, the "cannot be changed as the experiment has finished" text isn't necessarily true. This also applies to other edit buttons, so might be better to address separately.

yanirs (approved):

PR #273, opened 2020-08-21
Add experiment add metric modal

jessie-ross:
**This PR adds a modal for Assigning Metrics.**

Screen Shot 2020-08-22 at 4 23 04 am

Exactly the same deal as #271 

## How has this been tested?

- E2E tests at the end.


components/MetricAssignmentsPanel.tsx
```diff
@@ -109,6 +140,24 @@ function MetricAssignmentsPanel({ experiment, metrics }: { experiment: Experimen
           ))}
         </TableBody>
       </Table>
+      <Dialog open={isAssigningMetric} aria-labelledby='assign-metric-form-dialog-title'>
+        <DialogTitle id='assign-metric-form-dialog-title'>Add Metric</DialogTitle>
+        <Formik initialValues={{ metricAssignment: assignMetricInitialAssignMetric }} onSubmit={onSubmitAssignMetric}>
+          {(formikProps) => (
+            <form onSubmit={formikProps.handleSubmit}>
+              <DialogContent></DialogContent>
```
withinboredom, line 148: Is a list of metrics supposed to go here?

withinboredom (changes requested):
I think we should go ahead and implement the list of metrics in this PR or put this behind some kind of feature flag. As more and more people come to check out Abacus, we probably need for the app to not appear "broken."

image

components/MetricAssignmentsPanel.tsx
jessie-ross: Yep, I decided to stop where I was here as I had made about 5 PRs and didn't want to get too ahead of myself.
This feature should be done within the next day or so.

jessie-ross:
> I think we should go ahead and implement the list of metrics in this PR or put this behind some kind of feature flag. As more and more people come to check out Abacus, we probably need for the app to not appear "broken."

I was under the assumption that our App wasn't released yet - I would be developing pretty differently if it was.

This is a bit frustrating combining it with the split PRs - we would need to use feature flags a lot as none of the modals are working yet. I was hoping to get some efficiency gains out of doing the modals like this.

In general for a high stability app I suggest moving away from CI to versioned releases.

yanirs:
> I was under the assumption that our App wasn't released yet - I would be developing pretty differently if it was.

I think we can assume that from the end of the current milestone it'd be in active use, as this is what we're aiming for (full functionality that's needed to manage and run experiments).

> This is a bit frustrating combining it with the split PRs - we would need to use feature flags a lot as none of the modals are working yet. I was hoping to get some efficiency gains out of doing the modals like this.

In my opinion, it's fine to have the "broken" modals for now, as they'll be going away in the next week or two. Alternatively, you could also stack PRs from a root PR and then have a final big merge, or use a feature flag as you suggested.

> In general for a high stability app I suggest moving away from CI to versioned releases.

I'm not fussed either way. Can you please P2 this point? We will need to start caring more about stability very soon.

withinboredom (approved):
Thanks for comments @yanirs and @jessie-ross, let's merge this!

jessie-ross:
> Can you please P2 this point? We will need to start caring more about stability very soon.

I mentioned quite a few different approaches (including non-CI) to stability/correctness here: https://betterexperiments.wordpress.com/2020/07/06/thoughts-on-front-end-code-testing/

It might be good to discuss some of those points there.

PR #274, opened 2020-08-21
Remove ignores and fix remaining test coverage on Experiment Creation

jessie-ross:
**This PR removes the coverage and eslint ignores from Experiment Creation, fixing the tests where need be.**

## How has this been tested?

- Existing Automated tests pass.


components/experiment-creation/ExperimentForm.tsx
```diff
@@ -188,11 +186,13 @@ const ExperimentForm = ({
         }
 
         const prevStage = () => {
+          // istanbul ignore next; This condition should never happen
```
yanirs, line 189: Shouldn't this and line 195 be `istanbul ignore else`? It's the `else` that should never happen.

PR #278, opened 2020-08-26
Add experiment creation exposure events

### LINKED ISSUE #262: Add exposure events to experiment creation flow
Experimenters should have the option to specify exposure events. This should probably go under the Metrics step, as it affects the analysis.

jessie-ross:
**This PR adds Exposure Events to Experiment Creation:**

Screen Shot 2020-08-26 at 2 08 47 pm

- Video: https://d.pr/v/Wu8qsX
- Internally in the form I use an array of `{ key, value }` objects to represent the properties. It needs to be an array in form state to work with the `<FieldArray>`.
- Added extra integration testing: form state conversion to outbound.
- Added extra integration testing: using a real form state example to test `ExperimentsApi.create`

Fixes #262 

## How has this been tested?

   - Create experiment using production data
   - Verify the fields are correct in the database

### DIFF
--- a/components/experiment-creation/Metrics.tsx
+++ b/components/experiment-creation/Metrics.tsx
@@ -228,21 +228,21 @@ const Metrics = ({ indexedMetrics }: { indexedMetrics: Record<number, MetricBare
                               }}
                               InputProps={
                                 indexedMetrics[metricAssignment.metricId].parameterType ===
-                                  MetricParameterType.Conversion
+                                MetricParameterType.Conversion
                                   ? {
-                                    endAdornment: (
-                                      <InputAdornment position='end'>
-                                        <Tooltip title='Percentage Points'>
-                                          <Typography variant='body1' color='textSecondary'>
-                                            pp
+                                      endAdornment: (
+                                        <InputAdornment position='end'>
+                                          <Tooltip title='Percentage Points'>
+                                            <Typography variant='body1' color='textSecondary'>
+                                              pp
                                             </Typography>
-                                        </Tooltip>
-                                      </InputAdornment>
-                                    ),
-                                  }
+                                          </Tooltip>
+                                        </InputAdornment>
+                                      ),
+                                    }
                                   : {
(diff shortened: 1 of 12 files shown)


components/experiment-creation/Metrics.tsx
```diff
@@ -269,6 +295,171 @@ const Metrics = ({ indexedMetrics }: { indexedMetrics: Record<number, MetricBare
           )
         }}
       />
+
+      <Typography variant='h4' gutterBottom className={classes.exposureEventsTitle}>
+        Exposure Events (Optional)
+      </Typography>
+
+      <FieldArray
```
yanirs, line 303: I think that this should be a separate component so we can reuse it for metric events. This is a case where we know for sure that we want the exact same behaviour, as the handling of exposure and metric events is identical throughout the backend. It's non-trivial code so duplicating it isn't going to do us any favours.

yanirs:
The video looks good! I'm not sure about the centred alignment of add buttons, but this is a minor concern that can be addressed later. A more major concern is adding exposure events with a reusable component, as noted inline.

components/experiment-creation/Metrics.tsx
jessie-ross: Yeah I am planning to put this one in a component, it is a bit tougher since it involves some more thinking about the schema for this so I would like to do it in a seperate PR.

jessie-ross:
The master tests might fail once this get's merged.

Fixed in https://github.com/Automattic/abacus/pull/284/commits/83ceb92c942ab6abc3292bf060a1c480401f56d6 which should be merged soon.

jessie-ross:
Oops, I have accidentally squashed #284 into here!

yanirs:
> Oops, I have accidentally squashed #284 into here!

No worries, I had a look at #284 earlier so I roughly know which parts are from there (and there's also the commit history). :slightly_smiling_face:

yanirs (approved):
Looks great! :ship: 

As noted earlier, I think we can make some minor styling adjustments, but this can wait until after it's a separate component.

PR #281, opened 2020-08-26
Add edit button and dialog to Conclusions panel

yanirs:
As part of addressing #196, this PR follows the same pattern as the editing modals so far, adding the conclusion edit dialog without the actual submission capabilities. 

Peek 2020-08-27 17-54

## How has this been tested?

- Manual testing with production data (see gif)

### DIFF
--- a/components/ConclusionsPanel.test.tsx
+++ b/components/ConclusionsPanel.test.tsx
@@ -1,3 +1,4 @@
+import { fireEvent, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react'
 import * as notistack from 'notistack'
 import React from 'react'
 
@@ -235,4 +236,36 @@ test('renders as expected without deployed variation', () => {
       </div>
     </div>
   `)
+})
+
+const queryEditDialog = () => screen.queryByRole('heading', { name: /Edit Experiment: Conclusions/ })
+
+test('opens and saves edit dialog', async () => {
+  const experiment = Fixtures.createExperimentFull()
+  const { container: _container } = render(<ConclusionsPanel experiment={experiment} />)
+
+  const editButton = screen.getByRole('button', { name: /Edit/ })
+  fireEvent.click(editButton)
+
+  await waitFor(queryEditDialog)
+
+  const saveButton = screen.getByRole('button', { name: /Save/ })
+  fireEvent.click(saveButton)
+
+  await waitForElementToBeRemoved(queryEditDialog)
+})
+
+test('opens and cancels edit dialog', async () => {
+  const experiment = Fixtures.createExperimentFull()
+  const { container: _container } = render(<ConclusionsPanel experiment={experiment} />)
+
+  const editButton = screen.getByRole('button', { name: /Edit/ })
+  fireEvent.click(editButton)
+
+  await waitFor(queryEditDialog)
+
+  const cancelButton = screen.getByRole('button', { name: /Cancel/ })
+  fireEvent.click(cancelButton)
+
+  await waitForElementToBeRemoved(queryEditDialog)
 })
--- a/components/ConclusionsPanel.test.tsx
+++ b/components/ConclusionsPanel.test.tsx
@@ -1,9 +1,17 @@
+import * as notistack from 'notistack'
 import React from 'react'
 
 import Fixtures from '@/test-helpers/fixtures'
 import { render } from '@/test-helpers/test-utils'
 
 import ConclusionsPanel from './ConclusionsPanel'
+
+jest.mock('notistack')
+const mockedNotistack = notistack as jest.Mocked<typeof notistack>
+mockedNotistack.useSnackbar.mockImplementation(() => ({
+  enqueueSnackbar: jest.fn(),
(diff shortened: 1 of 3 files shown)


components/ConclusionsPanel.tsx
```diff
@@ -38,12 +63,106 @@ function ConclusionsPanel({ experiment }: { experiment: ExperimentFull }) {
...
+                    InputLabelProps={{
+                      shrink: true,
+                    }}
+                  />
+                </div>
+
+                <div className={classes.row}>
+                  <FormControl component='fieldset'>
+                    <FormLabel component='legend'>Deployed variation</FormLabel>
+                    <Field
```
yanirs, line 137: @jessie-ross I pushed this now to ask about the Formik/Material way of getting this to behave. The problem would probably be obvious to you. :slightly_smiling_face: It sets the default radio button, but then I can't select any other option. What am I missing?

jessie-ross: Option values will end up as strings in the DOM since they go into html attributes.
At the same time the Radio component is just looking at whatever it receives from above.
So above the component is a number and below is a string.

The fix is to make this consistent at the component level: `value={String(variation.variationId)}`.

Also `value={experiment.deployedVariationId}` shouldn't be there, Formik handles form state.

components/ConclusionsPanel.test.tsx
```diff
@@ -172,3 +237,35 @@ test('renders as expected without deployed variation', () => {
...
+
+  const editButton = screen.getByRole('button', { name: /Edit/ })
+  fireEvent.click(editButton)
+
+  await waitFor(queryEditDialog)
+
+  const saveButton = screen.getByRole('button', { name: /Save/ })
+  fireEvent.click(saveButton)
+
+  await waitForElementToBeRemoved(queryEditDialog)
```
yanirs, line 255: @jessie-ross looks like this bit is missing from other dialog tests since I'm getting this warning in the console when I run them: `Warning: You seem to have overlapping act() calls, this is not supported. Be sure to await previous act() calls before making a new one.`

I can fix the other tests in a separate PR if this looks right. I also think it's better to separate the tests for save and cancel since their behaviour will diverge once we add form submission.

jessie-ross: Ah oops! it looks like I wasn't using async in the other test nice catch!
 
Makes sense to seperate them.

I would personally inline `queryEditDialog` so the tests are a bit easier to read.

I will leave the full review to @withinboredom

yanirs: Thanks @jessie-ross!

> I would personally inline `queryEditDialog` so the tests are a bit easier to read.

I had it inlined before but I actually found it less readable because it isn't as obvious that `screen.queryByRole('heading', { name: /Edit Experiment: Conclusions/ })` is querying the edit dialog. I'm not too happy with the name `queryEditDialog`, though. I guess a name like `editDialogHeading` would make it read more like a sentence (`waitFor(editDialogHeading)`), but it makes it less clear that it's actually a function. Naming things is hard... I'm open to suggestions!

Related: https://m.signalvnoise.com/hunting-for-great-names-in-programming/

components/ConclusionsPanel.tsx
```diff
@@ -38,12 +63,104 @@ function ConclusionsPanel({ experiment }: { experiment: ExperimentFull }) {
...
+                <div className={classes.row}>
+                  <Field
+                    component={TextField}
+                    name='experiment.endReason'
+                    id='experiment.endReason'
+                    label='Reason the experiment ended'
+                    placeholder='Completed successfully'
+                    variant='outlined'
+                    fullWidth
+                    required
```
yanirs, line 112: Made this the only required field since experiments that move from `staging` to `disabled` have this field set to `Discarded` automatically, so this is the only field we rely on for Slack pings to conclude the experiment.

The GIF in the description is from when I had the conclusion URL required (before https://github.com/Automattic/abacus/pull/281/commits/1ae7b6d95750ea72973ffce15059ccca0a320b1e).

components/ConclusionsPanel.test.tsx
jessie-ross: That was an interesting read!

Might mention [this book's chapter on names (freely available online and not specific to Clojure)](https://leanpub.com/elementsofclojure/read_sample) which is one of my favourite essays on the topic.

components/ConclusionsPanel.tsx
jessie-ross: Yeah this doesn't matter so much as we will be doing validation in javascript rather than use HTML validation. I kind of just added it on before but I am not sure it even adds anything.

components/ConclusionsPanel.test.tsx
jessie-ross: As an aside I was also thinking literal programming is of great value for testing, even just adding "headers" to delineate code sections (you may have seen my `// ### Heading` comments around, it helps to be able to have a double space before each heading to help when you scan code.).

jessie-ross (approved):
LGTM 🚀

PR #283, opened 2020-08-27
Add reload capabilities to useDataSource

jessie-ross:
**This PR enables data from `useDataSource` to be reloaded**
- Uses `isSubscribed` to ensure subscription state as things can get messier now.
   See https://juliangaramendy.dev/use-promise-subscription/ for how it works.
- Extends the return object to include `reloadRef`. 
   I'd argue it is simpler to reason about and use (particularly in combination with the internal isSubscribed state - it just works) while not particularly changing the code that uses it - it is easy to see it is a ref and know that we need to call `reloadRef.current()` instead of the hypothetical alternative `reload()`.

This change is necessary for all our update logic that will very soon need to reload data into the views.

Something to note is that the loading indicator can occur while we have old data present, we will be using this in our UIs - we should be both showing the old data and an indicator that the new data is loading.

## How has this been tested?

### DIFF
--- a/utils/data-loading.ts
+++ b/utils/data-loading.ts
@@ -25,9 +25,6 @@ export function useDataSource<Data, Deps extends DependencyList | undefined, E e
     let isSubscribed = true
     // For isSubscribed to work with reloading we need to use reload as a Ref
     reloadRef.current = () => {
-      if (!isSubscribed) {
-        return
-      }
       setIsLoading(true)
       createDataPromise()
         .then((data) => isSubscribed && setData(data))
--- a/utils/data-loading.ts
+++ b/utils/data-loading.ts
@@ -1,5 +1,6 @@
+import _ from 'lodash'
 import { useSnackbar } from 'notistack'
-import { DependencyList, useEffect, useState } from 'react'
+import { DependencyList, useEffect, useRef, useState } from 'react'
 
 /**
  * Declarative data loader.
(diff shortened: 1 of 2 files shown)


utils/data-loading.ts
```diff
@@ -15,12 +16,26 @@ export function useDataSource<Data, Deps extends DependencyList | undefined, E e
   const [isLoading, setIsLoading] = useState<boolean>(false)
   const [data, setData] = useState<Data | null>(null)
   const [error, setError] = useState<E | null>(null)
+  // eslint-disable-next-line @typescript-eslint/unbound-method
```
yanirs, line 19: Can avoid these by using `import noop from 'lodash/noop'` and using `noop` directly.

utils/data-loading.ts
```diff
@@ -15,12 +16,26 @@ export function useDataSource<Data, Deps extends DependencyList | undefined, E e
...
+  const reloadRef = useRef<() => void>(_.noop)
+
   useEffect(() => {
-    setIsLoading(true)
-    createDataPromise()
-      .then(setData)
-      .catch(setError)
-      .finally(() => setIsLoading(false))
+    // The isSubscribed logic is necessary to prevent setStates after unmounts or dependency changes
```
yanirs, line 23: It'd be useful to reference https://juliangaramendy.dev/use-promise-subscription/ here.

utils/data-loading.ts
```diff
@@ -15,12 +16,26 @@ export function useDataSource<Data, Deps extends DependencyList | undefined, E e
...
+    let isSubscribed = true
+    // For isSubscribed to work with reloading we need to use reload as a Ref
+    reloadRef.current = () => {
+      isSubscribed && setIsLoading(true)
```
yanirs, line 27: I think this would read better a spelled-out `if`:

```suggestion
      if (isSubscribed) {
        setIsLoading(true)
      }
````

File: utils/data-loading.ts [hunk 4]
```diff
@@ -29,6 +44,7 @@ export function useDataSource<Data, Deps extends DependencyList | undefined, E e
     data,
     isLoading,
     error,
+    reloadRef,
```
yanirs, line 47: Now that the return value is getting more complicated, it'd be good to add a `@return` part to the function's documentation and explain how each element is meant to be used.

yanirs (approved):
Looks good. Please see inline comments before merging. :slightly_smiling_face:

utils/data-loading.ts
jessie-ross: I have resolved this by returning if not subscribed which simplifies it out.

jessie-ross: Actually after writing a test for it I am realising checking for `isSubscribed` here is completely unnecessary, since caller has a ref they will always have the latest version of this function and `isSubscribed = false => reloadRef.current = noop`.

Issue #225, opened 2020-07-21
Implement e2e testing for experiment creation


jessie-ross:
This issue is mainly here to remind the implementer to consider all `istanbul ignores` upon implementation of the tests so that no coverage is missed.

jessie-ross:
Fixed by #284

PR #284, opened 2020-08-27
Improve Experiment Creation testing

jessie-ross:
**This PR fixes code coverage ignores and adds basic e2e testing for experiment creation.**
- Reviewed all istanbul ignores.
  - Unit tested the custom `exactly-one-params` data check
  - Unit tested the custom `experimentFullNewSchema->endDatetime` check
- Added e2e test to ensure the experiment creation form loads

This is roughly the level of testing I am going for with ExperimentForm as I have moved most of the tests to unit testing and don't want to double up with e2e.

To summarise where we are at with Experiment Form testing:
   - Form-stage navigation 
   - Form-stage validation
   - Filling out a form like a user would
   - Checking the values are in the right place
   - Checking the form data is valid
   - Integration test with our Mock server
   - Production tests

The only thing we aren't specifically testing are all the validation errors that can happen, I say we mostly don't exhaustively test this for a few reasons:
- The backend provides another layer of protection from the frontend being too general
- The frontend is guarded against being too specific with our live production input data checks that are symmetrical with output
- The chain `yup <-> formik <-> formik-material-ui <-> material-ui` is very solid. All highly used production libraries with extensive testing. We are testing all the custom parts but the rest are standard and descriptive code, it might be worth reviewing the yup schemas with a fine comb at some point if we want this to be airtight but I think this is very strong. 

I was unsure whether to P2 this but decided to put this here, let me know if you would like me to move it.

## How has this been tested?

### DIFF
--- a/__tests__/api/ExperimentsApi.test.ts
+++ b/__tests__/api/ExperimentsApi.test.ts
@@ -63,54 +63,52 @@ describe('ExperimentsApi.ts module', () => {
 
       const newExperiment = experimentFullNewOutboundSchema.cast(rawNewExperiment)
 
-      expect(newExperiment).toMatchInlineSnapshot(`
-        Object {
-          "description": "experiment description",
-          "end_datetime": "2020-09-03",
-          "existing_users_allowed": "true",
-          "exposure_events": Array [
-            Object {
-              "event": "event_name",
-              "props": Object {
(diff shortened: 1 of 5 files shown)


aaronyan:
@jessie-ross if merged, does this resolve #225?

PR #287, opened 2020-08-31
Add exposure events and status

### LINKED ISSUE #217: Add exposure events and status to experiment details view
In the experiments details view, we forgot to include the `exposure_events` and `status`! :) 

Let's add `exposure_events` to the Audience card with the current grouping design. Set as low priority for now in case we change the grouping design.

And let's add `status` as it's own row to the General card right above `Dates`. We can use the same `ExperimentStatus` component that is used in the Experiments List.

withinboredom:
resolves #217 

This adds exposure events and the status to the experiment details pane. I chose to use several layers of `LabelValueTable` because this sort-of provides a "tree-like" structure which is the best way to visualize the data IMHO. I'm happy to use a different component and/or a different approach. I also chose to keep the `ExposureEventsTable` as a non-reusable component because 1) it can be easily refactored into one if we chose to and 2) it's incredibly simple such that I'm not sure we need to make it it's own file yet. 3) I'm not sure what the conventions are around here yet :)

## How has this been tested?

### DIFF
--- a/components/AudiencePanel.tsx
+++ b/components/AudiencePanel.tsx
@@ -91,7 +91,7 @@ function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
           <span className={classes.eventName}>{ev.event}</span>
           {ev.props &&
             Object.entries(ev.props).map(([key, val]) => (
-              <span key={key + val} className={classes.entry}>
+              <span key={key} className={classes.entry}>
                 {key}: {val}
               </span>
             ))}
--- a/components/AudiencePanel.tsx
+++ b/components/AudiencePanel.tsx
@@ -52,6 +52,31 @@ const useStyles = makeStyles(() =>
   }),
 )
 
+function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
+  return (
+    <>
+      {experiment.exposureEvents?.map((ev) => {
+        return (
+          <LabelValueTable
+            key={ev.event}
+            data={[
+              {
+                label: ev.event,
+                value: Object.entries(ev.props ?? {}).map((entry) => (
+                  <LabelValueTable
+                    key={ev.event + entry[0] + entry[1]}
+                    data={[{ label: entry[0], value: <>{entry[1]}</> }]}
+                  />
+                )),
+              },
+            ]}
+          />
+        )
+      })}
+    </>
+  )
+}
+
 /**
  * Renders the audience information of an experiment in a panel component.
  *
@@ -90,6 +115,10 @@ function AudiencePanel({ experiment, segments }: { experiment: ExperimentFull; s
           />
         </>
       ),
+    },
+    {
+      label: 'Exposure Events',
+      value: <ExposureEventsTable experiment={experiment} />,
     },
   ]
   return (
--- a/components/AudiencePanel.tsx
+++ b/components/AudiencePanel.tsx
@@ -63,10 +63,6 @@ const eventStyles = makeStyles(() =>
     eventName: {
       fontFamily: theme.custom.fonts.monospace,
     },
-    propsList: {
-      margin: 0,
-      paddingInlineStart: 0,
-    },
     eventList: {
       '& p:not(:first-child)': {
         paddingTop: theme.spacing(2),
@@ -83,22 +79,22 @@ const eventStyles = makeStyles(() =>
 
 function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
   const classes = eventStyles()
+  const exposureEvents =
+    !Array.isArray(experiment.exposureEvents) || experiment.exposureEvents.length === 0
+      ? null
+      : experiment.exposureEvents
 
   return (
     <div className={classes.eventList}>
-      {experiment.exposureEvents?.map((ev) => (
+      {exposureEvents?.map((ev) => (
         <Typography key={ev.event}>
           <span className={classes.eventName}>{ev.event}</span>
-          {ev.props && (
-            <ul className={classes.propsList}>
-              {' '}
-              {Object.entries(ev.props).map(([key, val]) => (
-                <li key={key + val} className={classes.entry}>
-                  {key}: {val}
-                </li>
-              ))}
-            </ul>
-          )}
+          {ev.props &&
+            Object.entries(ev.props).map(([key, val]) => (
+              <span key={key + val} className={classes.entry}>
+                {key}: {val}
+              </span>
+            ))}
         </Typography>
       )) ?? <Typography>No exposure events defined</Typography>}
     </div>
--- a/components/AudiencePanel.tsx
+++ b/components/AudiencePanel.tsx
@@ -8,6 +8,7 @@ import SegmentsTable from '@/components/SegmentsTable'
 import VariationsTable from '@/components/VariationsTable'
 import { ExperimentFull, Segment, SegmentAssignment, SegmentType } from '@/lib/schemas'
 import * as Variations from '@/lib/variations'
+import theme from '@/styles/theme'
 
 /**
  * Resolves the segment ID of the segment assignment with the actual segment.
@@ -52,28 +53,55 @@ const useStyles = makeStyles(() =>
   }),
 )
 
+const eventStyles = makeStyles(() =>
+  createStyles({
+    entry: {
+      display: 'block',
+      fontFamily: theme.custom.fonts.monospace,
+      color: 'gray',
+    },
+    eventName: {
+      fontFamily: theme.custom.fonts.monospace,
+    },
+    propsList: {
+      margin: 0,
+      paddingInlineStart: 0,
+    },
+    eventList: {
+      '& p:not(:first-child)': {
+        paddingTop: theme.spacing(2),
+      },
+      '& p': {
+        paddingBottom: theme.spacing(2),
+      },
+      '& p:not(:last-child)': {
+        borderBottom: '1px solid rgb(224,224,224)',
+      },
+    },
+  }),
+)
+
 function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
+  const classes = eventStyles()
+
   return (
-    <>
-      {experiment.exposureEvents?.map((ev) => {
-        return (
-          <LabelValueTable
-            key={ev.event}
-            data={[
-              {
-                label: ev.event,
-                value: Object.entries(ev.props ?? {}).map((entry) => (
-                  <LabelValueTable
-                    key={ev.event + entry[0] + entry[1]}
-                    data={[{ label: entry[0], value: <>{entry[1]}</> }]}
-                  />
-                )),
-              },
-            ]}
-          />
-        )
-      })}
-    </>
+    <div className={classes.eventList}>
+      {experiment.exposureEvents?.map((ev) => (
+        <Typography key={ev.event}>
+          <span className={classes.eventName}>{ev.event}</span>
+          {ev.props && (
+            <ul className={classes.propsList}>
+              {' '}
+              {Object.entries(ev.props).map(([key, val]) => (
+                <li key={key + val} className={classes.entry}>
+                  {key}: {val}
+                </li>
+              ))}
+            </ul>
+          )}
+        </Typography>
+      )) ?? <Typography>No exposure events defined</Typography>}
+    </div>
   )
 }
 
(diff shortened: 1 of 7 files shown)


jessie-ross:
The code and code decisions looks good, but I'm not too sure about the design.

Screen Shot 2020-09-01 at 11 53 06 am

I can some potential issues here:
- Event names can be quite long, that can really mess with everything.
- There will be a lot of custom text - every instance of 'string', 'additionalPropX' could be any text. This means there can be a big blob of unstructured text over that space - it is hard to tell what you are looking at, are `additionalProps` the events?
- Bold text so far hasn't meant user-data but rather serves as a header for user-data, this could be confusing.

(I do still think JSON is a good representation for these especially being so data-intense and having custom object properties, but I don't think I could convince @yanirs.)

This is quite a hard component to style as we are presenting a lot of complex and structured information within a larger also complex display.

I'm not too sure these will work but how about:
- Not using LabelValueTable, it doesn't do much anyway, very easy to inline and will make the below simpler.
- Having `eventName` on its own row above the properties. Have `<strong>Event:</strong> event_name`
- Use monospace fonts, we should be using it for all user input: `fontFamily: theme.custom.fonts.monospace`
- Having props below it but indented(? not crazy about this but might be good for now).
- It also looks like you've run into a problem of deeply nested tables not showing the borders between items, might be worth fixing, but I could fix in a future PR if it is too troublesome.

yanirs:
> (I do still think JSON is a good representation for these especially being so data-intense and having custom object properties, but I don't think I could convince @yanirs.)

Maybe JSON would work for the props? We can also consider moving exposure events to a separate panel, since they're technically a part of the metric calculation rather than of the audience.

Another option is to mimic the way we have it in funnels (props under the event name and greyed). Example:

image
 
Anyway, I think we can push it with minor tweaks and iterate on the design. @aaronyan may also have some design ideas.

jessie-ross:
Yeah I think greying the props might help, we might not need the 'Event:' text if that's the case.

I also like the simple `key: value` style so the props themselves don't have to be in a table.

withinboredom:
Maybe something like this:

image

PS. Last commit will fail tests (I had a hard stop so will fix here shortly)

jessie-ross:
Yeah that is what I mean as far as the layout.

I am not sure if you mean that image literally but if it is I would say let's reduce it a bit further:
- No need for 'Event:' now.
- The underline is new design language, it might be a bit confusing.
- Let's stick to greyscale to keep it simple and not distracting from other parts of the page.
- With the grey props and use of colons (`:`), it should be obvious the event is an event, I don't think it needs the indent any more.

jessie-ross:
This is what it might look with the points above:

Screen Shot 2020-09-01 at 11 59 18 pm

It will look best once monospace is uniform throughout.

jessie-ross:
Ahh, it is also a bit strange that Exposure Events is with Audience, but why not for now. 
I would still love to have a Meta-Parameters section.

components/AudiencePanel.tsx
```diff
@@ -52,6 +53,45 @@ const useStyles = makeStyles(() =>
...
+}))
+
+function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
+  const classes = eventStyles();
+
+  const renderProps = (props: {} | null | undefined) => {
+    if(props === undefined || props === null || props == {}) return null
+    const asArray = Object.entries(props)
+    return <ul className={classes.propsList}>
+      {asArray.map(entry => <li className={classes.entry}>{entry[0]}:{entry[1]}</li>)}
```
jessie-ross, line 79: It is good to use list destructuring for this sort of thing:
```
      {asArray.map(([key, val]) => <li className={classes.entry}>{key}:{val}</li>)}

File: components/AudiencePanel.tsx [hunk 2]
```diff
@@ -52,6 +53,45 @@ const useStyles = makeStyles(() =>
...
+    </ul>
+  }
+
+  return (
+    <>
+      {experiment.exposureEvents?.map((ev) =>
+              <Typography key={ev.event}>
```
jessie-ross, line 86: Might want to add a style here for multiple events, ensuring a border between and a nice padding on both sides (`theme.spacing(2 or 3 or 4)` is your friend here.)

components/AudiencePanel.tsx
```diff
@@ -52,6 +53,45 @@ const useStyles = makeStyles(() =>
...
+    marginTop: 0
+  }
+}))
+
+function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
+  const classes = eventStyles();
+
+  const renderProps = (props: {} | null | undefined) => {
+    if(props === undefined || props === null || props == {}) return null
+    const asArray = Object.entries(props)
```
jessie-ross, line 77: `eventPropertyEntries` might be a nice name here

components/AudiencePanel.tsx
```diff
@@ -52,6 +53,45 @@ const useStyles = makeStyles(() =>
...
+  propsList: {
+    marginTop: 0
+  }
+}))
+
+function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
+  const classes = eventStyles();
+
+  const renderProps = (props: {} | null | undefined) => {
+    if(props === undefined || props === null || props == {}) return null
```
jessie-ross, line 76: No need to check `props ==(=?) {}` as the rest of the code will just work.

This means this can be collapsed quite a bit:

`{exposureEvents && <ul> Object.entries(exposureEvents).map(([key, val]) => <li> key:val </li> </ul>}`

components/AudiencePanel.tsx
```diff
@@ -52,6 +53,45 @@ const useStyles = makeStyles(() =>
...
+  },
+  propsList: {
+    marginTop: 0
+  }
+}))
+
+function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
+  const classes = eventStyles();
+
+  const renderProps = (props: {} | null | undefined) => {
```
jessie-ross, line 75: A good general rule is that any function that produces JSX should be a component.

This function signature is particularly strange as `props` doesn't follow the usual conventions, `props` is always an object and it normally contains data on a property of `props`.

In this case I would just inline it.

components/AudiencePanel.tsx
```diff
@@ -52,6 +53,45 @@ const useStyles = makeStyles(() =>
...
+    if(props === undefined || props === null || props == {}) return null
+    const asArray = Object.entries(props)
+    return <ul className={classes.propsList}>
+      {asArray.map(entry => <li className={classes.entry}>{entry[0]}:{entry[1]}</li>)}
+    </ul>
+  }
+
+  return (
+    <>
+      {experiment.exposureEvents?.map((ev) =>
```
jessie-ross, line 85: Might also want to show "No exposure events" if there are no exposure events.

withinboredom:
Thanks for all the amazing feedback @jessie-ross!

image

Much, much better. Still probably not perfect, but it looks great to my untrained eye... ha!

components/AudiencePanel.tsx
```diff
@@ -52,6 +53,58 @@ const useStyles = makeStyles(() =>
...
+    },
+    eventList: {
+      '& p:not(:first-child)': {
+        paddingTop: theme.spacing(2),
+      },
+      '& p': {
+        paddingBottom: theme.spacing(2),
+      },
+      '& p:not(:last-child)': {
+        borderBottom: '1px solid rgb(224,224,224)',
```
withinboredom, line 78: I'm not sure if this is the proper way to do colors here :thinking:

withinboredom:
I have no idea what the tests are complaining about, it looks like everything passed to me.

yanirs:
@withinboredom it looks much better!

> Ahh, it is also a bit strange that Exposure Events is with Audience, but why not for now.
> I would still love to have a Meta-Parameters section.

@jessie-ross yeah, it being in Audience isn't great, but good enough for now. Shuffling things around might be a good thing to tackle in the next milestone &ndash; can note it in our planning meeting. :slightly_smiling_face: 

> I have no idea what the tests are complaining about, it looks like everything passed to me.

@withinboredom it's the coverage:
```
Jest: "global" coverage threshold for statements (100%) not met: 99.75%
Jest: "global" coverage threshold for branches (100%) not met: 98.33%
Jest: "global" coverage threshold for lines (100%) not met: 99.74%
Jest: "global" coverage threshold for functions (100%) not met: 99.07%
```

Adding a test case for `AudiencePanel` with non-null exposure events should fix it.

components/AudiencePanel.tsx
jessie-ross: Generally prefer to use the "#" notation but it doesn't really matter all that much 😉

components/AudiencePanel.tsx
```diff
@@ -52,6 +53,58 @@ const useStyles = makeStyles(() =>
...
+              {' '}
+              {Object.entries(ev.props).map(([key, val]) => (
+                <li key={key + val} className={classes.entry}>
+                  {key}: {val}
+                </li>
+              ))}
+            </ul>
+          )}
+        </Typography>
+      )) ?? <Typography>No exposure events defined</Typography>}
```
jessie-ross, line 103: There are two cases that could give no exposure events:
- `exposureEvents === null`
- `exposureEvents === [] `

It looks to me like only the first is covered here:

Screen Shot 2020-09-02 at 11 27 15 am

withinboredom: I'll fix it when I add tests (I had assumed we'd already had tests for all the cases).

components/AudiencePanel.tsx
```diff
@@ -91,7 +91,7 @@ function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
           <span className={classes.eventName}>{ev.event}</span>
           {ev.props &&
             Object.entries(ev.props).map(([key, val]) => (
-              <span key={key + val} className={classes.entry}>
+              <span key={key} className={classes.entry}>
```
jessie-ross, line 94: Nice! I don't think it would have mattered either way but this is what I would call the preferred choice.

withinboredom: I'm more concerned with more than one event in the array having the same name and then the same key. Dunno what to do about that...

jessie-ross: Yep that's a good point, we could add it to our yup schema very easily later on, but is it much of a problem if that happens? Do we track exposures separately or are they combined?

withinboredom: In theory, there won't be duplicate event names (I would guess that behavior is undefined in the analysis). It's probably a non-issue.

components/AudiencePanel.tsx
```diff
@@ -52,6 +53,54 @@ const useStyles = makeStyles(() =>
...
+      '& p:not(:last-child)': {
+        borderBottom: '1px solid rgb(224,224,224)',
+      },
+    },
+  }),
+)
+
+function ExposureEventsTable({ experiment }: { experiment: ExperimentFull }) {
+  const classes = eventStyles()
+  const exposureEvents =
```
jessie-ross, line 82: Might be clearer for a `const isExposureEvents = experiment.exposureEvents && experiment.exposureEvents !== []`

jessie-ross (approved):
See my last comment but otherwise good to go! 🚀

components/AudiencePanel.tsx
withinboredom: I thought about that, but for whatever reason decided to be extra defensive. I'll update it to be more permissive for clarity.

withinboredom: Ah, that doesn't work :) 

```
[] && [] !== []
true && false
true
```

Which isn't what we want.

withinboredom: That should be much better now :)

components/AudiencePanel.tsx
```diff
@@ -77,26 +77,26 @@ const eventStyles = makeStyles(() =>
...
-          <span className={classes.eventName}>{ev.event}</span>
-          {ev.props &&
-            Object.entries(ev.props).map(([key, val]) => (
-              <span key={key} className={classes.entry}>
-                {key}: {val}
-              </span>
-            ))}
-        </Typography>
-      )) ?? <Typography>No exposure events defined</Typography>}
+      {exposureEvents && exposureEvents.length ? (
```
jessie-ross, line 85: Nice! Yeah the ternary is the right way to go here.

PR #290, opened 2020-09-01
Add GeneralPanel Edit submit

jessie-ross:
**This PR adds submission to the GeneralPanel Edit Modal:**
- Adds `ExperimentsApi.patch`
- Uses `ExperimentsApi.patch` for the submit.
- Beefs up the Save button in the modal.

Currently fails integration tests and local as swagger is failing, it seems to be requiring the `name` property to be present?

Next and Final PR is calling the refresh on the experiment data.

## How has this been tested?

### DIFF
--- a/__tests__/api/ExperimentsApi.test.ts
+++ b/__tests__/api/ExperimentsApi.test.ts
@@ -1,13 +1,15 @@
 import { format } from 'date-fns'
+import MockDate from 'mockdate'
 
 import ExperimentsApi from '@/api/ExperimentsApi'
 import { ExperimentFullNew, experimentFullNewOutboundSchema } from '@/lib/schemas'
 import { validationErrorDisplayer } from '@/test-helpers/test-utils'
 
+MockDate.set('2020-08-13')
+
 describe('ExperimentsApi.ts module', () => {
   describe('create', () => {
     it('should transform a new experiment into a valid outbound form', () => {
-      // We need to make some dates relative to today since mocking the schema to work with MockDate is a pain!
       const now = new Date()
       now.setDate(now.getDate() + 1)
       const nextWeek = new Date()
@@ -112,7 +114,6 @@ describe('ExperimentsApi.ts module', () => {
     })
 
     it('should create a new experiment', async () => {
-      // We need to make some dates relative to today since mocking the schema to work with MockDate is a pain!
       const now = new Date()
       now.setDate(now.getDate() + 1)
       const nextWeek = new Date()
--- a/__tests__/api/ExperimentsApi.test.ts
+++ b/__tests__/api/ExperimentsApi.test.ts
@@ -185,7 +185,7 @@ describe('ExperimentsApi.ts module', () => {
         ownerLogin: 'owner-nickname',
       }
       const returnedExperiment = await validationErrorDisplayer(
-        ExperimentsApi.patch(0, (rawNewExperiment as unknown) as Partial<ExperimentFull>),
+        ExperimentsApi.patch(1, (rawNewExperiment as unknown) as Partial<ExperimentFull>),
       )
       expect(returnedExperiment.experimentId).toBeGreaterThan(0)
     })
@@ -204,7 +204,7 @@ describe('ExperimentsApi.ts module', () => {
       }
       try {
         await validationErrorDisplayer(
-          ExperimentsApi.patch(0, (rawNewExperiment as unknown) as Partial<ExperimentFull>),
+          ExperimentsApi.patch(1, (rawNewExperiment as unknown) as Partial<ExperimentFull>),
         )
       } catch (e) {
         expect(e).toMatchInlineSnapshot(`[Error: Cannot patch experiment: contains blacklisted key.]`)
--- a/__tests__/api/ExperimentsApi.test.ts
+++ b/__tests__/api/ExperimentsApi.test.ts
@@ -2,7 +2,7 @@ import { format } from 'date-fns'
 import MockDate from 'mockdate'
 
 import ExperimentsApi from '@/api/ExperimentsApi'
-import { ExperimentFullNew, experimentFullNewOutboundSchema, ExperimentFull } from '@/lib/schemas'
+import { ExperimentFull, ExperimentFullNew, experimentFullNewOutboundSchema } from '@/lib/schemas'
 import { validationErrorDisplayer } from '@/test-helpers/test-utils'
 
 MockDate.set('2020-08-13')
--- a/__tests__/api/ExperimentsApi.test.ts
+++ b/__tests__/api/ExperimentsApi.test.ts
@@ -189,27 +189,6 @@ describe('ExperimentsApi.ts module', () => {
       )
       expect(returnedExperiment.experimentId).toBeGreaterThan(0)
     })
-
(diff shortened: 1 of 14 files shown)


api/ExperimentsApi.ts
```diff
@@ -24,6 +25,39 @@ async function create(newExperiment: ExperimentFullNew) {
...
+  const blacklistedKeys = [
+    'experimentId',
+    'exposureEvents',
+    'segmentAssignments',
+    'variations',
+    // TODO: Unimplemented as we aren't snake_casing nested fields yet.
+    //       We will add special logic for this one...
+    'metricAssignments',
+  ]
+  if (blacklistedKeys.some((key) => keys.includes(key))) {
```
yanirs, line 45: Do we need this to be a runtime check? Can't we define `experimentPatch` to be of type `Partial<Omit<ExperimentFull, 'experimentId'  | 'exposureEvents' | ...>>`?

jessie-ross: Well runtime has some advantages over compile-time, generally these functions (`patch` and `create`) don't actually get well typed inputs as we are receiving from user data. I would argue in this case it is preferable to do this runtime.

Although it is arguable whether this is useful at all...

jessie-ross:
Ok, finally got to the failing integration tests:

> Currently fails integration tests and local as swagger is failing, it seems to be requiring the name property to be present?

@yanirs it might be worth you experimenting with swagger as you have it set up.

yanirs:
> Ok, finally got to the failing integration tests:
> 
> > Currently fails integration tests and local as swagger is failing, it seems to be requiring the name property to be present?
> 
> @yanirs it might be worth you experimenting with swagger as you have it set up.

It was because I didn't specify 200 as a possible response in the mock. :man_facepalming:
Fixed on SwaggerHub and I'll push a PR to update the doc.

__tests__/api/ExperimentsApi.test.ts
```diff
@@ -173,6 +173,45 @@ describe('ExperimentsApi.ts module', () => {
...
+      now.setDate(now.getDate() + 1)
+      const nextWeek = new Date()
+      nextWeek.setDate(now.getDate() + 7)
+      const rawNewExperiment = {
+        description: 'experiment description',
+        endDatetime: format(nextWeek, 'yyyy-MM-dd'),
+        ownerLogin: 'owner-nickname',
+      }
+      const returnedExperiment = await validationErrorDisplayer(
+        ExperimentsApi.patch(0, (rawNewExperiment as unknown) as Partial<ExperimentFull>),
```
yanirs, line 188: Let's use a realistic experiment ID (> 0).

api/ExperimentsApi.ts
yanirs: > Well runtime has some advantages over compile-time, generally these functions (`patch` and `create`) don't actually get well typed inputs as we are receiving from user data. I would argue in this case it is preferable to do this runtime.

Looking at how it's called (with the output of `_.pick()`), there is some type checking going on. Also, an advantage of using the `Partial<Omit<...>>` approach is that the allowed fields are explicitly declared in the function signature. 

> Although it is arguable whether this is useful at all...

Yeah, maybe drop it completely? The backend will reject any fields that can't be updated anyway, and the list of blacklisted keys isn't exhaustive. If we do end up supporting additional fields on the backend, it'll be less work here. :slightly_smiling_face:

components/GeneralPanel.test.tsx
```diff
@@ -185,6 +192,12 @@ test('checks edit dialog does not allow end datetime changes with disabled exper
   const experiment = Fixtures.createExperimentFull({ status: Status.Disabled })
   const { container: _container } = render(<GeneralPanel experiment={experiment} />)
 
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  global.fetch = jest.fn().mockImplementationOnce(async () => ({
+    json: async () => experiment,
+  }))
+
```
yanirs, line 200: This is unnecessary as this test doesn't submit the form.

```suggestion
```

components/GeneralPanel.test.tsx
```diff
@@ -163,6 +164,12 @@ test('opens, submits and cancels edit dialog with running experiment', async ()
   const experiment = Fixtures.createExperimentFull({ status: Status.Running })
   const { container: _container } = render(<GeneralPanel experiment={experiment} />)
 
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  global.fetch = jest.fn().mockImplementationOnce(async () => ({
```
yanirs, line 169: As this is a unit test, it'd be better to mock `ExperimentsApi.patch()` and verify that it's called with the expected fields on save. Otherwise, this test isn't really testing much (the behaviour of `Save` is identical to that of `Cancel`).

yanirs (changes requested):
Let's make the panel test verify that Save and Cancel do different things. :slightly_smiling_face: 

Also, I did a quick check [on Vercel](https://abacus-git-add-general-experiments-submit.a8c.vercel.app/experiments/2) and we shouldn't be validating disabled fields:

Peek 2020-09-02 08-22

components/GeneralPanel.tsx
```diff
@@ -102,11 +119,16 @@ function GeneralPanel({ experiment }: { experiment: ExperimentFull }) {
   const onCancelEdit = () => {
     setIsEditing(false)
   }
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/require-await
-  const onSubmitEdit = async (formData: unknown) => {
-    // TODO: Full submission
-    enqueueSnackbar('Experiment Updated!', { variant: 'success' })
-    setIsEditing(false)
+  const onSubmitEdit = async (formData: { experiment: unknown }) => {
```
yanirs, line 122: I think we can use `{ experiment: typeof generalEditInitialExperiment }` here.

components/GeneralPanel.tsx
```diff
@@ -46,6 +48,21 @@ const useStyles = makeStyles((theme: Theme) =>
         color: theme.palette.text.hint,
       },
     },
+    submitContainer: {
```
yanirs, line 51: Shall we move some of these styles to a higher level? We're losing out on CSS benefits by localising this stuff.

jessie-ross: It might be worth extracting this into a `ProgressButtonContainer` component

components/GeneralPanel.test.tsx
jessie-ross: Mocking `ExperimentsApi.patch()` was a good idea!

components/GeneralPanel.test.tsx
jessie-ross: It happens to be needed now as we have slightly different submit logic for disabled experiments.

jessie-ross:
Ok, I've fixed the issues you have mentioned, these in particular:
- Made sure `endDatetime` isn't validated when it isn't editable.
- Extracted out `LoadingButtonContainer`

yanirs (approved):
I think we can improve the testing a bit, but nothing for this PR, so let's :ship:

PR #291, opened 2020-09-01
Add GeneralPanel Edit Modal data reload

jessie-ross:
This PR adds reloading after submit:
- Passes down the reload function to the modal.
- Shows the linear loader independently to the data state.

## How has this been tested?

### DIFF
--- a/components/GeneralPanel.test.tsx
+++ b/components/GeneralPanel.test.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/require-await */
 import { fireEvent, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react'
+import { noop } from 'lodash'
 import MockDate from 'mockdate'
 import * as notistack from 'notistack'
 import React from 'react'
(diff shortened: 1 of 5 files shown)


yanirs (approved):
I haven't tested this manually but the code looks good. Feel free to merge this into #290 before addressing the comments there &ndash; I'm happy to look at everything together once the comments are addressed. :slightly_smiling_face:

yanirs:
> I haven't tested this manually but the code looks good. 

Hmm... But now I see that the unit tests are failing. I might push something here or to a separate PR since I'm relying on this branch for the conclusion fields.

components/GeneralPanel.test.tsx
```diff
@@ -162,7 +165,7 @@ test('renders as expected', () => {
 
 test('opens, submits and cancels edit dialog with running experiment', async () => {
   const experiment = Fixtures.createExperimentFull({ status: Status.Running })
-  const { container: _container } = render(<GeneralPanel experiment={experiment} />)
+  render(<GeneralPanel experiment={experiment} experimentReloadRef={experimentReloadRef} />)
```
yanirs, line 168: It's probably worth using a real mock here and checking that it's called on save, but that can be done as part of the test changes in #290.

PR #292, opened 2020-09-01
Add fields to add Metric Assignment Modal

jessie-ross:
This PR adds fields to the Add MetricAssignment Modal:

Screen Shot 2020-09-02 at 2 29 07 am

- Allows you to add a new metric assignment and have it set to primary, no way to change primary otherwise, this should be good enough for now.

## How has this been tested?

- More testing to come later

### DIFF
--- a/components/MetricAssignmentsPanel.test.tsx
+++ b/components/MetricAssignmentsPanel.test.tsx
@@ -107,7 +107,7 @@ test('renders as expected with all metrics resolvable', () => {
               >
                 metric_1
                 <span
-                  class="makeStyles-primary-1 makeStyles-root-3"
+                  class="makeStyles-primary-1 makeStyles-root-6"
                 >
                   Primary
                 </span>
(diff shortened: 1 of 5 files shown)


yanirs:
> Allows you to add a new metric assignment and have it set to primary, no way to change primary otherwise, this should be good enough for now.

We don't support changing the primary metric for non-staging experiments, so we should exclude the "make primary" field. It may also be worth changing the heading of the modal to "Add Secondary Metric" to make this more explicit.

jessie-ross:
> We don't support changing the primary metric for non-staging experiments, so we should exclude the "make primary" field.

Nice! That makes it a lot simpler!

> It may also be worth changing the heading of the modal to "Add Secondary Metric" to make this more explicit.

I think it will be just fine not including it, if an Experimenter never knows it is an option it shouldn't be a problem.

jessie-ross:
## Update

- Fixed the tests
- Extract test helper function `changeFieldByRole`

withinboredom (approved):
LGTM.

PR #288, opened 2020-09-01
GeneralPanel Edit Modal Validation

jessie-ross:
**This PR adds validation to the GeneralPanel edit modal:**
- Video: https://d.pr/v/jpDZgI
- Uses `yup.reach` and creates `yupPick` (the `_.pick` equivalent for yup) allowing picking subsets of a schema
- Fixes date testing in the schema by using `.test()` rather than `.min()` allowing mocking of `Date`
- Moves `endDatetime` tests from `experimentFullNewSchema` to `experimentBareSchema` so it will flow through to `experimentFullSchema` and allow use in editing (as well as better validating incoming data).

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)
- More testing to come!

PR #295, opened 2020-09-02
Add validation to MetricAssignment assign modal

jessie-ross:
**This PR adds validation to the modal, much the same as #288.** 
Video: https://d.pr/v/a5S0Wq

## How has this been tested?


aaronyan (changes requested):
Looking good so far! A couple of requests for changes:

- After selecting a metric, the `attribution window` goes red. Can we only make the `attribution window` go red after the user interacts with that form?
- The `minimum difference` form accepts negative numbers. Let's only accept positive numbers.
- The `minimum difference` hint is currently "-". Can we adjust that to be `1.30` or some other placeholder value?
- I wasn't able to test a metric that is type conversions. Does the `minimum difference` field change to % instead of $?

yanirs:
@jessie-ross it now expects the min difference to be an integer:

Peek 2020-09-04 13-14

jessie-ross:
> it now expects the min difference to be an integer:

Ah! This should be fixed now.

yanirs (approved):

Issue #300, opened 2020-09-03
Replace our custom fetch wrapper with a library


jessie-ross:
We have our own custom fetch wrapper, I think it would be worthwhile to replace it with a library like [wretch](https://github.com/elbywan/wretch).

This should be a fairly small change.

Problems with the current fetch:
- See https://github.com/Automattic/abacus/pull/303/commits/b38156f9b084403f13d0eec3a4a86152bb619faa
- Easier to make mistakes with, more maintaining and testing to do.

PR #303, opened 2020-09-03
Implement experiment disable

jessie-ross:
**This PR implements the experiment disable functionality**

Screen Shot 2020-09-04 at 1 35 23 am

- Had to fix our `fetch` abstraction to allow requests without JSON responses.
- The 'Run' functionality will be identical to this, may refactor some of the logic out - I usually have a custom `useResolvable` hook for this sort of thing.

## How has this been tested?


__tests__/api/ExperimentsApi.test.ts
```diff
@@ -191,6 +191,16 @@ describe('ExperimentsApi.ts module', () => {
     })
   })
 
+  describe('changeStatus', () => {
+    it('should disable an existing experiment', async () => {
+      await validationErrorDisplayer(ExperimentsApi.changeStatus(1, Status.Disabled))
+    })
+
+    it('should run an existing experiment', async () => {
+      await validationErrorDisplayer(ExperimentsApi.changeStatus(1, Status.Running))
```
aaronyan, line 200: Do these tests run independently? If not, then this might be trying to change the status from `disabled` to `running` because of the previous test setting the experiment status to `disabled`.

And do we want some `expect()` here? Maybe checking for `200` response?

aaronyan (approved):
LGTM! Just a few comments on the tests that might use some upgrades.

yanirs:
We may also want to give experimenters an optional to "disable" the experiment to the completed status. I was thinking of it as a dropdown in the dialog. We can leave it out for now, though, and only implement it if it's requested later. :slightly_smiling_face:

__tests__/api/ExperimentsApi.test.ts
jessie-ross: > Do these tests run independently? If not, then this might be trying to change the status from disabled to running because of the previous test setting the experiment status to disabled.

These only reach our Swagger Mock Server which doesn't do anything with the request so this should be no problem :)


> And do we want some expect() here? Maybe checking for 200 response?

Ah that is a good point, ideally our `fetchApi` abstraction should be throwing an error if the status isn't `200`. The core issue here is #300, as our abstraction doesn't output the status. It is a fairly small job to replace it and I might do it during the unplanned week.

Since this is just hitting swagger checking the return doesn't mean all that much... I might mock the `fetch` function and snapshot the calls made to it.

jessie-ross: Actually it was better for now to have `fetchApi` throw an error on non-2XX responses.

PR #306, opened 2020-09-04
Add Wizard Edit Link

jessie-ross:
**This PR is split from #304, it is just the link and cleaning up the tabs:**
- Adds `/experiment/[id]/wizard-edit` page, I am not using `edit` so it won't clash with any future use (inline editing).
- Simplified the Tab code while I was adding the button link in.

## How has this been tested?


yanirs (approved):
:ship:

PR #308, opened 2020-09-07
Add assignment bookmarklet

### LINKED ISSUE #285: Assignment to Variations as a user
1. Experimenters expect to be able to be able to easily switch between experiment variations while testing.
2. Developers expect to be able to easily switch between variations while developing.
3. Experimenters and developers are usually only interfacing with a single experiment at a time.
4. Many times, you're testing with an incognito window.

It turns out that bookmarklets are pretty powerful and easy to use, so this is what I propose.

1. Create buttons you can drag to your bookmark bar.
2. The buttons just call the API to set your variation, and post an alert telling you the variation.
3. I created a basic proof of concept here: https://withinboredom.info/test-love.htm (you can drag the anchor tag to your bookmarks)

withinboredom:
closes #285 

This makes the variations into links that can be dragged to your bookmarks along with a helpful title. It uses an `alert()` to notify the user that the variation was set successfully.

## How has this been tested?

Snapshots and manual testing in browser.

### DIFF
--- a/components/VariationsTable.tsx
+++ b/components/VariationsTable.tsx
@@ -5,7 +5,6 @@ import TableBody from '@material-ui/core/TableBody'
 import TableCell from '@material-ui/core/TableCell'
 import TableHead from '@material-ui/core/TableHead'
 import TableRow from '@material-ui/core/TableRow'
-import { withStyles } from '@material-ui/styles'
 import React from 'react'
 
 import Label from '@/components/Label'
@@ -19,6 +18,13 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     variation: {
       borderBottom: '1px dashed #a3a3a3',
+    },
+    tooltip: {
+      maxWidth: '500px',
+      padding: theme.spacing(2),
+      '& a:link': {
+        color: '#cee6f8',
+      },
     },
   }),
 )
@@ -49,17 +55,6 @@ function dangerousAssignmentLink(variationName: string, experimentName: string, 
   }
 }
 
-const HtmlTooltip = withStyles((theme) => ({
-  tooltip: {
-    maxWidth: '500px',
-    border: '1px solid #dadde9',
-    padding: theme.spacing(2),
-    '& a:link': {
-      color: '#cee6f8',
-    },
-  },
-}))(Tooltip)
-
 /**
  * Renders the variations in tabular formation, in the order that they're given.
  *
@@ -88,10 +83,13 @@ function VariationsTable({
           return (
             <TableRow key={variation.variationId}>
               <TableCell>
-                <HtmlTooltip
+                <Tooltip
                   interactive
                   arrow
                   placement={'left'}
+                  classes={{
+                    popper: classes.tooltip,
+                  }}
                   title={
                     <>
                       <Typography color='inherit' variant='body1' gutterBottom>
@@ -110,7 +108,7 @@ function VariationsTable({
                   }
                 >
                   <span className={classes.variation}>{variation.name}</span>
-                </HtmlTooltip>
+                </Tooltip>
                 {variation.isDefault && <Label className={classes.default} text='Default' />}
               </TableCell>
               <TableCell>{variation.allocatedPercentage}%</TableCell>
--- a/components/VariationsTable.tsx
+++ b/components/VariationsTable.tsx
@@ -7,7 +7,8 @@ import TableRow from '@material-ui/core/TableRow'
 import React from 'react'
 
 import Label from '@/components/Label'
-import { Variation } from '@/lib/schemas'
+import { ExperimentFull } from '@/lib/schemas'
+import * as Variations from '@/lib/variations'
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
@@ -31,13 +32,9 @@ function assignmentHref(variationName: string, experimentName: string, experimen
  * @param props.variations - The variations to render.
  */
 function VariationsTable({
-  variations,
-  experimentName,
-  experimentPlatform,
+  experiment: { variations, name: experimentName, platform: experimentPlatform },
 }: {
-  variations: Variation[]
-  experimentName: string
-  experimentPlatform: string
+  experiment: ExperimentFull
 }) {
   const classes = useStyles()
   return (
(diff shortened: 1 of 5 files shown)


jessie-ross:
Wow looks good! (I'm surprised our API doesn't have CSP enabled.)

I think some small changes would make this a bit more intuitive to use.
- Maybe show the bookmarklet in a popover with a bit of explanation instead of being just a link on variation. [See the MUI Tooltip doccos](https://material-ui.com/components/tooltips/#customized-tooltips)
- Have the name of the bookmark include the experiment so when an Experimenter drags it to their bookmarks bar it will make a bit more sense.
  `Assignment Bookmarklet (Drag to your bookmark bar): <a href="---">[variationName] - [experimentName]</a>`

withinboredom:
I'm not sure CSP would be helpful since we have millions of domains our API might be called from ;) There are policies client-side, however. We should probably implement some of those in Abacus, now that you mention it.

I like the idea of a tooltip, especially because I was trying to figure out how to name it something useful without making the table look bad!

components/VariationsTable.tsx
```diff
@@ -37,11 +76,32 @@ function VariationsTable({ variations }: { variations: Variation[] }) {
...
+                  arrow
+                  placement={'left'}
+                  title={
+                    <>
+                      <Typography color='inherit'>
+                        Drag this link to your bookmarks to make it easier to switch between active variations:
+                      </Typography>
+                      <Typography
+                        color='inherit'
+                        dangerouslySetInnerHTML={dangerousAssignmentLink(
```
withinboredom, line 94: I got a warning that `javascript:` hrefs will be disabled in a future react version, so using `dangerouslySetInnerHTML` instead.

jessie-ross: That reminds me, this does put us into XXS danger territory... It might be worth validating the names one last time before interpolating them in the link?:

```
import { nameSchema } from '@/lib/schemas'

nameSchema.validateSync(experimentName)
nameSchema.validateSync(variationName)
```

components/VariationsTable.tsx
```diff
@@ -1,28 +1,67 @@
...
+      experimentPlatform,
+    )}">${variationName} - ${experimentName}</a>`,
+  }
+}
+
+const HtmlTooltip = withStyles((theme) => ({
+  tooltip: {
+    backgroundColor: '#f5f5f9',
+    color: 'rgba(0, 0, 0, 0.87)',
+    maxWidth: '10dw',
```
jessie-ross, line 49: Haha?? What is a dw?

components/VariationsTable.tsx
```diff
@@ -1,28 +1,67 @@
...
+    fontSize: theme.typography.pxToRem(12),
```
jessie-ross, line 50: Ah I didn't know about this. Btw if you use `variant="body1"` it should set font-size for you and give you some other nice defaults.

components/VariationsTable.tsx
```diff
@@ -37,11 +76,32 @@ function VariationsTable({ variations }: { variations: Variation[] }) {
...
             <TableRow key={variation.variationId}>
               <TableCell>
-                {variation.name}
+                <HtmlTooltip
+                  interactive
+                  arrow
+                  placement={'left'}
+                  title={
+                    <>
+                      <Typography color='inherit'>
```
jessie-ross, line 89: I think adding a `gutterBottom` here might be nice.

components/VariationsTable.tsx
```diff
@@ -1,28 +1,67 @@
...
+import { ExperimentFull } from '@/lib/schemas'
+import * as Variations from '@/lib/variations'
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     default: {
       marginLeft: theme.spacing(1),
     },
+    variation: {
+      borderBottom: '1px dashed #a3a3a3',
```
jessie-ross, line 21: I like this a lot!

components/VariationsTable.tsx
```diff
@@ -1,28 +1,67 @@
...
+      variationName,
+      experimentName,
+      experimentPlatform,
+    )}">${variationName} - ${experimentName}</a>`,
+  }
+}
+
+const HtmlTooltip = withStyles((theme) => ({
+  tooltip: {
+    backgroundColor: '#f5f5f9',
```
jessie-ross, line 47: I think the contrasting default dark background might be useful here:

Screen Shot 2020-09-08 at 7 41 52 pm


As the this one seems to match the background colour and can be a bit visually confusing:

Screen Shot 2020-09-08 at 7 43 08 pm

Also `padding: theme.spacing(2)` is a loyal friend 😉

components/VariationsTable.tsx
```diff
@@ -1,28 +1,67 @@
...
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    __html: `<a href="${assignmentHref(
+      variationName,
+      experimentName,
+      experimentPlatform,
+    )}">${variationName} - ${experimentName}</a>`,
+  }
+}
+
+const HtmlTooltip = withStyles((theme) => ({
```
jessie-ross, line 45: Ah also we use the newer `useStyles` API instead of the HOC API, see Line 15.

components/VariationsTable.tsx
withinboredom: I think I was going more for `vw` but for some reason I wrote `dw`, probably thinking "device width" :joy:

components/VariationsTable.tsx
withinboredom: Maybe? Don't these names come from the API?

I think escaping them might be all that's needed though, in the event the name/variation schema changes to something more permissive.

components/VariationsTable.tsx
withinboredom: I couldn't figure out how to make that work. It seems to butcher the existing class instead of mixing it together.

components/VariationsTable.tsx
jessie-ross: Nice, yeah I think defence in depth is generally worth it 😄

jessie-ross (approved):
LGTM 🚢

PR #309, opened 2020-09-07
Add button state based on experiment status

jessie-ross:
**This PR adds disable state and tooltips to experiment CRUD buttons.**
- Tooltips can probably be improved
- Note sure if I got all the states correct particularly assign metric and conclude.
- For each status we test each button at the same time so we can clearly see what is enabled and what isn't.

## How has this been tested?
- Automated tests that cover added functionality


components/ExperimentPageView.tsx
```diff
@@ -1,5 +1,5 @@
 // istanbul ignore file; Even though it sits with components this is a "page" component
```
yanirs, line 1: Should we remove this now that it's tested?

components/ExperimentPageView.tsx
```diff
@@ -129,18 +137,27 @@ export default function ExperimentPageView({
...
-              component={NextMuiLink}
-              href={`/experiments/[id]/wizard-edit`}
-              hrefAs={`/experiments/${experimentId}/wizard-edit`}
-            >
-              Edit In Wizard
-            </Button>{' '}
-            <Button variant='outlined' color='secondary'>
-              Run
-            </Button>{' '}
+            <Tooltip title={canEditInWizard ? '' : 'Only available while staging.'}>
```
yanirs, line 140: `Only available for staging experiments` sounds a bit better to me.

components/ExperimentPageView.tsx
```diff
@@ -98,6 +98,14 @@ export default function ExperimentPageView({
 
   const isLoading = or(experimentIsLoading, metricsIsLoading, segmentsIsLoading, analysesIsLoading)
 
+  const canEditInWizard = experiment && experiment.status === Status.Staging
+  const canRunExperiment = experiment && experiment.status === Status.Staging
+  const cantRunExperimentReason: Record<string, string> = {
+    [Status.Running]: 'This experiment is running.',
+    [Status.Completed]: 'This experiment has ran.',
+    [Status.Disabled]: 'This experiment is disabled.',
+  }
```
yanirs, line 107: `This experiment has ran` is also possibly true for disabled experiments. :slightly_smiling_face: 

It'd be simpler to change this to `This experiment is <status>`, which makes the mapping unnecessary.

components/ExperimentPageView.tsx
```diff
@@ -129,18 +137,27 @@ export default function ExperimentPageView({
...
+                  component={NextMuiLink}
+                  href={`/experiments/[id]/wizard-edit`}
+                  hrefAs={`/experiments/${experimentId}/wizard-edit`}
+                  disabled={!canEditInWizard}
+                >
+                  Edit In Wizard
+                </Button>
+              </span>
+            </Tooltip>{' '}
+            <Tooltip title={!canRunExperiment ? cantRunExperimentReason[experiment?.status as string] : ''}>
```
yanirs, line 154: Please reverse the order of the ternary to avoid the double negative and make it consistent with the edit button. Can also inline the reason and turn this into:
```
canRunExperiment ? '' : `This experiment is ${experiment?.status as string}.`
```

By the way, why do we render the top bar when the experiment can be undefined?

components/GeneralPanel.tsx
```diff
@@ -102,6 +103,7 @@ function GeneralPanel({
     // Needed for endDatetime validation
     startDatetime: experiment.startDatetime,
   }
+  const canEditExperiment = experiment.status !== Status.Staging
```
yanirs, line 106: This variable name is a bit confusing because we _can_ edit the experiment. Maybe change to `canEditPanel` or `canEdit`?

components/MetricAssignmentsPanel.tsx
```diff
@@ -123,6 +124,11 @@ function MetricAssignmentsPanel({
 
   // Assign Metric Modal
   const { enqueueSnackbar } = useSnackbar()
+  const canAssignMetric = experiment.status === Status.Running
+  const cantAssignReasons: Record<string, string> = {
+    [Status.Staging]: 'Use "Edit in Wizard" for staging experiments.',
+    [Status.Disabled]: `This experiment is disabled.`,
+  }
```
yanirs, line 131: This is incorrect. Metrics can always be added for non-staging experiments.

components/ExperimentPageView.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+const mockedAnalysesApi = AnalysesApi as jest.Mocked<typeof AnalysesApi>
+
+jest.mock('notistack')
+const mockedNotistack = notistack as jest.Mocked<typeof notistack>
+mockedNotistack.useSnackbar.mockImplementation(() => ({
+  enqueueSnackbar: jest.fn(),
+  closeSnackbar: jest.fn(),
+}))
+
+test('staging experiment shows correct features enabled in overview', async () => {
```
yanirs, line 39: It's great to have more tests, but the duplication in this file makes it hard to see what gets tested and why. Can you please add a couple of local helpers to reduce duplication and make it clearer? Helpers that may work well are `renderExperimentPageView(status)` to replace lines 40-56 and their duplicates, and `checkButtonStates(expectedStates)` to replace lines 59-71 and their duplicates (where `expectedStates` is a mapping of the form `{editInWizard: true, run: false, conclusionEdit: null, ...}`).

I also think we can do without the snapshots, as the individual panels are already covered by snapshot tests.

yanirs (changes requested):
Good progress. Should be good to go with the inline comments, which are all pretty straightforward.

components/GeneralPanel.tsx
jessie-ross: :+1: Changing to `canEditGeneralPanel`

components/ExperimentPageView.tsx
jessie-ross: > By the way, why do we render the top bar when the experiment can be undefined?

That's a good point, I think ideally we would have it at the top but since we reload the data on every tab at the moment due to the router (plugging #212 here) it doesn't give a good UX for the tabs: https://d.pr/v/uW3T9K

I did just carry it forward from the original implementation but I have a feeling that this was the reasoning.

It is a pretty trivial change, easy to do now or later.

components/ExperimentPageView.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+  mockedMetricsApi.findAll.mockImplementationOnce(async () => metrics)
+
+  const segments = Fixtures.createSegments(10)
+  mockedSegmentsApi.findAll.mockImplementationOnce(async () => segments)
+
+  // @ts-ignore
+  render(<ExperimentPageView experimentId={null} view={ExperimentView.Overview} debugMode={false} />)
+})
+
+test('staging experiment shows correct features enabled in overview', async () => {
```
yanirs, line 98: So much clearer with the helper functions! :tada:

components/ExperimentPageView.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+    run: !run.hasAttribute('disabled'),
+    disable: !disable.hasAttribute('disabled'),
+    generalPanelEdit: !generalPanelEdit.hasAttribute('disabled'),
+    assignMetric: !assignMetric.hasAttribute('disabled'),
+    conclusionEdit: conclusionEdit && !conclusionEdit.hasAttribute('disabled'),
+  }
+}
+
+test('experiment page view renders results without crashing', async () => {
+  const { container: _container } = renderExperimentPageView({}, ExperimentView.Results)
```
yanirs, line 80: Here and on other lines, what's the idea behind assigning the result to an unused variable?

```suggestion
  renderExperimentPageView({}, ExperimentView.Results)
```

yanirs (approved):
See inline comment about unused variables, but other than that LGTM. :ship:

components/ExperimentPageView.test.tsx
jessie-ross: Oh just because I copy and pasted it... I will be more careful in the future but will just leave these as they are harmless.

PR #310, opened 2020-09-08
Eliminate console warnings when running tests

### LINKED ISSUE #229: Eliminate noise when running unit tests
When running `npm run test:unit`, we get some noisy output from the console and other warnings. It shouldn't happen on a clean run.

yanirs:
Fix #229 by making tests fail if anything is logged to the console and fixing all the existing console warnings.

## How has this been tested?

Ran `npm run verify` before and after adding `jest-setup.ts` and fixing existing warnings, and verified that all the console warnings are gone and that the existing warnings fail the test with the new ``jest-setup.ts`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
                 command: npm run lint
             - run:
                 name: Unit Testing
-                command: npm run test:unit -- -- --ci --maxWorkers 2
+                command: npm run test:unit -- -- --ci --runInBand
             - run:
                 name: Integration Testing
                 command: npm run test:integration -- --ci
--- a/components/GeneralPanel.test.tsx
+++ b/components/GeneralPanel.test.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable @typescript-eslint/require-await */
-import { act, fireEvent, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react'
+import { fireEvent, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react'
 import { noop } from 'lodash'
 import MockDate from 'mockdate'
 import * as notistack from 'notistack'
@@ -198,6 +197,7 @@ test('opens, submits and cancels edit dialog with running experiment', async () 
   mockedExperimentsApi.patch.mockReset()
   // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
   // @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/require-await
   mockedExperimentsApi.patch.mockImplementationOnce(async () => experiment)
 
   const editButton = screen.getByRole('button', { name: /Edit/ })
@@ -205,12 +205,10 @@ test('opens, submits and cancels edit dialog with running experiment', async () 
 
   await waitFor(() => screen.getByRole('button', { name: /Save/ }))
 
-  changeFieldByRole('textbox', /Experiment description/, 'Edited description.')
+  await changeFieldByRole('textbox', /Experiment description/, 'Edited description.')
   // This date was picked as it is after the fixture start date.
-  await act(async () => {
-    fireEvent.change(screen.getByLabelText(/End date/), { target: { value: '2020-10-20' } })
-  })
-  changeFieldByRole('textbox', /Owner/, 'changed_test_a11n')
+  fireEvent.change(screen.getByLabelText(/End date/), { target: { value: '2020-10-20' } })
+  await changeFieldByRole('textbox', /Owner/, 'changed_test_a11n')
 
   const saveButton = screen.getByRole('button', { name: /Save/ })
   fireEvent.click(saveButton)
@@ -256,6 +254,7 @@ test('checks edit dialog does not allow end datetime changes with disabled exper
   mockedExperimentsApi.patch.mockReset()
   // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
   // @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/require-await
   mockedExperimentsApi.patch.mockImplementationOnce(async () => experiment)
 
   const editButton = screen.getByRole('button', { name: /Edit/ })
@@ -263,9 +262,9 @@ test('checks edit dialog does not allow end datetime changes with disabled exper
 
   await waitFor(() => screen.getByRole('button', { name: /Save/ }))
 
-  changeFieldByRole('textbox', /Experiment description/, 'Edited description.')
+  await changeFieldByRole('textbox', /Experiment description/, 'Edited description.')
   expect(screen.getByLabelText(/End date/)).toBeDisabled()
-  changeFieldByRole('textbox', /Owner/, 'changed_test_a11n')
+  await changeFieldByRole('textbox', /Owner/, 'changed_test_a11n')
 
   const saveButton = screen.getByRole('button', { name: /Save/ })
   fireEvent.click(saveButton)
--- a/components/experiment-creation/BasicInfo.test.tsx
+++ b/components/experiment-creation/BasicInfo.test.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable no-irregular-whitespace */
-import { act, fireEvent, getByLabelText, render } from '@testing-library/react'
+import { fireEvent, getByLabelText, render } from '@testing-library/react'
 import MockDate from 'mockdate'
 import React from 'react'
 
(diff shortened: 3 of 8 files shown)


components/GeneralPanel.test.tsx
```diff
@@ -198,19 +197,18 @@ test('opens, submits and cancels edit dialog with running experiment', async ()
...
+  // eslint-disable-next-line @typescript-eslint/require-await
   mockedExperimentsApi.patch.mockImplementationOnce(async () => experiment)
 
   const editButton = screen.getByRole('button', { name: /Edit/ })
   fireEvent.click(editButton)
 
   await waitFor(() => screen.getByRole('button', { name: /Save/ }))
 
-  changeFieldByRole('textbox', /Experiment description/, 'Edited description.')
+  await changeFieldByRole('textbox', /Experiment description/, 'Edited description.')
```
yanirs, line 208: These extra `await`s fixed the overlapping `act` warnings. Not sure why the `act()` was needed.

components/experiment-creation/BasicInfo.test.tsx
```diff
@@ -21,52 +20,42 @@ test('renders as expected', () => {
 test('renders sensible dates as expected', () => {
   MockDate.set('2020-07-21')
   const { container } = render(
-    <MockFormik>
+    <MockFormik initialValues={{ experiment: { startDatetime: '', endDatetime: '' } }}>
```
yanirs, line 23: Setting initial values fixed the complaints about changing uncontrolled fields. Looks like the `act()` calls weren't necessary.

test-helpers/jest-setup.ts
```diff
@@ -0,0 +1,10 @@
+import { format } from 'util'
+
+// Throw errors on any console logging.
+// Adapted from https://github.com/facebook/jest/issues/6121#issuecomment-529591574
+;['error', 'info', 'log', 'warn'].forEach((funcName) => {
```
yanirs, line 5: Not sure why Prettier added these leading semicolons...

test-helpers/jest-setup.ts
```diff
@@ -0,0 +1,10 @@
+import { format } from 'util'
```
yanirs, line 1: We can use this file for other common setup tasks, like mocks that are used in many tests (e.g., `notistack`). We may want to use different files for different test suites, though.

.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
                 command: npm run lint
             - run:
                 name: Unit Testing
-                command: npm run test:unit -- -- --ci --maxWorkers 2
+                command: npm run test:unit -- -- --ci --runInBand
```
yanirs, line 42: Changed in an attempt to reduce flakiness after unrelated tests timed out on CI three times in a row. Not sure if it helped, but it passed. :man_shrugging: Should explore speeding up the tests more separately from this change.

components/experiment-creation/BasicInfo.test.tsx
withinboredom: `act()` behaves quite differently than without.

Here's the rendering with your changes:

```
rendered
start - line 29
rendered
rendered
rendered
rendered
rendered
end
```

and with `act()`

```
rendered
start - line 29
rendered
rendered
rendered
rendered
end
```

You can see that with `act()` we render one less time than without. `act()` basically suspends rendering, while without `act()` we render after each and every change to the component. In this test's particular case, I think it's more accurate around line 30/31 because a user isn't going to change the date in two different places before another rendering of the component.

test-helpers/jest-setup.ts
withinboredom: Because js is supposed to have semi-colons, though it's optional for some types of statements. After the comments are stripped by the parser, the lexer needs to know that the `[` token isn't part of the `import` token.

test-helpers/test-utils.tsx
```diff
@@ -46,9 +46,15 @@ export function createMatchMedia(width: number) {
...
+export const MockFormik = ({
+  children,
+  initialValues,
+}: {
+  children: React.ReactNode
+  initialValues?: FormikValues
+}) => {
   return (
-    <Formik initialValues={{}} onSubmit={() => undefined}>
+    <Formik initialValues={initialValues ?? {}} onSubmit={() => undefined}>
```
withinboredom, line 57: Since initialValues isn't required above, it might make sense to just have a default value be `{}` and then just pass `initialValues` into the component instead of null coalescing.

test-helpers/jest-setup.ts
```diff
@@ -0,0 +1,10 @@
...
+
+// Throw errors on any console logging.
+// Adapted from https://github.com/facebook/jest/issues/6121#issuecomment-529591574
+;['error', 'info', 'log', 'warn'].forEach((funcName) => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ;(global.console as any)[funcName] = function (message?: any, ...optionalParams: any[]) {
+    throw new Error(format(message, optionalParams))
```
withinboredom, line 8: This is a terrible idea btw. I had to jump through a ton of hoops (ended up using `process.stdout.write`) to show how `act()` worked above. Generally, with debugging, we want to be able to output stuff. Remembering to turn this off/on is also just asking for it to accidentally be removed.

Also, we probably do want specific console errors or logging to be in production for certain kinds of things.

test-helpers/jest-setup.ts
withinboredom: It's generally a good practice to keep the file global for all tests and not have specific files for certain tests. Eventually, you don't want to touch this file unless it is absolutely necessary, because if you do have to touch it, you're probably doing it wrong. :D

test-helpers/jest-setup.ts
jessie-ross: Yep this is a tradeoff of not using semi-colons - we have to rely on our tooling to recognise special cases where there needs to be a semi-colon. 

Another tradeoff is that this can hide accidental line-split mistakes where a line we expect to be a single line becomes two lines with unexpected behaviour, granted this is less likely with eslint/prettier but with the amount of complexity in implementing safe-guards for it I am sure there are some exceptions. [In fact this is likely why these semi-colons are in place](https://www.freecodecamp.org/news/codebyte-why-are-explicit-semicolons-important-in-javascript-49550bea0b82/),

More generally I have been thinking of suggesting moving towards A8Cs JS code style for making it just that bit smoother for other A12s dig in and aid future maintenance. Thanks to eslint/prettier it should be an automated process.

test-helpers/jest-setup.ts
jessie-ross: I agree, this would be infuriating in practice for debugging/needing production logging.

Additionally this elevates any console log/warn/debug/info/error to a blocking error in our tests - I thought we had agreed that we could treat console errors as non-critical and deal with them in batches when appropriate.
That is what I got from our conversation at: https://github.com/Automattic/abacus/issues/229#issuecomment-663361149

components/experiment-creation/BasicInfo.test.tsx
jessie-ross: Thanks for fixing these @yanirs.

I haven't been worrying too much about `act` but since I have wrote this code I have found that `fireEvent` and `render` comes wrapped in `act`, so these do double up and line 30/31 is as you've said Rob.

withinboredom: For test performance reasons, I think using `act()` makes a ton of sense, since we won't have to render every change (thereby speeding up the tests). It probably depends, largely, on what we're changing between assertions. A render *shouldn't* have side-effects, but with functional components and react-hooks, they actually might. So, like I said, it just depends.

components/GeneralPanel.test.tsx
jessie-ross: Ahh yeah, this makes sense as changeFieldByRole is an async function (in certain situations it is necessary to `await` an `act` which is why I made it async.)

Since it is async the `fireEvent.change` below can open up an `act` before `changeFieldByRole`s `act` has closed.

components/experiment-creation/BasicInfo.test.tsx
jessie-ross: Yeah that is a good point about performance I didn't think about that.

.circleci/config.yml
jessie-ross: Hmm I am not so sure about this... We are speeding up single tests in order to prevent timeout but this isn't a sustainable way to do it - if tests take longer this won't work anymore.

I am more in favour of just [increasing the default timeout](https://jestjs.io/docs/en/configuration#testtimeout-number), we could bump it to `20000` and it shouldn't be a problem - I believe the main reason for timeouts are to detect non-termination early, but tests take as long as they take.

yanirs: Hmm... For some reason I remembered that our CircleCI setup is a single CPU container, so I assumed that dropping the parallel run would fix things. Now I see that it's two CPUs. I'll increase the default timeout instead and see how it goes.

test-helpers/jest-setup.ts
yanirs: We'll see. We already have different configs, e.g., for e2e and unit tests, and I don't think we want to do a lot of global mocking outside unit tests.

test-helpers/jest-setup.ts
yanirs: The problem is that the _same_ warnings (that we know how to handle) keep popping up again and again, making the test console output noisy and essentially useless. It's also pretty annoying to fix a few warnings only to come back to the code a week or two later and find new warnings.

The points about allowing debugging _while writing tests_ and deferring hard-to-handle warnings make sense, though. What do you think of throwing errors on specific messages instead, like the overlapping act? Any other console output will be retained.

components/experiment-creation/BasicInfo.test.tsx
yanirs: Yeah, I don't mind keeping the `act()` around multiple changes for performance reasons. I removed them since they weren't necessary.

test-helpers/jest-setup.ts
jessie-ross: Let's attach this here: p1599636897000600-slack-G017FFWSQ

yanirs:
Closing &ndash; to be replaced with separate PRs.

PR #311, opened 2020-09-08
Implement Run Experiment

jessie-ross:
**This PR implements Run experiment.**
Vid: https://d.pr/v/UZXYMR

- Identical to the Disable button #303 
- Decided not to refactor it just yet

## How has this been tested?

### DIFF
--- a/components/ExperimentRunButton.tsx
+++ b/components/ExperimentRunButton.tsx
@@ -1,4 +1,4 @@
-import { Button, Dialog, DialogActions, DialogContent, Typography } from '@material-ui/core'
+import { Button, Dialog, DialogActions, DialogContent, Typography, Tooltip } from '@material-ui/core'
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
 import { useSnackbar } from 'notistack'
 import React, { useState } from 'react'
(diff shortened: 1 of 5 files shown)


components/ExperimentRunButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  createStyles({
+    root: {},
+  }),
+)
+
+const ExperimentRunButton = ({
+  experiment,
+  experimentReloadRef,
+}: {
+  experiment: ExperimentFull | null
```
yanirs, line 21: Since this is its own component that doesn't make much sense without an experiment, maybe declare this to be `ExperimentFull` and avoid all the checks below?

jessie-ross: This is due to the buttons being in the tab bar above the loader

I think the UX is better with them starting disabled and showing after an experiment loads.

yanirs: Not sure it's better, but I'm ok with it for now. :slightly_smiling_face:

yanirs (approved):
:ship:

jessie-ross:
Eeep, I didn't merge this into master 🤦‍♀️

PR #314, opened 2020-09-09
Fix general panel UI

jessie-ross:
**This PR improves the general panel UI.**

_After:_

Screen Shot 2020-09-09 at 2 31 15 pm

_Before:_

Screen Shot 2020-09-09 at 2 15 57 pm

**Changes:**

- Makes the "to" in between dates grey.
- Uses the MUI Link component for the P2 link.
- Applies monospace fonts.
- (Controversially?) changes the order of the fields: 
   - Status should be at the top, the most important thing for an Experimenter to know about an experiment.
      Is it running? planning? finished?
      It affects the UI state dramatically so is also important UX for this to be prominent.
   - Dates should come after, it is very important that dates are frequently looked so no surprises are had.
   - P2 Url should come last - I think it fits really well to have a link at the end, the equivalent of having a "view more" link at the bottom.
   

## How has this been tested?


yanirs (approved):
I'm happy with the reordering/restyling, but let's give @aaronyan and @withinboredom time to comment in case they have any objections.

aaronyan:
LGTM to as well!

PR #316, opened 2020-09-09
Fix date formatting

### LINKED ISSUE #307: The time of Start and End datetimes
In testing the "Edit in Wizard" functionality I've realised abacus wasn't setting the time of `startDatetime` and `endDatetime` to `00:00:000Z`. This is a bug and has resulted in some experiments with strange start/end datetimes.

jessie-ross:
**This PR fixes a date bug where dates aren't being validated or coerced as UTC resulting in datetimes with non-`00:00:000Z` times possibly causing wrong dates to be entered:**
- `date-fns` is quite unintuitive when working in UTC:
   https://stackoverflow.com/questions/58561169/date-fns-how-do-i-format-to-utc
- Replaces usages of `date-fns` with native date object methods.
- Use new `formatIsoDate` function instead of `dateFns.format`

Fixed #307 

## How has this been tested?


test-helpers/fixtures.ts
```diff
@@ -260,10 +259,14 @@ function createMetricAssignment(fieldOverrides: Partial<MetricAssignment>): Metr
 /* istanbul ignore next; All coverage to be removed, see https://github.com/Automattic/abacus/issues/231 */
 function createExperimentFullNew(fieldOverrides: Partial<ExperimentFullNew> = {}): ExperimentFullNew {
   const now = new Date()
+  // This is a bit funky as we want to work in UTC but DateFns only does local time
+  // and we want to avoid DST problems.
+  const startDatetime = new Date(new Date(new Date().setMonth(now.getMonth() + 2)).setUTCHours(0))
```
yanirs, line 264: What's the idea behind `.setUTCHours(0)`? If it is to round to midnight UTC, it doesn't work because we still have minutes and seconds.

jessie-ross: Nice catch!

yanirs (approved):

PR #317, opened 2020-09-09
Clean up audience panel UI

jessie-ross:
This PR cleans up the Audience Panel UI.
- Uses monospace
- Visually reduces the "default" label.

After:

Screen Shot 2020-09-10 at 3 49 09 am

Before:

Screen Shot 2020-09-10 at 3 57 03 am

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)

PR #318, opened 2020-09-09
Clean up metric assignments panel UI

jessie-ross:
**This PR improves the MetricAssignments Panel UI.**
- Uses monospace.
- Visually reduces the default text.

After:

Screen Shot 2020-09-10 at 4 02 57 am

Before:

Screen Shot 2020-09-10 at 4 05 04 am

## How has this been tested?


aaronyan:
Thanks for the clean up! LGTM. I would prefer adjusting the [Chip](https://material-ui.com/components/chips/) to use the **outlined** **small** variation like this! It will have a smaller presence, but keeps the Chip design pattern of an "attribute".

Screen Shot 2020-09-09 at 4 17 37 PM

jessie-ross:
## Update
- Plugged the chip in like in #317 

Screen Shot 2020-09-12 at 2 46 15 am

aaronyan (approved):
Looking great! Nice addition of the disabled version of the chip, it looks clean and out of the way.

PR #319, opened 2020-09-10
Filter out test data

### LINKED ISSUE #320: Filter from old test experiments from Abacus
With the soft launch of Abacus approaching, we should clean up our /experiments view to only show real experiments. :)

jessie-ross:
**This PR filters out test data.**
- Filters out experiments and metrics starting with `explat_test_`

Resolves #320 

## How has this been tested?

- Manual testing with production data (locally):
  - Visit `/experiments` with and without `?debug=true`, notice the filtering of `explat_test_`
  - Visit `/metrics` with and without `?debug=true`, notice the filtering of `explat_test_`
  - Visit `/experiments/new` with and without `?debug=true`, notice the filtering of `explat_test_` of metric options
  - Visit `/experiments/20015/wizard-edit` with and without `?debug=true`, notice the filtering of `explat_test_`

### DIFF
--- a/components/ExperimentPageView.tsx
+++ b/components/ExperimentPageView.tsx
@@ -26,7 +26,7 @@ import ExperimentDisableButton from '@/components/ExperimentDisableButton'
 import Layout from '@/components/Layout'
 import { Analysis, ExperimentFull, Status } from '@/lib/schemas'
 import { useDataLoadingError, useDataSource } from '@/utils/data-loading'
-import { createUnresolvingPromise, or } from '@/utils/general'
+import { createUnresolvingPromise, or, testDataNamePrefix } from '@/utils/general'
 
 import ExperimentRunButton from './ExperimentRunButton'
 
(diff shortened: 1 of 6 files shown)


aaronyan:
Resolves #320

yanirs:
@aaronyan adding a comment doesn't activate the automatic resolution. It needs to be added to the description, which just did. :slightly_smiling_face:

jessie-ross:
## Update

This is ready for review, it requires testing locally with production data (see the readme). I have metrics and experiments starting with `explat_test_` in there already.

aaronyan:
Tested locally using `NEXT_PUBLIC_PRODUCTION_CONFIG_IN_DEVELOPMENT=true npm run dev` and it works as expected! Experiments and metrics prefixed with `explat_test_` were filtered out from view.

yanirs (changes requested):
I find this a bit too repetitive, which makes it error-prone.

How about instead of `testDataNamePrefix`, we export a function like this one:
```typescript
function filterTestData<T extends {name: string}>(records: Array<T>, debugMode: boolean) {
  return debugMode ? records.filter((record) => !record.name.startsWith('explat_test_')) : records
}
```

This function should be documented and tested in isolation. It can be used everywhere, making the code clearer, self-documenting, and consistent.

jessie-ross:
Ah, so I was talking with Rob yesterday and he suggested it might be better if we filter on the backend based on whether the user is an `explat_dev`, I think that would make things a bit simpler.

yanirs:
That's a good idea. It'd definitely be cleaner. :slightly_smiling_face: 

I'll close this PR and move issue #320 to the backend.

PR #330, opened 2020-09-18
Add schema for autocomplete

withinboredom:
Adding the schema for the autocomplete endpoints.

## How has this been tested?
`npm run test`

### DIFF
--- a/lib/schemas.ts
+++ b/lib/schemas.ts
@@ -378,6 +378,36 @@ export const analysisSchema = yup
   .camelCase()
 export type Analysis = yup.InferType<typeof analysisSchema>
 
+export const autocompleteItemSchema = yup.object({
+  name: yup.string(),
+  value: yup.string().nullable().notRequired(),
+})
+export type AutocompleteItem = yup.InferType<typeof autocompleteItemSchema>
+
+export const autocompleteSchema = yup.object({
+  completions: yup.array<AutocompleteItem>(autocompleteItemSchema),
+})
+export type UserCompletions = yup.InferType<typeof autocompleteSchema>
+export type EventCompletions = yup.InferType<typeof autocompleteSchema>
+export type TransactionTypeCompletions = yup.InferType<typeof autocompleteSchema>
+export type ProductCompletions = yup.InferType<typeof autocompleteSchema>
+
+export const eventPropsSchema = yup.object({
+  name: yup.string(),
+  description: yup.string(),
+})
+export type EventProp = yup.InferType<typeof eventPropsSchema>
+
+export const eventDetailsSchema = yup.object({
+  name: yup.string(),
+  description: yup.string(),
+  owner: yup.string(),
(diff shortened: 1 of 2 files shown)


lib/schemas.ts
```diff
@@ -351,6 +351,36 @@ export const analysisSchema = yup
   .camelCase()
 export type Analysis = yup.InferType<typeof analysisSchema>
 
+export const autocompleteItemSchema = yup.object({
+  name: yup.string(),
+  value: yup.string().nullable().notRequired(),
```
jessie-ross, line 356: If it is not too much effort at this point: For consistency it might be good to always have a `value` and just use the `name` as the `value` in those cases.

jessie-ross: It would also mean less type-guards against missing values later on in the code.

lib/schemas.ts
```diff
@@ -351,6 +351,36 @@ export const analysisSchema = yup
...
+})
+export type AutocompleteItem = yup.InferType<typeof autocompleteItemSchema>
+
+export const autocompleteSchema = yup.object({
+  completions: yup.array<AutocompleteItem>(autocompleteItemSchema),
+})
+export type UserCompletions = yup.InferType<typeof autocompleteSchema>
+export type EventCompletions = yup.InferType<typeof autocompleteSchema>
+export type TransactionTypeCompletions = yup.InferType<typeof autocompleteSchema>
+export type ProductCompletions = yup.InferType<typeof autocompleteSchema>
```
jessie-ross, line 366: What is the reasoning behind having seperate types for each completion?

lib/schemas.ts
withinboredom: Good point. I'll update the API.

lib/schemas.ts
withinboredom: Each type represents a single endpoint, which may or may not change independently. So, it's a good separation of concerns to make them independent types. ie, it's much easier to search everywhere for `ProductCompletions` vs. `Autocomplete` and trying to deduce which endpoint it's referring to when/if we want to change something or add metadata. It's even better to use a base type so we can cast to `Autocomplete` or something for generalization reasons. I'll be figuring that out today.

lib/schemas.ts
```diff
@@ -351,6 +351,52 @@ export const analysisSchema = yup
...
+  .object({
+    name: yup.string().defined(),
+    description: yup.string().defined(),
+    owner: yup.string().defined(),
+    is_registered: yup.boolean().defined(),
+    is_validated: yup.boolean().defined(),
+    props: yup.array<EventProp>(eventPropsSchema).defined(),
+  })
+  .defined()
+export type EventDetails = yup.InferType<typeof eventDetailsSchema>
```
jessie-ross, line 398: This looks like nice data to have :)

jessie-ross (approved):
LGTM!

PR #334, opened 2020-09-22
Add basic result graphs

jessie-ross:
**This PR adds basic graphs to our Result view.**
[Image Deleted - see the slack images]

Future Planned PRs:
- Add legend
- Add brighter colors for "winning" variants
- Add a tint to the graph for inconclusive/needs manual analysis

## How has this been tested?


test-helpers/unit-test-setup.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable @typescript-eslint/ban-ts-ignore */
+// @ts-ignore
+window.URL.createObjectURL = jest.fn()
```
jessie-ross, line 3: This file is not really necessary as I am mocking `plotly.js` now but I thought I may as well leave it in.

jessie-ross:
We have another SSR related pain:

Screen Shot 2020-09-22 at 5 41 11 pm

Linking to #212

yanirs:
Nice! I especially like how the plots fit in with the detail panel colour scheme. It's looking way better than it did before the restyling. :slightly_smiling_face: 

Some thoughts on the current PR:

* We should use lines rather than area graphs for the variation intervals. I find it way too hard to parse and understand the overlapping colours, even when hovering to see the data points. See related [Material Visualisation guideline](https://material.io/design/communication/data-visualization.html): _"Don’t use overlapped area charts as it obscures data values and reduces readability."_
* We should multiply all the conversion values by 100 rather than rely on the x100 label.
* We can make the graphs less busy by dropping the y-axis labels and changing the titles:
  * `Variation Credible Intervals` + `Metric Estimates (x100%)` &rarr; `Conversion rate estimates by variation [%]`
  * `Difference Credible Interval` + `Difference Estimates (x100%)` &rarr; `Conversation rate difference estimates [%]`
  * `Variation Credible Intervals` + `Metric Estimates ($)` &rarr; `Revenue estimates by variation [$]`
  * `Difference Credible Interval` + `Difference Estimates ($)` &rarr; `Revenue difference estimates [$]`

Regarding future plans:

* I'm not sure whether a legend would work versus adding annotations on the graph.
* I think we should keep the colour scheme consistent across metrics, i.e., always use the same colours for control/treatment, regardless of the result.

jessie-ross:
> We should multiply all the conversion values by 100 rather than rely on the x100 label.

Ahhh! Yep that would be simplest.

jessie-ross:
Moving this discussion to: pbmo2S-vR-p2#comment-1098

jessie-ross:
## Update

- x100 the conversion estimates
- Fixed the titles
- Use lines instead of areas

components/experiment-results/CondensedLatestAnalyses.tsx
```diff
@@ -179,17 +196,146 @@ const useAnalysisDetailStyles = makeStyles((theme: Theme) =>
...
 }) {
   const classes = useAnalysisDetailStyles()
+  const theme = useTheme()
+
+  // We hand out colors based on the order of the variant.
+  // Control and Treatment should have a consistent order even across experiments so this should give good results.
+  // Ideally we would add some marker to differentiate control variants
+  //
+  // These come from a data pallete generator and should be visually equidistant.
+  const variantColors = ['rgba(71, 27, 92, .3)', 'rgba(205, 112, 85, .3)']
```
withinboredom, line 232: Here's the Looker palettes:


```
Categories: #1f78b4, #ff7f00, #a6cee3, #33a02c, #fb9a99, #e31a1c, #fdbf6f, #cab2d6, #b2df8a
Diverging: #8c510a, #dfc27d, #f5f5f5, #80cdc1, #01665e
```

Dunno if we want to use the same one.

components/experiment-results/CondensedLatestAnalyses.tsx
```diff
@@ -179,17 +196,146 @@ const useAnalysisDetailStyles = makeStyles((theme: Theme) =>
...
+          type: 'scatter' as 'scatter',
+        },
+        {
+          name: `${variation.name}: upper bound`,
+          x: dates,
+          y: analyses
+            .map(({ metricEstimates }) => metricEstimates && metricEstimates[variationKey].top)
+            .map(estimateTransform),
+          // fill: 'tonexty' as 'tonexty',
+          // fillcolor: variantColors[index],
```
withinboredom, line 266: Do we need to keep these comments?

withinboredom (approved):
Looked up the Looker palette for you, but it's not necessary to use it. But we can if we want to be consistent.

components/experiment-results/CondensedLatestAnalyses.tsx
jessie-ross: Yeah I can remove them

components/experiment-results/CondensedLatestAnalyses.tsx
jessie-ross: I'll have a try with them

jessie-ross:
## Update

I've made Rob's changes, feel free to merge once the checks pass - I am heading out for now!

PR #335, opened 2020-09-22
Improve metric assignment sorting

jessie-ross:
**This PR improves the sorting on Metric Assignments**
- Primary first
- Assignments with the same metricId are next to each other
- Assignments with the same metricId are ordered by attributionWindow asc

This makes it easier to compare metric assignments with the same metric.

## How has this been tested?


yanirs (approved):
:ship:

PR #336, opened 2020-09-22
Reduce errors in unit tests

jessie-ross:
**This PR reduces the unit test warnings**
- Fixes the act warnings

There only remains a `forwardRef` warning I haven't been able to figure out in the time-box for this:

Screen Shot 2020-09-23 at 1 11 04 am

Part of a series on #229

## How has this been tested?

### DIFF
--- a/components/ExperimentPageView.test.tsx
+++ b/components/ExperimentPageView.test.tsx
@@ -83,11 +83,11 @@ const getButtonStates = () => {
 }
 
 test('experiment page view renders results without crashing', async () => {
-  const { container: _container } = await renderExperimentPageView({}, ExperimentView.Results)
+  await renderExperimentPageView({}, ExperimentView.Results)
 })
 
 test('experiment page view renders code-setup without crashing', async () => {
-  const { container: _container } = await renderExperimentPageView({}, ExperimentView.CodeSetup)
+  await renderExperimentPageView({}, ExperimentView.CodeSetup)
 })
 
 test('experiment page view renders with null experimentId without crashing', async () => {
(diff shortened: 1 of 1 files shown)


components/ExperimentPageView.test.tsx
```diff
@@ -77,11 +83,11 @@ const getButtonStates = () => {
 }
 
 test('experiment page view renders results without crashing', async () => {
-  const { container: _container } = renderExperimentPageView({}, ExperimentView.Results)
+  const { container: _container } = await renderExperimentPageView({}, ExperimentView.Results)
```
yanirs, line 86: Can also use this opportunity to remove the unnecessary assignments. :slightly_smiling_face:  (I'll do it now)

components/ExperimentPageView.test.tsx
```diff
@@ -49,7 +49,13 @@ const renderExperimentPageView = ({ experiment: experimentOverrides = {} }, view
...
   mockedAnalysesApi.findByExperimentId.mockImplementationOnce(async () => analyses)
 
-  return render(<ExperimentPageView experimentId={experiment.experimentId} view={view} debugMode={false} />)
+  // This is needed to fix warnings:
+  let renderResult = null
+  await act(async () => {
+    renderResult = render(<ExperimentPageView experimentId={experiment.experimentId} view={view} debugMode={false} />)
+  })
+
+  return renderResult as ReturnType<typeof render>
```
yanirs, line 58: Removing the unnecessary assignments of the return value from this function has revealed that we don't really care about what it returns, so I'll drop that too.

yanirs (approved):
Thanks for doing this, @jessie-ross! The output is much more manageable now. :slightly_smiling_face:

PR #337, opened 2020-09-22
Use localStorage for debug mode

jessie-ross:
**This PR changes debug mode from the query parameter to localStorage.**
- To enable it set `abacus-debug-mode` to `true` in your localStorage via your browser devTools
- You can also double click on the footer while holding shift
- I've also added a banner for debug mode as it is harder to know if it is enabled now:
Screen Shot 2020-09-23 at 1 52 48 am

Future PRs:
- Move the debug mode results to it's own hidden tab so we can leave debug mode enabled.
- In general we can use debug mode as a feature flag for unreleased features 

## How has this been tested?


yanirs:
Looks good! Just a couple of comments:

* Can you please document the toggling in the readme so we can look it up easily if we forget the magic shift + double click combo?
* Why do we lose the Abacus header when changing to debug mode? See gif (tested on Vercel):

Peek 2020-09-23 15-26

jessie-ross:
> Can you please document the toggling in the readme so we can look it up easily if we forget the magic shift + double click combo?

I've added it in! (I wasn't sure before whether we wanted to have it as a documented feature.)

> Why do we lose the Abacus header when changing to debug mode? See gif (tested on Vercel):

Yeah this is a weird SSR related bug where the wrong classes are being applied, this is likely because the server doesn't know about local storage. 

It is safe to ignore and goes away on navigation.

I am not exactly sure of the cause but fixes include:
- Moving debugMode to a React context so updates happen a bit better
- Dropping NextJS #212 

I say we just leave as is.

yanirs (approved):

PR #338, opened 2020-09-22
Seperate out debug tab

jessie-ross:
**This PR separates the debug results into the debug tab, allowing us to permanently keep debug mode on.**

Screen Shot 2020-09-23 at 2 01 42 am

See #337 

## How has this been tested?


components/ExperimentPageView.tsx
```diff
@@ -164,6 +166,14 @@ export default function ExperimentPageView({
               href='/experiments/[id]/results'
               hrefAs={`/experiments/${experimentId}/results`}
             />
+            <Tab
```
yanirs, line 169: Why is this always visible? The debug result view should be for us, not for experimenters.

jessie-ross: Ah oops, yep it is supposed to be debug mode only 🤦

components/ExperimentPageView.tsx
```diff
@@ -26,8 +26,9 @@ import ExperimentDisableButton from '@/components/ExperimentDisableButton'
 import Layout from '@/components/Layout'
 import { Analysis, ExperimentFull, Status } from '@/lib/schemas'
 import { useDataLoadingError, useDataSource } from '@/utils/data-loading'
-import { createUnresolvingPromise, or } from '@/utils/general'
+import { createUnresolvingPromise, isDebugMode, or } from '@/utils/general'
```
yanirs, line 29: Do we need this import in this file? It's a bit confusing since `debugMode` is an argument of the component.

components/ExperimentPageView.tsx
```diff
@@ -199,6 +211,9 @@ export default function ExperimentPageView({
               <ExperimentDetails {...{ experiment, metrics, segments, experimentReloadRef }} />
             )}
             {view === ExperimentView.Results && <ExperimentResults {...{ experiment, metrics, analyses, debugMode }} />}
+            {view === ExperimentView.Debug && isDebugMode && (
```
yanirs, line 214: Shouldn't `isDebugMode` be `isDebugMode()` or `debugMode`?

jessie-ross: Nice catch!

components/ExperimentPageView.tsx
jessie-ross: I could fix the names

yanirs: It's not so much the names, it's the mixing of the reliance on the global state from `isDebugMode()` versus the component's argument. I think we should use one or the other in a single file, not both.

jessie-ross: Well the idea here is that I like to keep "inputs" coming from the top level.

yanirs: OK, then let's drop the use of the function and rely on the `debugMode` argument. :slightly_smiling_face:

jessie-ross: Ah sorry I didn't look at this properly at the time, I'm definitely on your side on this.

yanirs (approved):
:ship:

PR #339, opened 2020-09-23
WIP: Add Autocomplete Everywhere

withinboredom:
This PR will likely need to be broken up. It depends on #330 

It's mostly here for sharing and testing while working out the kinks.

## How has this been tested?
None

### DIFF
--- a/api/AutocompleteApi.ts
+++ b/api/AutocompleteApi.ts
@@ -7,30 +7,64 @@ import {
     EventDetails, eventDetailsSchema
 } from "@/lib/schemas"
 import {fetchApi} from "@/api/utils"
+import NotFoundError from "@/api/NotFoundError";
 
 async function getCompletion<T extends Autocompletions>(name: string) {
     const autocompleteData = await fetchApi<T>('GET', `/autocomplete/${name}`)
     return await autocompleteSchema.validate(autocompleteData, {abortEarly: false})
 }
 
-export async function getUserCompletions(): Promise<UserCompletions> {
-    return getCompletion('users')
+const sessionCache: {
+    userCompletions: null | UserCompletions
+    eventCompletions: null | EventCompletions
+    eventDetails: any
+} = {
+    userCompletions: null,
+    eventCompletions: null,
+    eventDetails: {}
 }
 
-export async function getEventNameCompletions(): Promise<EventCompletions> {
-    const options = await getCompletion('events')
-    // todo: remove this
-    return {
-        completions: options.completions.map(v => {
-            return {
-                name: v.name,
-                value: v.name
+export async function getUserCompletions(): Promise<AutocompleteItem[]> {
+    return (sessionCache.userCompletions = sessionCache.userCompletions ?? await getCompletion('users')).completions
+}
+
+export async function getEventNameCompletions(): Promise<AutocompleteItem[]> {
+    // todo: change swagger file so this isn't necessary
+    return (sessionCache.eventCompletions = sessionCache.eventCompletions ?? await getCompletion('events')).completions.map(v => ({
+        name: v.name,
+        value: v.name
+    }))
+}
+
+export function getPropCompletions(eventName: string): () => Promise<AutocompleteItem[]> {
+    console.log('setting up', eventName)
+    const noEventFound = {
+        props: [
+            {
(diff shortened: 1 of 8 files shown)


api/AutocompleteApi.ts
```diff
@@ -0,0 +1,70 @@
...
+} from "@/lib/schemas"
+import {fetchApi} from "@/api/utils"
+import NotFoundError from "@/api/NotFoundError";
+
+async function getCompletion<T extends Autocompletions>(name: string) {
+    const autocompleteData = await fetchApi<T>('GET', `/autocomplete/${name}`)
+    return await autocompleteSchema.validate(autocompleteData, {abortEarly: false})
+}
+
+const sessionCache: {
```
jessie-ross, line 17: Instead of caching can we fetch the data at the page level and pass it down, just like we do with metrics and segments:
https://github.com/Automattic/abacus/blob/92ed0b47c0bdbd5ddad1227e044b3223d16c2c0c/pages/experiments/new.tsx#L37-L41

That will simplify the Autocomplete code below as well as it will no longer be responsible for any data fetching/data management.

api/utils.ts
```diff
@@ -13,7 +13,7 @@ import UnauthorizedError from './UnauthorizedError'
  *
  * @throws UnauthorizedError
  */
-async function fetchApi(method: string, path: string, body: unknown | null = null) {
+async function fetchApi<TReturn, TBody = {}>(method: string, path: string, body?: TBody) : Promise<TReturn> {
```
jessie-ross, line 16: Every endpoint we currently use takes the data and uses a schema to validate the incoming data.

Schema validators are super useful as they are of the form `(data: unknown) => SchemaType` and guarantee us the output type from any input.

This means we get a real level of type safety after fetching data in that we know the data matches the type that is output.

E.g:
https://github.com/Automattic/abacus/blob/92ed0b47c0bdbd5ddad1227e044b3223d16c2c0c/api/ExperimentsApi.ts#L113-L121

So I would say this isn't necessary.

components/experiment-creation/ExposureEventEditor.tsx
```diff
@@ -0,0 +1,181 @@
...
+            top: -3,
+            marginRight: theme.spacing(2),
+            color: theme.palette.text.disabled,
+        },
+    }),
+)
+
+export default function ExposureEventEditor({exposureEvent, index, onRemoveExposureEvent}: Props) {
+    const classes = useStyles()
+    const propRetriever = useCallback(getPropCompletions(exposureEvent.event), [exposureEvent])
```
jessie-ross, line 53: Ah, so each exposure event has it's own properties, can we fetch them all and pass them down?

```
{
   [exposureEventName]: [
      { completionKey: completionVal }
   ]
}
```

api/AutocompleteApi.ts
withinboredom: :+1: That's my thought as well. I mostly just did it as experimentation to see what would break.

api/utils.ts
withinboredom: Ah, that makes sense

components/experiment-creation/ExposureEventEditor.tsx
withinboredom: Perhaps we should change the api to accept a list of events instead of a singular one? That was my thought but I'm not sure how many events we'll have per experiment. If it's just a few, hammering the API with one, two, or three requests in quick succession probably isn't an issue. It's not really important how/when the data is retrieved at this point. Most of this PR was discovering how the autocomplete component works and it's shortcomings. I literally had to step through an entire render for [this little line](https://github.com/Automattic/abacus/pull/339/files#diff-bd3005f251d7145a49016b2ba539b338R71-R72) just because when the object comes from the API, it's a string, but when you make a selection, it's an object.

I'm still scratching my head how a textbox has a value of an object, but ok.

jessie-ross: We can't send one request to get them all?

withinboredom: It's too much data to get all the props for all events (16k+ entries, IIRC). However, I think making it so we can get the props for more than one event is probably a decent idea.

PR #343, opened 2020-09-24
Fix the graph title for the difference interval

aaronyan:
Currently reads "Conversation" when it should read "Conversion". 😄

## How has this been tested?


yanirs (approved):
:ship:

Issue #305, opened 2020-09-04
Use different Eslint rules for test files


jessie-ross:
We generally want to allow `@ts-ignore` and `no-await` for test files.

Looks like it might be possible: https://stackoverflow.com/a/32539830

_Originally posted by @jessie-ross in https://github.com/Automattic/abacus/pull/297#discussion_r483395145_

PR #345, opened 2020-09-25
Update allthethings

withinboredom:
This updates everything to the latest and greatest. In my testing, nothing looks broken even though there were some major upgrades.

## How has this been tested?


yanirs:
@withinboredom looks like this broke the linting: See [build output](https://app.circleci.com/pipelines/github/Automattic/abacus/1308/workflows/cc8f016e-7f82-4516-b4db-e42c598cdacd/jobs/1396).

By the way, you can use `npm run verify` to run the CircleCI checks locally.

withinboredom:
Oh, nice. For these types of merges, I've usually found that you don't want to fix them in the same PR. It takes time to address the issues while new code is written that might break this branch. Instead, the "best" path forward is to usually ignore them, and address individual rule fixes in singular PR's by un-ignoring the rules.

Thoughts about that @yanirs, @jessie-ross ?

yanirs:
I think it's worth having a quick look at individual errors since some of them seem to be about rules that no longer exist and some fixes may be trivial. In general, adding new rules to the ignore list and handling them separately (i.e., adding an issue for later) makes sense, as it doesn't affect code quality. We need the build to pass, though. :slightly_smiling_face:

jessie-ross:
I take a very conservative approach to updates, it can be a major source of bugs and problems in JavaScript land.

We should be extra careful for non-dev dependencies.
Here that would be:
- `notistack`
- `formik-material-ui`
- `formik-material-ui-lab`

Some process around it might be:
- Read the change-logs.
- Wait until a Major rev has been around for a while before updating.
- Writing about the above checks in update PRs:

I've just had a look at the above and it seems like there has been no breaking changes that will affect us.
The new releases have been around for a month, which is newish but is fine for the size of these changes and the scope of use of these packages.

Lastly, I do think this is an important time to check all warnings and rule fixes as all our tooling is changing under us.
I think it is an important layer of safety in updates.

So LGTM once everything passes.

components/MetricsTable.test.tsx
```diff
@@ -73,9 +73,7 @@ test('with some metrics, loads and opens metric details', async () => {
 
 test('with some metrics and canEditMetrics can click on the edit button', () => {
   const onEditMetric = jest.fn()
-  const { container: _container } = render(
-    <MetricsTable metrics={Fixtures.createMetricBares(2)} onEditMetric={onEditMetric} />,
-  )
+  render(<MetricsTable metrics={Fixtures.createMetricBares(2)} onEditMetric={onEditMetric} />)
```
withinboredom, line 76: Same as above with `useStyles`, it just bothered me, so I removed the return :)

components/RenderErrorView.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 
 import type { RenderError } from '@/components/RenderErrorBoundary'
 
-const RenderErrorView = (props: { renderError: RenderError }) => {
+export default function (props: { renderError: RenderError }): JSX.Element {
```
withinboredom, line 5: I don't remember why I changed this. But will change it back and see if I remember why I did it.

components/__snapshots__/ExperimentDetails.test.tsx.snap
```diff
@@ -2220,7 +2220,7 @@ exports[`renders as expected without conclusion data 1`] = `
               </h3>
               <div
                 class=""
-                title="Use \\"Edit in Wizard\\" for staging experiments."
+                title="Use \\\\\\"Edit in Wizard\\\\\\" for staging experiments."
```
withinboredom, line 2223: More shenanigans.

components/experiment-creation/Audience.tsx
```diff
@@ -71,6 +71,7 @@ const SegmentsAutocomplete = (
 ) => {
   const {
     form: { setFieldValue },
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     field: { name, value: outerValue },
```
withinboredom, line 75: `outerValue` comes from formik and is of type `any` so need to ignore.

components/experiment-creation/Audience.tsx
```diff
@@ -216,11 +217,13 @@ const Audience = ({
             options={Object.values(indexedSegments)}
             // TODO: Error state, see https://stackworx.github.io/formik-material-ui/docs/api/material-ui-lab
             renderInput={(params: AutocompleteRenderInputParams) => (
+              /* eslint-disable @typescript-eslint/no-unsafe-member-access */
               <MuiTextField
                 {...params}
                 variant='outlined'
                 placeholder={segmentAssignmentsField.value.length === 0 ? 'Search and select to customize' : undefined}
```
withinboredom, line 224: `segmentAssignmentsField.value` is of type `any`, so need to ignore.

components/experiment-creation/Audience.tsx
```diff
@@ -245,7 +248,7 @@ const Audience = ({
                 </TableRow>
               </TableHead>
               <TableBody>
-                {(formikProps.values.experiment.variations as VariationNew[]).map((variation, index) => {
+                {formikProps.values.experiment.variations.map((variation, index) => {
```
withinboredom, line 251: This is now typed. See below.

components/experiment-creation/ExperimentForm.test.tsx
```diff
@@ -500,12 +497,12 @@ test('form submits with valid fields', async () => {
       ],
       variations: [
         {
-          allocatedPercentage: 50,
+          allocatedPercentage: '50',
```
withinboredom, line 500: By having to specify the return type, I discovered that `allocatedPercentage` is supposed to be a `string` not a number.

components/experiment-creation/ExperimentForm.test.tsx
```diff
@@ -584,14 +581,18 @@ test('form submits an edited experiment without any changes', async () => {
 
   // We need to remove Ids, status, conclusion data, reformat exposure events to make it like new
   const newShapedExperiment = _.clone(experiment)
-  delete newShapedExperiment.experimentId
-  delete newShapedExperiment.status
+  newShapedExperiment.experimentId = (undefined as unknown) as number
```
withinboredom, line 584: you can't delete from a type when the value is required. So, I forcibly set them to undefined. Could also `ts-ignore` here.

components/experiment-creation/Metrics.tsx
```diff
@@ -98,13 +97,13 @@ const createMetricAssignment = (metric: MetricBare) => {
...
 
-const Metrics = ({ indexedMetrics }: { indexedMetrics: Record<number, MetricBare> }) => {
+const Metrics = ({ indexedMetrics }: { indexedMetrics: Record<number, MetricBare> }): JSX.Element => {
   const classes = useStyles()
 
   // Metric Assignments
-  const [metricAssignmentsField, _metricAssignmentsFieldMetaProps, metricAssignmentsFieldHelperProps] = useField<
-    MetricAssignment[]
-  >('experiment.metricAssignments')
+  const [metricAssignmentsField, , metricAssignmentsFieldHelperProps] = useField<MetricAssignment[]>(
```
withinboredom, line 104: I should change this back.

components/experiment-creation/Metrics.tsx
```diff
@@ -120,9 +119,7 @@ const Metrics = ({ indexedMetrics }: { indexedMetrics: Record<number, MetricBare
   }
 
   // ### Exposure Events
-  const [exposureEventsField, _exposureEventsFieldMetaProps, _exposureEventsFieldHelperProps] = useField<EventNew[]>(
-    'experiment.exposureEvents',
-  )
+  const [exposureEventsField, ,] = useField<EventNew[]>('experiment.exposureEvents')
```
withinboredom, line 122: Also this one.

components/experiment-results/CondensedLatestAnalyses.tsx
```diff
@@ -252,8 +252,8 @@ function AnalysisDetailPanel({
           line: {
             color: variantColors[index],
           },
-          mode: 'lines' as 'lines',
-          type: 'scatter' as 'scatter',
+          mode: 'lines' as const,
+          type: 'scatter' as const,
```
withinboredom, line 256: Must use `as const` now.

config.ts
```diff
@@ -10,7 +10,7 @@ const productionConfig = {
 const mockConfig = {
   experimentApi: {
     needsAuth: false,
-    authPath: null,
+    authPath: 'http://example.com',
```
withinboredom, line 13: Instead of being forced to now handle the "sometimes null case" I just set it to `example.com`

lib/form-data.ts
```diff
@@ -11,7 +11,31 @@ import {
...
+interface VariationFormData {
+  allocatedPercentage: string
+  isDefault: boolean
+  name: string
+}
+
+interface ExposureEventFormData {
+  event: string
+  props: { value: unknown; key: string }[]
+}
```
withinboredom, line 36: As mentioned above, these are now typed due to having to specify return types. This is good because it resulted in discovering the variation split being a `string` and not a number. I'm pretty sure a form input will translate a number to a string just fine if we wanted to go that route.

lib/form-data.ts
```diff
@@ -21,32 +45,47 @@ function metricAssignmentToFormData(metricAssignment: MetricAssignment) {
...
     allocatedPercentage: String(variation.allocatedPercentage),
   }
 }
 
-function exposureEventToFormData(exposureEvent: Event) {
+function exposureEventToFormData(exposureEvent: Event): ExposureEventFormData {
   return {
     event: exposureEvent.event,
-    props: Object.entries(exposureEvent.props as object).map(([key, value]) => ({ key, value })),
+    props: Object.entries(exposureEvent.props as Record<string, unknown>).map(([key, value]) => ({ key, value })),
```
withinboredom, line 66: `as object` or `as {}` is no longer allowed, instead use `Record<string unknown>`

lib/schemas.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
withinboredom, line 1: This is needed since we use `any`'s to test the schema.

lib/schemas.ts
```diff
@@ -92,13 +99,16 @@ export const metricFullSchema = metricBareSchema
   .defined()
   .camelCase()
   .test('event-params-required', 'Event Params is required and must be valid JSON.', (metricFull) => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
     return !(metricFull.parameterType === MetricParameterType.Conversion && !metricFull.eventParams)
```
withinboredom, line 103: It's debatable whether we should do proper assertions to turn the `any` into what we expect, but ignoring this for now.

pages/auth.tsx
```diff
@@ -56,14 +56,27 @@ const AuthPage = function AuthPage() {
...
+      switch (typeof str) {
+        case 'undefined':
+          return def
+        case 'string':
+          return str
+        default:
+          return str.join(', ')
+      }
+    }
+
```
withinboredom, line 69: Since query strings may be no value (`?`), a single value (`?key=value`), or an array (`?key[]=value1&key[]=value2`) we need to handle all these cases to resolve down to a single string.

pages/auth.tsx
```diff
@@ -75,7 +88,7 @@ const AuthPage = function AuthPage() {
       return
     }
 
-    const expiresInSeconds = toInt(authInfo.expires_in)
+    const expiresInSeconds = toInt(authInfo.expires_in, 0) as unknown
```
withinboredom, line 91: I added a default fo when it's not a number.

test-helpers/test-utils.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-return,@typescript-eslint/explicit-module-boundary-types */
```
withinboredom, line 1: There are a ton of errors in this file for use of `any`, so I just disabled it since it is't our fault :)

utils/auth.ts
```diff
@@ -14,8 +14,10 @@ interface ExperimentsAuthInfo {
  */
 const getExperimentsAuthInfo = (): ExperimentsAuthInfo | null => {
   try {
-    const experimentsAuthInfo = JSON.parse(localStorage.getItem('experiments_auth_info') || 'null')
-    if (experimentsAuthInfo && experimentsAuthInfo.expiresAt > Date.now()) {
+    const experimentsAuthInfo = JSON.parse(
+      localStorage.getItem('experiments_auth_info') || 'null',
+    ) as ExperimentsAuthInfo | null
```
withinboredom, line 19: I think it's probably safe to assume that it will be one of these two types. A todo would be to use `yup` to validate it to prevent malicious payloads.

utils/data-loading.ts
```diff
@@ -31,12 +31,16 @@ export function useDataSource<Data, Deps extends DependencyList | undefined, E e
...
-        .catch((error) => isSubscribed && setError(error))
-        .finally(() => isSubscribed && setIsLoading(false))
+      try {
+        const data = await createDataPromise()
+        if (isSubscribed) setData(data)
+      } catch (error) {
+        if (isSubscribed) setError(error)
+      } finally {
+        if (isSubscribed) setIsLoading(false)
+      }
```
withinboredom, line 43: I rewrote this as a try/catch so coverage requirements would catch that the error case wasn't being tested. Previously, it considered it all one line.

withinboredom (commented):
I've left comments explaining how the rules changed the code and highlighted any areas changed for other reasons. There's a few shenanigans going on as well, so this PR isn't over. I'm happy to break this PR up into as little as six PRs:

1. Add returns types to everything
2. Add `void` before promises we don't await
3. (maybe before 1) fix the bugs highlighted in this PR
4. Fix the use of `any`
5. Actually update everything
6. Fix any remaining issues.

This would take longer, but be marginally safer. Or we can spend a few days reviewing this and merge as-is. There's a lot here, but I feel like the code is "safer."

yanirs:
> I'm happy to break this PR up into as little as six PRs [...] This would take longer, but be marginally safer. 

I think it may actually take less time as it'd be easy to approve the non-functional changes separately, but I'll @jessie-ross decide as the assigned reviewer. :slightly_smiling_face:

components/ExperimentsTable.tsx
```diff
@@ -15,14 +15,14 @@ const debug = debugFactory('abacus:components/ExperimentsTable.tsx')
...
-const ExperimentsTable = ({ experiments }: { experiments: ExperimentBare[] }) => {
+const ExperimentsTable = ({ experiments }: { experiments: ExperimentBare[] }): JSX.Element => {
   debug('ExperimentsTable#render')
   const router = useRouter()
   const theme = useTheme()
 
   /* istanbul ignore next; to be handled by an e2e test */
   const handleRowClick = (event?: React.MouseEvent, rowData?: ExperimentBare) => {
-    router.push('/experiments/[id]', `/experiments/${rowData?.experimentId}`)
+    void router.push('/experiments/[id]', `/experiments/${rowData?.experimentId ?? ''}`)
```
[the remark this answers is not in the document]
yanirs: I like it, as such promises were causing warnings in the tests.

components/GeneralPanel.test.tsx
```diff
@@ -44,7 +45,7 @@ test('renders as expected', () => {
           </h3>
           <div
             class=""
-            title="Use \\"Edit in Wizard\\" for staging experiments."
+            title="Use \\\\\\"Edit in Wizard\\\\\\" for staging experiments."
```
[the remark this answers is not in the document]
yanirs: :joy:

components/experiment-creation/Audience.tsx
yanirs: Can we cast it to a known type instead?

api/utils.ts
```diff
@@ -56,7 +56,7 @@ async function fetchApi(method: string, path: string, body: unknown | null = nul
   if (responseText === '') {
     return
   }
-  return JSON.parse(responseText)
+  return JSON.parse(responseText) as unknown
```
yanirs, line 59: Should we make this `Record<string, unknown>`?

components/Autocomplete.test.tsx
```diff
@@ -17,8 +17,7 @@ import { changeFieldByRole, MockFormik } from '@/test-helpers/test-utils'
 document.createRange = () => ({
   setStart: () => undefined,
   setEnd: () => undefined,
-  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-  // @ts-ignore; This is just for mocking
+  // @ts-ignore
```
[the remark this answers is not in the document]
yanirs: Yeah, there's an issue to suppress these eslint errors on all test files: #305.

components/ExperimentPageView.test.tsx
```diff
@@ -49,9 +48,11 @@ const renderExperimentPageView = async ({ experiment: experimentOverrides = {} }
   const analyses = Fixtures.createAnalyses()
   mockedAnalysesApi.findByExperimentId.mockImplementationOnce(async () => analyses)
 
+  const debug = view === ExperimentView.Debug
```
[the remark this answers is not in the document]
yanirs: I see no harm in keeping them. :slightly_smiling_face:

components/ExperimentRunButton.tsx
```diff
@@ -20,8 +20,8 @@ const ExperimentRunButton = ({
 }: {
   experiment: ExperimentFull | null
   experimentReloadRef: React.MutableRefObject<() => void>
-}) => {
-  const _classes = useStyles()
+}): JSX.Element => {
+  useStyles()
```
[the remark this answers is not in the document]
yanirs: Yeah, there are a few of these lying around. No need to assign to an unused variable. :slightly_smiling_face:

components/MetricsTable.test.tsx
yanirs: It looks much better. :slightly_smiling_face:

components/__snapshots__/ExperimentDetails.test.tsx.snap
yanirs: \\\\\\\\\\\\\\\\\\\\\\\\\

components/experiment-results/CondensedLatestAnalyses.tsx
yanirs: It looks better than `'str' as 'str'` :slightly_smiling_face:

lib/schemas.ts
yanirs: It might be better to suppress on the file level and explain why with an inline comment.

pages/auth.tsx
yanirs: Should it still be `as unknown` then?

pages/auth.tsx
```diff
@@ -56,14 +56,27 @@ const AuthPage = function AuthPage() {
     // #error=access_denied&error_description=You+need+to+log+in+to+WordPress.com&state=
     const authInfo = qs.parse(window.location.hash.substring(1))
 
+    const getStringFromQueryString = (str: string | string[] | undefined, def = 'unkown'): string => {
```
yanirs, line 59: `unkown` -> `unknown`?

components/experiment-creation/Audience.tsx
withinboredom: Maybe, but I didn't investigate what that type would be. It would be better if we either checked to see if the project has a `@types/` repo or just wait for them to export their types.

pages/auth.tsx
withinboredom: Yes, because it's `unknown` because it may still not be a number :D what joy javascript is... (aka `NaN` which is still a number but not numeric, IIRC)

api/utils.ts
withinboredom: Trust me, this is simpler :) and it may be a non-object (aka, `null`). We also return `undefined` which complicates things...

utils/data-loading.ts
yanirs: Maybe we can also make the `isSubscribed` case more explicit?

```suggestion
      if (isSubscribed) {
        try {
          setData(await createDataPromise())
        } catch (error) {
          setError(error)
        } finally {
          setIsLoading(false)
        }
      } else {
        try {
          await createDataPromise()
        } catch (error) {}
      }
```

yanirs:
> > I'm happy to break this PR up into as little as six PRs [...] This would take longer, but be marginally safer.
> 
> I think it may actually take less time as it'd be easy to approve the non-functional changes separately, but I'll @jessie-ross decide as the assigned reviewer. :slightly_smiling_face:

I did a quick pass because I was curious and it's actually not too bad when marking files as viewed:

image

This would still get stale quickly, so it may be worth cherry picking the bulk of this PR if some discussions drag on.

api/utils.ts
yanirs: We don't return `null` from the API, only an empty response or an object. We can change the return type of the function to `Promise<undefined | Record<string, unknown>>` to account for the `undefined` case.

I don't feel _too_ strongly about it, but it seems like it's worth being more specific where possible.

yanirs: Oh, and we can also return an empty object instead of `undefined` when the response is empty to make it cleaner.

api/MetricsApi.ts
```diff
@@ -51,8 +48,13 @@ async function put(metricId: number, newMetric: MetricFullNew) {
  */
 async function findAll(): Promise<MetricBare[]> {
   // istanbul ignore next; debug only
-  const { metrics } = await fetchApi('GET', isDebugMode() ? '/metrics?debug=true' : '/metrics')
-  return await yup.array(metricBareSchema).defined().validate(metrics, { abortEarly: false })
+  const { metrics } = await metricBareResponse.validate(
```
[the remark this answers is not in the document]
jessie-ross: I would prefer to create inline schemas for incidental data types like response objects so we can keep our core data types to a minimum:

```const { metrics } = await yup.object({ metrics: yup.array(metricBareSchema) }).validate(``` 

And if we don't we should follow our naming convention for schemas `metricBareResponseSchema`.

File: api/SegmentsApi.ts [hunk 30]
```diff
@@ -12,8 +10,8 @@ import { fetchApi } from './utils'
  * @throws UnauthorizedError
  */
 async function findAll(): Promise<Segment[]> {
-  const { segments } = await fetchApi('GET', '/segments')
-  return await yup.array(segmentSchema).defined().validate(segments, { abortEarly: false })
+  const { segments } = await segmentResponseSchema.validate(await fetchApi('GET', '/segments'), { abortEarly: false })
```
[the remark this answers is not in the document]
jessie-ross: See my comment on the metric endpoint

components/Autocomplete.test.tsx
```diff
@@ -128,7 +127,6 @@ test('An autocomplete component can be rendered', async () => {
 
 test('it shows as loading data', () => {
   const isLoading = true
-  const options: AutocompleteItem[] = []
```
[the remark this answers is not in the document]
jessie-ross: Wow I would find this super annoying but ok

components/ExperimentRunButton.tsx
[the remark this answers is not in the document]
jessie-ross: There shouldn't be any side effect here so this line can be removed.

I leave them in as eventually every element needs styles and I'd have to add all this code back in. It is useful for rapid UI development.

components/ExperimentStatus.tsx
```diff
@@ -32,11 +32,11 @@ const useStyles = makeStyles((theme: Theme) =>
   }),
 )
 
-function ExperimentStatus({ status }: { status: Status }) {
+function ExperimentStatus({ status }: { status: Status }): JSX.Element {
```
jessie-ross, line 35: I am surprised these JSX.Element types work, there must be some typescript magic going on - typescript still knows what props and prop types our components accept.

components/RenderErrorView.tsx
jessie-ross: This should be fine

utils/data-loading.ts
jessie-ross: @yanirs I don't think this will work as it is possible for `isSubscribed` to have changed while waiting for the promise resolution, and we want the value at the execution time of `setX`. Unfortunately the new await syntax makes it a little less obvious that this is happening in a different "queued sub-task", and where exactly that line is.

@withinboredom I am surprised the inline `if` statements here pass our code conventions? We don't do that anywhere AFAIK.

jessie-ross (changes requested):
Phew! I made it! See my comments.

I don't mind reviewing it like this, but for the future a good way to split this is by package, we could update all the other packages and then do typescript and eslint separately.

And then with eslint if there are new rules enabled by default we can disable them and then add them on one by one along with its associated changes.

api/MetricsApi.ts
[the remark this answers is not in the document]
withinboredom: Oh nice, didn't know that was possible but it makes sense!

utils/data-loading.ts
withinboredom: > I am surprised the inline if statements here pass our code conventions? We don't do that anywhere AFAIK.

I always write inline ifs unless it's more than one statement and let autoformat / linters fix them if they aren't allowed (habit from writing C/C++ for years) but maybe the rules changed with the updated eslint and it wasn't caught. I'll update the code here.

api/utils.ts
withinboredom: See https://github.com/Automattic/abacus/pull/339/commits/e43cf76b6222543cfddc69d0b0e93852393fb5f0#r493677967

utils/data-loading.ts
jessie-ross: Yeah I generally prefer having stable syntax (syntax that doesn't change if the semantics doesn't), this means less editing and better diffs.

utils/data-loading.test.ts
```diff
@@ -75,9 +78,9 @@ describe('utils/data-loading.ts module', () => {
       })
     })
 
-    it('should resolve promise after unmount without logging an error', () => {
+    it('should resolve promise after unmount without logging an error', async () => {
       const originalConsoleError = console.error
-      console.error = jest.fn()
+      console.error = jest.fn((d) => originalConsoleError(d))
```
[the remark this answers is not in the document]
withinboredom: We're not expecting the error to actually occur, but if it does, it's useful to output it and jest will catch it as well. I ran into errors being output in these tests that shouldn't have been so I added this to see what the errors were so I could fix them.

pages/auth.tsx
```diff
@@ -44,7 +44,7 @@ const AuthPage = function AuthPage() {
 
     if (!window.location.hash || window.location.hash.length === 0) {
       console.error('Authentication Error:', 'Missing hash in auth callback url.')
-      setError(`An unknown error has occured: Missing hash in auth callback url.`)
+      setError(`An unknown error has occurred: Missing hash in auth callback url.`)
```
jessie-ross, line 47: Lol nice catch!

jessie-ross (approved):
LGTM! 
🛸 🛸 🛸

Issue #323, opened 2020-09-11
Use values from auto-complete endpoints


aaronyan:
After the endpoints have been created, we need to consume them and implement them into Abacus. This issue will be updated with the endpoint documentation once they have been engineered.

aaronyan:
@withinboredom what is the status on event prop value autocomplete? Testing out the new event autocompletes (feels and looks great btw!), I noticed the eventprop `key` works, but the `values` don't.

PR #346, opened 2020-09-29
Add an autocomplete component

withinboredom:
This component is a simple wrapper around the material UI autocomplete component See tests on how to use.

## How has this been tested?

Storybook

### DIFF
--- a/components/Autocomplete.test.tsx
+++ b/components/Autocomplete.test.tsx
@@ -1,0 +1,239 @@
+/* eslint-disable @typescript-eslint/require-await */
+import { InputAdornment, TextField as MuiTextField } from '@material-ui/core'
+import { act, fireEvent, render, screen } from '@testing-library/react'
+import { Field } from 'formik'
+import { Autocomplete, AutocompleteRenderInputParams } from 'formik-material-ui-lab'
+import React from 'react'
+
+import AbacusAutocomplete, {
+  autocompleteAttributes,
+  autocompleteInputProps,
+  getOptionValue,
+} from '@/components/Autocomplete'
+import { AutocompleteItem } from '@/lib/schemas'
+import { changeFieldByRole, MockFormik } from '@/test-helpers/test-utils'
+
+// Needed for testing the autocomplete popper
+document.createRange = () => ({
+  setStart: () => undefined,
+  setEnd: () => undefined,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore; This is just for mocking
+  commonAncestorContainer: {
+    nodeName: 'BODY',
+    ownerDocument: document,
+  },
+})
+
+test('An autocomplete component can be rendered', async () => {
+  const isLoading = false
+  const options = [
+    {
+      value: 'stuff',
+      name: 'Stuff',
+    },
+    {
+      value: 'other_stuff',
+      name: 'Other Stuff',
+    },
+  ]
+
+  const { container } = render(
+    <MockFormik initialValues={{ name: 'other_stuff' }}>
+      <Field
+        component={AbacusAutocomplete}
+        name='name'
+        id='name'
+        fullWidth
+        options={options}
+        loading={isLoading}
+        renderInput={(params: AutocompleteRenderInputParams) => {
+          return (
+            <MuiTextField
+              {...params}
+              placeholder='wp_username'
+              helperText='Use WordPress.com username.'
+              variant='outlined'
+              required
+              label='Owner'
+              InputProps={{
+                ...autocompleteInputProps(params, isLoading),
+                startAdornment: <InputAdornment position='start'>@</InputAdornment>,
+              }}
+              InputLabelProps={{
+                shrink: true,
+              }}
+            />
+          )
+        }}
+      />
+    </MockFormik>,
+  )
+  expect(container).toMatchSnapshot('initial render')
+
+  const textbox = screen.getByRole('textbox')
+  expect(textbox).toMatchInlineSnapshot(`
+    <input
+      aria-autocomplete="list"
+      aria-describedby="name-helper-text"
+      aria-invalid="false"
+      autocapitalize="none"
+      autocomplete="off"
+      class="MuiInputBase-input MuiOutlinedInput-input MuiAutocomplete-input MuiAutocomplete-inputFocused MuiInputBase-inputAdornedStart MuiOutlinedInput-inputAdornedStart MuiInputBase-inputAdornedEnd MuiOutlinedInput-inputAdornedEnd"
+      id="name"
+      placeholder="wp_username"
+      required=""
+      spellcheck="false"
+      type="text"
+      value="other_stuff"
+    />
+  `)
+
+  await act(async () => {
+    fireEvent.click(textbox)
+  })
+  expect(container).toMatchSnapshot('clicked textbox')
+
+  await act(async () => {
+    await changeFieldByRole('textbox', /Owner/, 'stuff')
+  })
+  expect(container).toMatchSnapshot('entered value')
+
+  await act(async () => {
+    fireEvent.click(screen.getByText('Stuff'))
+  })
+  expect(container).toMatchSnapshot('clicked result')
+  expect(textbox).toMatchInlineSnapshot(`
+    <input
+      aria-autocomplete="list"
+      aria-describedby="name-helper-text"
+      aria-invalid="false"
+      autocapitalize="none"
+      autocomplete="off"
+      class="MuiInputBase-input MuiOutlinedInput-input MuiAutocomplete-input MuiAutocomplete-inputFocused MuiInputBase-inputAdornedStart MuiOutlinedInput-inputAdornedStart MuiInputBase-inputAdornedEnd MuiOutlinedInput-inputAdornedEnd"
+      id="name"
+      placeholder="wp_username"
+      required=""
+      spellcheck="false"
+      type="text"
+      value="stuff"
+    />
+  `)
+
+  await act(async () => {
+    fireEvent.click(textbox)
+  })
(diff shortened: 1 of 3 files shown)


components/Autocomplete.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+              reloadData
+              getCompletionData={fakeData}
+            />
+          )
+        }}
+      </Formik>
+    </div>,
+  )
+
+  expect(container).toMatchInlineSnapshot(`
```
jessie-ross, line 45: FYI: No need to change this but I think we have generally settled on using non-inline snapshots for large snapshots.

components/Autocomplete.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const [open, setOpen] = useState(false)
+  const [options, setOptions] = useState<T[]>([])
+  const loading = open && options.length === 0
+
+  // this effect allows for deleting the stored state when the source state may have changed.
+  useEffect(() => {
+    if (reloadData) {
+      setOptions([])
+    }
+  }, [reloadData])
```
jessie-ross, line 33: Not a fan of passing a change variable from the outside, and I am not sure what the requirement of this is?

components/Autocomplete.tsx
```diff
@@ -0,0 +1,96 @@
...
+      const response = await getCompletionData()
+      if (active) {
+        setOptions(response)
+      }
+    })()
+
+    return () => {
+      active = false
+    }
+  }, [loading, getCompletionData])
```
jessie-ross, line 52: If we did want `reloadData` it would make sense to add it as a dep here.

I think this component could benefit from moving the options and data loading out of this component.

We have the data loading already abstracted as `useDataSource` so we then would have a nice composition of simple abstractions (simpler, more reusable):

```
// This could be up with the rest of the data loading, we could have a `completionsBag` that we pass down from high.
const { data: options, isLoading, error } = useDataSource(Users.getUserCompletions, [])

return <Autocomplete {...{ options, isLoading }} />
```

Inside Autocomplete would likely have loading state as `options === null || isLoading`.

withinboredom: So, I considered this, but this is where it gets kinda interesting... for things like user names, event names, product names, etc., it can work like you mention, where we can prefetch the data and pass it on down. Then you get to when a user enters the event name and we need to fetch the props, also fairly simple. Then the user changes the event name after selecting some props... and everything kinda falls apart. :) The reason it kinda falls apart is that the event name we need is buried in the form data, as we're rendering, and the props can change under us. So, we need to get that information out and pass the correct data back in. In the reference implementation, I had a function that returns a function to handle this case.

I like the idea of a custom hook, I'll look into that, but maybe that should be in #348?

components/Autocomplete.test.tsx
withinboredom: I did that mostly to make it easier to check the output HTML.

components/Autocomplete.tsx
jessie-ross: Ahh I see what you mean, I think I have some solutions here:

> I like the idea of a custom hook, I'll look into that, but maybe that should be in #348?

So not sure if you mean creating a different hook but just in case I might mention we do already have `useDataSource` here:
https://github.com/Automattic/abacus/blob/227faab8a5d4dc552baa0aa63bfa637cd8a3c44d/utils/data-loading.ts#L5-L49

I think it should do everything you require.

> So, I considered this, but this is where it gets kinda interesting... for things like user names, event names, product names, etc., it can work like you mention, where we can prefetch the data and pass it on down. Then you get to when a user enters the event name and we need to fetch the props, also fairly simple. 

I would personally do the ones we can from the top level, keep what we can simple.

> Then the user changes the event name after selecting some props... and everything kinda falls apart. :) The reason it kinda falls apart is that the event name we need is buried in the form data, as we're rendering, and the props can change under us. So, we need to get that information out and pass the correct data back in. In the reference implementation, I had a function that returns a function to handle this case.

For this one we can access the form state in the component that holds both the event name and event props and load it there, something like this:

```javascript
const [{ value: eventName }] = useField(`experiment.exposureEvents[${exposureEventIndex}].name`)
const { data: eventPropsOptionsData, isLoading: eventPropsOptionsIsLoading, error: eventPropsOptionsError } = useDataSource(() => Autocompletes.getEventProps(eventName), [eventName])
```

withinboredom:
Ok, it's a lot simpler now, also resolved a bunch of warnings and other fun stuff.

components/Autocomplete.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { AutocompleteItem } from '@/lib/schemas'
+
+export interface AutocompleteAttributes {
+  renderInput: (params: AutocompleteRenderInputParams) => unknown
+  getOptionSelected: (option: AutocompleteItem, value: string | AutocompleteItem) => boolean
+  options: AutocompleteItem[]
+  loading: boolean
+  getOptionLabel: (option: AutocompleteItem | string | null) => string
+}
```
jessie-ross, line 14: Not sure I see the need for this interface? Are you using it in a follow up PR?

withinboredom: Yeah, it's coming in a sec.

components/Autocomplete.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return {
+    ...params.InputProps,
+    endAdornment: (
+      <>
+        {loading ? <CircularProgress color='inherit' size={20} /> : null}
+        {params.InputProps.endAdornment}
+      </>
+    ),
+  }
+}
```
jessie-ross, line 40: Can this be merged in by default?

jessie-ross (commented):
Nice it is really good and much simpler!

I was confused a bit by all the exports and then I realised you are using them for testing, that is a bit unfortunate but I am not sure what you can do... It would be cool if there were test only imports?

Could probably use a comment about why you are doing the fancy `getOption*` functions.

withinboredom:
> I was confused a bit by all the exports and then I realised you are using them for testing, that is a bit unfortunate but I am not sure what you can do... It would be cool if there were test only imports?

I've often felt the exact same way... if only!

> Could probably use a comment about why you are doing the fancy getOption* functions.

Good idea!

components/Autocomplete.tsx
withinboredom: Oh, actually I thought you were talking about a different interface that isn't in this PR at all ... I'll remove it!

components/Autocomplete.tsx
withinboredom: Let me see what I can do...

withinboredom: I can't figure out how to make that work. Any ideas?

jessie-ross: Something like:

```javascript
return <Autocomplete 
    {...fieldToAutocomplete(props)} 
    {...autocompleteAttributes} 
    onChange={onChange}
    InputProps={{
        ...(props.InputProps || {}), 
        endAdornment: (
            <>
                {loading ? <CircularProgress color='inherit' size={20} /> : null}
                {props?.InputProps?.endAdornment}
            </>
        ),
    }} 
/>

Comment (jessie-ross) [reply components/Autocomplete.tsx:32]: Ah I see, we are not specifying the `renderInput` now... Maybe this is easiest...

Although I wouldn't mind just removing `autocompleteInputProps` and just manually adding the endAdornment:
```js
        renderInput={(params: AutocompleteRenderInputParams) => {
          return (
            <MuiTextField
              {...params}
              placeholder='wp_username'
              helperText='Use WordPress.com username.'
              variant='outlined'
              required
              label='Owner'
              InputProps={{
                endAdornment:  loading ? <CircularProgress color='inherit' size={20} /> : undefined
                startAdornment: <InputAdornment position='start'>@</InputAdornment>,
              }}
              InputLabelProps={{
                shrink: true,
              }}
            />
          )
        }}
```

It would give us more control and (depending on you thoughts on code) be less tangled.

withinboredom: It would look more like:

```tsx
        renderInput={(params: AutocompleteRenderInputParams) => {
          return (
            <MuiTextField
              {...params}
              placeholder='wp_username'
              helperText='Use WordPress.com username.'
              variant='outlined'
              required
              label='Owner'
              InputProps={{
                ...params.InputProps,
                endAdornment:  (
                  <>
                    {loading ? <CircularProgress color='inherit' size={20} /> : undefined}
                    {params.InputProps.endAdornment}
                  </>
                )
                startAdornment: <InputAdornment position='start'>@</InputAdornment>,
              }}
              InputLabelProps={{
                shrink: true,
              }}
            />
          )
        }}
```

We can try it, it just seemed very human-error-prone.

withinboredom: oh, we're keeping the mixin... :D when I removed it, I discovered we have to test the loading case for every component to get 100% coverage. I don't think we want/need to do that.

jessie-ross: Oh well

jessie-ross (approved):
Beautiful! Looking nice and simple!

🚢 🚢 🚢

components/Autocomplete.tsx
```diff
@@ -0,0 +1,49 @@
+import { CircularProgress } from '@material-ui/core'
+import { Autocomplete } from '@material-ui/lab'
+import { AutocompleteProps, AutocompleteRenderInputParams, fieldToAutocomplete } from 'formik-material-ui-lab'
+import React, { useCallback } from 'react'
+
+import { AutocompleteItem } from '@/lib/schemas'
+
+// break out these functions to achieve 100% coverage
```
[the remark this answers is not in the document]
withinboredom: haha, ah, good point.

aaronyan:
Used to resolve #323

PR #347, opened 2020-09-29
Add an existing experiment form

withinboredom:
Adds another form example with some example data.

## How has this been tested?

Storybook


jessie-ross (approved):
LGTM

PR #348, opened 2020-09-29
Add user completions

withinboredom:
Adds user completions to the basic information panel.

## How has this been tested?

### DIFF
--- a/api/AutocompleteApi.ts
+++ b/api/AutocompleteApi.ts
@@ -1,5 +1,10 @@
 import { fetchApi } from '@/api/utils'
 import { AutocompleteItem, autocompleteSchema, Autocompletions } from '@/lib/schemas'
+import { DataSourceResult } from '@/utils/data-loading'
+
+export interface CompletionBag {
+  userCompletions: DataSourceResult<AutocompleteItem[]>
+}
 
 async function getCompletion<T extends Autocompletions>(name: string) {
   const autocompleteData = await fetchApi('GET', `/autocomplete/${name}`)
--- a/api/AutocompleteApi.ts
+++ b/api/AutocompleteApi.ts
@@ -3,7 +3,7 @@ import { AutocompleteItem, autocompleteSchema, Autocompletions } from '@/lib/sch
 import { DataSourceResult } from '@/utils/data-loading'
 
 export interface CompletionBag {
-  userCompletions: DataSourceResult<AutocompleteItem[]>
+  userCompletionDataSource: DataSourceResult<AutocompleteItem[]>
 }
 
 async function getCompletion<T extends Autocompletions>(name: string) {
--- a/api/AutocompleteApi.ts
+++ b/api/AutocompleteApi.ts
@@ -1,0 +1,11 @@
+import { fetchApi } from '@/api/utils'
+import { AutocompleteItem, autocompleteSchema, Autocompletions } from '@/lib/schemas'
+
+async function getCompletion<T extends Autocompletions>(name: string) {
+  const autocompleteData = await fetchApi('GET', `/autocomplete/${name}`)
+  return await autocompleteSchema.validate(autocompleteData, { abortEarly: false })
+}
+
+export async function getUserCompletions(): Promise<AutocompleteItem[]> {
+  return (await getCompletion('users')).completions
+}
--- a/components/experiment-creation/BasicInfo.stories.tsx
+++ b/components/experiment-creation/BasicInfo.stories.tsx
@@ -1,13 +1,18 @@
 import React from 'react'
 
+import { getUserCompletions } from '@/api/AutocompleteApi'
 import { MockFormik } from '@/test-helpers/test-utils'
+import { useDataSource } from '@/utils/data-loading'
 
 import BasicInfo from './BasicInfo'
 
 export default { title: 'ExperimentCreation.Form Parts.BasicInfo' }
 
-export const FormPart = () => (
-  <MockFormik>
-    <BasicInfo />
-  </MockFormik>
-)
+export const FormPart = () => {
+  const userCompletions = useDataSource(getUserCompletions, [])
+  return (
+    <MockFormik>
+      <BasicInfo userCompletions={userCompletions} />
+    </MockFormik>
+  )
+}
--- a/components/experiment-creation/BasicInfo.stories.tsx
+++ b/components/experiment-creation/BasicInfo.stories.tsx
@@ -9,10 +9,12 @@ import BasicInfo from './BasicInfo'
 export default { title: 'ExperimentCreation.Form Parts.BasicInfo' }
 
 export const FormPart = () => {
-  const userCompletions = useDataSource(getUserCompletions, [])
+  const completionBag = {
+    userCompletions: useDataSource(getUserCompletions, []),
+  }
   return (
     <MockFormik>
-      <BasicInfo userCompletions={userCompletions} />
+      <BasicInfo completionBag={completionBag} />
     </MockFormik>
   )
 }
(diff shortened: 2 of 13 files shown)


api/ExperimentsApi.ts
```diff
@@ -18,13 +19,33 @@ import {
...
 import { fetchApi } from './utils'
 
+function maybeAutocompleteItem(item: AutocompleteItem | string): string {
+  if (typeof item === 'object') {
+    return item.value
+  }
+  return item
+}
+
+// todo: is this possible using nothing but yup?
```
jessie-ross, line 29: Yup! Should be possible using `.transform()` for casts.

But a better place to do it might be within the component like we do for segments:
https://github.com/Automattic/abacus/blob/7a9e65b59347faaf5d87803d54a1bdf84b8d0646/components/experiment-creation/Audience.tsx#L78-L100

jessie-ross: We transform the value coming in, and the changed value on its way out.

components/experiment-creation/BasicInfo.tsx
```diff
@@ -37,7 +39,7 @@ const useStyles = makeStyles((theme: Theme) =>
   }),
 )
 
-const BasicInfo = () => {
```
jessie-ross, line 40: It's a bit weird to be passing a function for this rather than the data.

api/ExperimentsApi.ts
withinboredom: > Yup!

:i-see-what-you-did-there: :laughing: 

Thanks for the heads up of where to look next!

components/experiment-creation/ExperimentForm.tsx
```diff
@@ -106,10 +107,12 @@ const ExperimentForm = ({
   indexedSegments,
   initialExperiment,
   onSubmit,
+  userCompletions,
```
jessie-ross, line 110: In the future we may want to throw down a `completionsBag` but no harm in refactoring it later.

pages/experiments/new.tsx
```diff
@@ -73,7 +75,7 @@ const ExperimentsNewPage = function () {
       </div>
       {isLoading && <LinearProgress className={classes.progress} />}
       {!isLoading && indexedMetrics && indexedSegments && (
-        <ExperimentForm {...{ indexedMetrics, indexedSegments, initialExperiment, onSubmit }} />
+        <ExperimentForm {...{ indexedMetrics, indexedSegments, initialExperiment, onSubmit, userCompletions }} />
```
jessie-ross, line 78: Ah this threw me off, I had expected this to be an array of completions. What do you think about `userCompletionsDataSourceResult`?

api/AutocompleteApi.ts
```diff
@@ -0,0 +1,11 @@
+import { fetchApi } from '@/api/utils'
+import { AutocompleteItem, autocompleteSchema, Autocompletions } from '@/lib/schemas'
+
+async function getCompletion<T extends Autocompletions>(name: string) {
+  const autocompleteData = await fetchApi('GET', `/autocomplete/${name}`)
+  return await autocompleteSchema.validate(autocompleteData, { abortEarly: false })
+}
+
+export async function getUserCompletions(): Promise<AutocompleteItem[]> {
```
jessie-ross, line 9: Seems like we are not using these guys https://github.com/Automattic/abacus/blob/ef0251b1db348f88c99ba1ece7f90472ece45be6/lib/schemas.ts#L401-L404

What do you think about removing them?

utils/data-loading.ts
```diff
@@ -1,6 +1,13 @@
 import { noop } from 'lodash'
 import { useSnackbar } from 'notistack'
-import { DependencyList, useEffect, useRef, useState } from 'react'
+import { DependencyList, MutableRefObject, useEffect, useRef, useState } from 'react'
+
+export interface DataSourceResult<T> {
+  data: T | null
+  isLoading: boolean
+  error: Error | null
```
jessie-ross, line 8: This is not technically correct as it is `E extends Error`.

Pretty benign though and so fine for now.

jessie-ross (approved):
See comments, but other than that LGTM!

jessie-ross:
And here is a reminder to change the base before merging, something I do a lot... 🤦

api/AutocompleteApi.ts
withinboredom: I was thinking the same thing. I'll remove them in a follow-up PR after we merge all of these.

utils/data-loading.ts
withinboredom: Oh, maybe I don't understand this subtlety. It's my understanding that `E extends Error` means that `E` must be `Error` or some subtype of `Error`, and since we don't know what `E` is, exactly, we must only accept the base type or any of it's subtypes here. Anything that extends `Error` should be a valid type to pass as `error`, right?

jessie-ross: Yeah so the actual result gives `error` as `E extends Error`.

For example we have an `UnauthorizedError` that extends `Error` that could be returned, if it had an extra method or prop defined that method or prop would be hidden. E.g. `error.extraProp` wouldn't be accessible

The "correct" way to type this would be:
```
export interface DataSourceResult<T, E extends Error> {
  data: T | null
  isLoading: boolean
  error: E | null
  ...
}
```

In reality this shouldn't actually add much as typescript isn't static enough to track thrown types like Java for example might.

withinboredom: Ahhh... I see!

components/experiment-creation/ExperimentForm.tsx
withinboredom: I'll do this now, which will really clean up followup diffs.

utils/data-loading.ts
withinboredom: This complicates code that isn't worried about errors, which Error type to expect? However, your suggestion feels more "correct" than the original implementation.

So, I propose this signature:

```ts
export interface DataSourceResult<T, E extends Error = Error> {}
```

For code that won't be doing any error handling, the second type parameter can remain unspecified, while code that cares about errors can easily deal with it. What do you think?

jessie-ross: Yeah that works quite nice :)

components/experiment-creation/BasicInfo.stories.tsx
```diff
@@ -1,13 +1,20 @@
...
 export default { title: 'ExperimentCreation.Form Parts.BasicInfo' }
 
-export const FormPart = () => (
-  <MockFormik>
-    <BasicInfo />
-  </MockFormik>
-)
+export const FormPart = () => {
+  const completionBag = {
+    userCompletions: useDataSource(getUserCompletions, []),
```
jessie-ross, line 13: Might mention this again in case you missed it: could we mention that this variable isn't a symbol for data but rather a dataSource result

withinboredom: Oh, yep!

pages/experiments/[id]/wizard-edit.tsx
```diff
@@ -62,12 +63,15 @@ export default function WizardEditPage() {
   }
 
   const initialExperiment = experiment && experimentToFormData(experiment)
+  const completionBag = {
+    userCompletions: useDataSource(getUserCompletions, []),
```
Comment (jessie-ross) [inline pages/experiments/[id]/wizard-edit.tsx:67]: It looks like this should be `userCompletionDataSource` and same in `new.tsx`

jessie-ross (approved):
See last comment but other than that 🚢

pages/experiments/[id]/wizard-edit.tsx
Comment (withinboredom) [reply pages/experiments/[id]/wizard-edit.tsx]: Just fixed it :)

aaronyan:
Used to partially resolve #323

PR #349, opened 2020-09-30
Add debug query parameter to requests in debug mode

jessie-ross:
**This PR adds debug query parameter to requests in debug mode to allow the server to known when to filter requests.**

Successor of #319 

## How has this been tested?

- Automated tests that cover existing functionality
   - Checked that debug parameter was being applied in debug more and not applied in non-debug mode
   - I also checked that the filtering behaviour was as expected in my sandbox

### DIFF
--- a/api/ExperimentsApi.ts
+++ b/api/ExperimentsApi.ts
@@ -15,6 +15,7 @@ import {
   Status,
   yupPick,
 } from '@/lib/schemas'
+import { isDebugMode } from '@/utils/general'
 
 import { fetchApi } from './utils'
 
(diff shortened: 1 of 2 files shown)


yanirs:
I thought we were going to add it to `fetchApi()`. We may want to add debug options to other endpoints as we go (e.g., hidden fields).

jessie-ross:
Yeah, I found it annoying to implement it in fetchAPI and then I was thinking that less magical might be better here, so I thought I would see what you think.

yanirs:
It doesn't seem too magical to me, but I don't mind deferring it to a later point since it's only two endpoints for now.

yanirs (approved):

PR #350, opened 2020-10-01
Improve CircleCI setup

### LINKED ISSUE #89: Use npm ci Instead of npm install in CI Environment
After reading about the `npm ci` command, I think we should be using it instead of `npm install`.

See https://docs.npmjs.com/cli/ci.html for details.

### LINKED ISSUE #331: Investigate flaky CircleCI tests
We're seeing [CircleCI tests fail on unrelated code](https://github.com/Automattic/abacus/pull/328#issuecomment-694741002). Probably something worth looking into to improve the quality of life when developing PRs.

yanirs:
A bunch of improvements to our CircleCI setup:

* Use [the latest Node Orb](https://circleci.com/orbs/registry/orb/circleci/node)
* Change the Node executor version to match what we have in `.nvmrc`
* Use the `node/install-packages` step from the orb, which runs `npm ci` with caching (closes #89)
* Split the workflow into steps that can be run in parallel, which means that "rerun from failed" is now useful. :tada: As noted in https://github.com/Automattic/abacus/issues/331#issuecomment-701865831, I think that this is enough to close #331, as any remaining flakiness would be less of a pain to deal with.

## How has this been tested?

By observing [a faster and more modular CircleCI run](https://app.circleci.com/pipelines/github/Automattic/abacus/1342/workflows/23a46617-3b72-46c8-9f2c-672746eca974):

Peek 2020-10-01 14-02

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,11 +52,10 @@ jobs:
       # Reuse the workspace from the build job
       - attach_workspace:
           at: .
-      # Adapted from https://circleci.com/orbs/registry/orb/threetreeslight/puppeteer.
-      # Not using this Orb directly to prevent re-downloading Chromium which was already
-      # done in the build step.
-      # Ideally, we'd use an image/executor that already have these dependencies, but
-      # that's not supported yet.
+      # Adapted from https://circleci.com/orbs/registry/orb/threetreeslight/puppeteer. Not using this Orb directly to
+      # prevent re-downloading Chromium which was already done in the build step.
+      # We only run this step here rather than in the build job because Chrome isn't needed for the other jobs.
+      # Ideally, we'd use an image/executor that already have these dependencies, but that's not supported yet.
       # See https://discuss.circleci.com/t/new-ruby-convenience-image-public-beta/33274/6
       - run:
           name: Install Headless Chrome Dependencies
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,53 +1,127 @@
 version: 2.1
 
 orbs:
-  node: circleci/node@1.1.6
+  node: circleci/node@4.0.1
 
 jobs:
-  build-and-verify:
+  build:
```
yanirs, line 7: These name changes mean that the required checks don't pass because they never run. I'll change the branch protection rules before merging to make GitHub happy.

.circleci/config.yml
```diff
@@ -1,53 +1,127 @@
...
     executor:
       name: node/default
-      tag: '10.18.1'
+      tag: '12.18.4'
```
yanirs, line 10: Taken from `.nvmrc`. Now that I think about it, I'm unsure why we went with this version. :man_shrugging:

.circleci/config.yml
```diff
@@ -1,53 +1,127 @@
...
-                name: Unit Testing
-                command: npm run test:unit -- -- --ci --maxWorkers 2
-            - run:
-                name: Integration Testing
-                command: npm run test:integration -- --ci
-            - run:
-                name: End-to-End Testing
-                command: npm run test:e2e -- --ci
+      - node/install-packages
+      # Save workspace for subsequent jobs
```
yanirs, line 14: This was useful: https://circleci.com/docs/2.0/sample-config/

jessie-ross (approved):
LGMT and TQVM! 🚢 🚢 🚢 🚢

PR #352, opened 2020-10-01
Add event autocompletions

withinboredom:
This adds event and props autocomplete to the experiment wizard.

## How has this been tested?


components/experiment-creation/Metrics.tsx
```diff
@@ -98,7 +106,173 @@ const createMetricAssignment = (metric: MetricBare) => {
   }
 }
 
-const Metrics = ({ indexedMetrics }: { indexedMetrics: Record<number, MetricBare> }) => {
+const EventEditor = ({
```
withinboredom, line 109: This block of code is pulled from below because hooks cannot be called in a branch and must be top-level. So, I pulled it into it's own component. This bit is probably easier to view from this [commit view](https://github.com/Automattic/abacus/pull/352/commits/b2fff2b123dc552dadcfaca422e0ae927c6acffa).

components/experiment-creation/Metrics.tsx
```diff
@@ -98,7 +106,173 @@ const createMetricAssignment = (metric: MetricBare) => {
...
+    | 'exposureEventsEventNameCell'
+    | 'exposureEventsEventName'
+    | 'exposureEventsEventRemoveButton'
+    | 'exposureEventsEventPropertiesRow'
+    | 'addMetric'
+    | 'addMetricAddSymbol'
+    | 'exposureEventsEventPropertiesKey'
+    | 'exposureEventsEventPropertiesKeyAutoComplete',
+    string
+  >
```
withinboredom, line 129: I wasn't sure if this should be broken into it's own classes. It seemed simpler to do it this way, but it makes sense either way.

components/experiment-creation/Metrics.tsx
```diff
@@ -98,7 +106,173 @@ const createMetricAssignment = (metric: MetricBare) => {
...
+  eventCompletions: DataSourceResult<AutocompleteItem[]>
+  exposureEvent: EventNew
+}) => {
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const { isLoading, data: propList } = useDataSource(getPropCompletions(exposureEvent.event), [exposureEvent.event])
```
withinboredom, line 134: Can add this as a comment, but it seems pretty obvious: had to add this disable because we know the function returned by `getPropCompletions` can't rely on anything more than the string we pass to it. In fact, the error given makes no sense (unless it's trying to prevent people from creating a closure on deps in the same function, which I guess makes sense).

We also need to make the actual string a dep (vs the object) because formik passes a new object every time, even though it's exactly the same value.

api/AutocompleteApi.ts
```diff
@@ -9,3 +9,25 @@ async function getCompletion<T extends Autocompletions>(name: string) {
 export async function getUserCompletions(): Promise<AutocompleteItem[]> {
   return (await getCompletion('users')).completions
 }
+
+export async function getEventCompletions(): Promise<AutocompleteItem[]> {
+  return (await getCompletion('events')).completions
+}
+
+export function getPropCompletions(eventName: string): () => Promise<AutocompleteItem[]> {
```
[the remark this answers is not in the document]
jessie-ross: Yeah the code at this level shouldn't need to be involved with how it is used, it should just return a promise of the data rather than another function that returns a promise.

components/experiment-creation/Metrics.tsx
jessie-ross: Yeah I think the standard way of doing this is to create a function here that returns a promise, that will make the error go away and allow proper dep checking.

E.g.:
```
const { isLoading, data: propList } = useDataSource(() => getPropCompletions(exposureEvent.event), [exposureEvent.event])
```

components/experiment-creation/Metrics.tsx
```diff
@@ -98,7 +106,173 @@ const createMetricAssignment = (metric: MetricBare) => {
...
 }
 
-const Metrics = ({ indexedMetrics }: { indexedMetrics: Record<number, MetricBare> }) => {
+const EventEditor = ({
+  arrayHelpers,
+  index,
+  classes,
+  eventCompletions,
+  exposureEvent,
+}: {
```
jessie-ross, line 115: This is a reasonable midway abstraction, but if you feel like it now you should be able to get the props down to just: `{ name, onDelete }`.

That will help it on it's way to being moved to the metric edit/add form.

components/experiment-creation/Metrics.tsx
jessie-ross: 👍  Makes sense, and this is something we will want to do anyway.

jessie-ross (changes requested):
Main changes requested:
- Having `getPropsCompletions` return the data rather than a function
- Separating the classes.

See comments for more info, other than that this is looking good.

api/AutocompleteApi.ts
```diff
@@ -9,3 +9,25 @@ async function getCompletion<T extends Autocompletions>(name: string) {
...
+      return [{ name: 'Enter an event name', value: '' }]
+    }
+
+    try {
+      const apiResponse = await eventDetailsSchema.validate(await fetchApi('GET', `/autocomplete/events/${eventName}`))
+      return apiResponse.props.map((p) => ({
+        name: p.name,
+        value: p.name,
+      }))
+    } catch (er) {
```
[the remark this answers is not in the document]
withinboredom: > but we are conflating any error here as missing data and not reporting actual errors to the user.

I made this decision after thinking about the types of errors a user may encounter:
1. unauthorized: they need to login but there should be (hopefully) better systems to detect this and inform the user. Perhaps we need to rethrow here so that system catches it.
2. disconnected: there's no point in displaying an error about that since we don't prevent the user from entering whatever value they want and we don't have any way to re-fetch once they do reconnect. Also, they likely wouldn't have been able to select an event in the first place and the error would be reported there.
3. server error (500): things will get logged on the server. There's nothing the user can do except notify us, but we should already be monitoring those logs. They can still enter the props, albeit, manually.
4. not found: we cover this case.

There's no disaster by ignoring most errors and the user can/should be able to continue.

api/AutocompleteApi.ts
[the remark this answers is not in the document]
withinboredom: I think that makes sense with the changes [you suggested here](https://github.com/Automattic/abacus/pull/352#discussion_r498392652)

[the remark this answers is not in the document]
jessie-ross: Yep that's what I meant, it would probably help if I had explicitly linked the two.

jessie-ross:
I'll merge this guy once/if CRA manages to land

jessie-ross:
I've just merged the master into here so @withinboredom doesn't have to deal with the restructure.

jessie-ross:
@withinboredom Is this ready to merge?

withinboredom:
@jessie-ross Still blocked by your last review, feel free to change it and merge away :p :D

jessie-ross:
Not sure if this needs a second review, I have moved some of the UI code out of the API, fixed the tests as a result and fixed some warnings, and removed the unused schemas.

jessie-ross:
@withinboredom I am going to accept it, if you think it is good feel free to merge.

jessie-ross (approved):

aaronyan:
Helps resolve #323!

PR #357, opened 2020-10-08
React Router: Part 1

jessie-ross:

**This PR is the first of two migrating this app from NextJS to CRA, this one changes our underlying routing to use `react router`.**
- Move content in `/pages` into `/pages-real` I have only changed the top of some files relating to `useRouter`.
- Create new routing system at `/Routes.tsx`
- Makes all `/pages/*` run `/Routes.tsx`

See #212 for the full story.

## How has this been tested?


jessie-ross:
I am not sure if I am so happy with the naming scheme but so long as the structure is good to make this easier it will be better to change them in a PR after this series.

jessie-ross:
Ignore this for now, I am just keeping it up to date

jessie-ross:
I have split this PR into #383 and #384 for easier reviewing.

PR #359, opened 2020-10-08
CRA Part 2: The big restructure

jessie-ross:

**This PR is the second of three migrating this app from NextJS to CRA, this one restructures our files.**
- CRA supports absolute imports but not path aliasing (the import from `@/...` part)
- This PR _only_ moves source files to `/src` and changes imports from `@/` to `src/` and updates 3 config files
- Using `src/` references is functionally the same as using `@/`
- Unfortunately the imports are sorted together by eslint :/

Part of a series on #212, the last one should be quite reasonable.

## How has this been tested?


jessie-ross:
The integration tests should pass but it looks like swagger is timing out, likely just from rate limiting so should be good to retry  in a bit.

withinboredom (commented):
Did we mean to include `.DS_Store`?

jessie-ross:
> Did we mean to include .DS_Store?

Nope!

jessie-ross:
This is no longer needed thanks to #364

PR #362, opened 2020-10-09
CRA Part 3: Mostly working with CRA

jessie-ross:

**This PR is the third of N migrating this app from NextJS to CRA, this one adds a working CRA base.**
- Fixes some typescript quirks
- Looks like quite a few `package-lock.json` versions have changed which might be causing the above issues.

## How has this been tested?

This PR doesn't intend to pass tests, but serves to split this PR from the next.
The next PR will pass and should get merged into this PR before this PR merges.


.gitignore
```diff
@@ -41,7 +41,6 @@ typings/
 .yarn-integrity
 
 # dotenv environment variables file
```
yanirs, line 43: As noted on Slack, last time this came up [we followed the dotenv FAQ](https://github.com/motdotla/dotenv#should-i-commit-my-env-file) that says not to commit `.env` files, but I see that [CRA says they should be committed](https://create-react-app.dev/docs/adding-custom-environment-variables/#adding-development-environment-variables-in-env): "`.env` files should be checked into source control (with the exclusion of `.env*.local`)." I'm to go with that, but let's adjust this file accordingly and only ignore `.env*.local`.

package.json
```diff
@@ -48,10 +49,9 @@
     "@package @storebook/react": "Requires `babel-loader`."
   },
   "devDependencies": {
-    "@next/bundle-analyzer": "^9.3.6",
     "@storybook/preset-scss": "^1.0.2",
     "@storybook/preset-typescript": "^3.0.0",
-    "@storybook/react": "^6.0.21",
+    "@storybook/react": "^5.3.21",
```
yanirs, line 54: Why was this downgraded?

package.json
```diff
@@ -73,7 +73,6 @@
     "cross-env": "^7.0.2",
     "css-mediaquery": "^0.1.2",
     "dotenv": "^8.2.0",
-    "eslint": "^7.9.0",
```
yanirs, line 76: Why were this and `jest` removed?

src/lib/schemas.ts
```diff
@@ -93,14 +93,23 @@ export const metricFullSchema = metricBareSchema
   .defined()
   .camelCase()
   .test('event-params-required', 'Event Params is required and must be valid JSON.', (metricFull) => {
+    if (!metricFull) {
```
yanirs, line 96: Why were these added?

src/utils/material-table.ts
```diff
@@ -1,10 +1,10 @@
-import { Options } from 'material-table'
 import theme from 'src/styles/theme'
 
 /**
  * Default table options for dynamic MaterialTable instances.
  */
-export const defaultTableOptions: Options = {
+// @ts-ignore -- Options now requires a type argument which we don't have
```
yanirs, line 6: Why don't we have it?

src/lib/schemas.ts
```diff
@@ -352,16 +370,21 @@ export enum RecommendationWarning {
   WideCi = 'wide_ci',
 }
 
+export interface Recommendation {
```
yanirs, line 373: Why are these changes needed?

yanirs (commented):
It's very unclear why many of the changes here are needed, but I'll wait for the full stack.

Issue #365, opened 2020-10-14
Use id-slugs instead of ids


jessie-ross:
Currently we use id's in our URLs, this is not ideal.

A simple backwards and forwards compatible way to deal with this is to use "id-slugs" like this:
```/experiments/15-experiment-name/overview```

instead of the current:
```/experiments/15/overview```

Where when we parse the idSlug we just ignore anything after the id. 
This has the benefit of allowing our links to be permalinks and work even through name changes.

jessie-ross:
This is another issue that Next router currently prevents being done easily, in particular next causes experiment loading to need to happen after ui loads which means that UI links won't have access to the experiment name.

#357, #366 would fix this.

Linking also to #212

PR #366, opened 2020-10-14
React Router: Part 2: Use react router links

### LINKED ISSUE #294: Speeding up experiment fetches
This is a small quality of life improvement for experimenters. However, we could use [`React Helmet`](https://github.com/nfl/react-helmet) to add a `<link  rel="prefetch"` to the head to preload the api request for their latest experiment when landing on the "home page". It probably needs some experimenting to see if it actually works and it would probably be a good idea to instrument the application first.

jessie-ross:
**This PR uses react router for links giving much better perceived performance.**

This is part of breaking apart the CRA changes, I thought I should PR this part because it fixes some very annoying issues for little effort. Best to test it yourself and see - just try navigating around.

This also makes it simple enough to add in idSlugs #365.

I believe this resolves #294 

## How has this been tested?

### DIFF
--- a/src/components/ExperimentPageView.test.tsx
+++ b/src/components/ExperimentPageView.test.tsx
@@ -63,7 +63,7 @@ const renderExperimentPageView = async ({ experiment: experimentOverrides = {} }
  * - null: not on screen
  */
 const getButtonStates = () => {
-  const editInWizard = screen.getByRole('link', { name: /Edit In Wizard/ })
+  const editInWizard = screen.getByRole('button', { name: /Edit In Wizard/ })
   const run = screen.getByRole('button', { name: /Run/ })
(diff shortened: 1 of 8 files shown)


withinboredom:
@jessie-ross, does this need review still? Just wondering because I wasn’t sure whether we are still doing the CRA stuff.

jessie-ross:
So this is a part of the argument for CRA see pbmo2S-y0-p2#comment-1233, demonstrating using a better router. 

At the same time it is also a discrete change that could give us some benefits but is "unclean code" if we aren't moving away from NextJS. It is basically done, just needs some tweaks to have e2e working.

I think it is worth taking a look at the preview: https://abacus-git-add-react-router-links.a8c.vercel.app
And if you are interesting running on production data.

jessie-ross:
Replaced by #385

PR #369, opened 2020-10-14
Improve results page

### LINKED ISSUE #355: Trend plots are unclear when there are results from a single day
Looking at https://abacus-git-add-participant-graph.a8c.vercel.app/experiments/1/results, it's unclear that experimenters need to hover over the graphs to see the results. We can fix it by `lines+markers` for the participant plot and hiding the other plots when there's only one day of results, but there are a bunch of other options. Thoughts @jessie-ross?

jessie-ross:
**This PR improves on the top level results display.**

I felt something wasn't quite right here and at the same time I wanted to get the total and not-final data in. Then I got inspired by Storytelling with Data, why not just put the figures on the page.

I also add in the `lines+markers` to partially fix #355. 
Hiding the graphs will come next.

See the latest image on slack: p1602687608015800-slack-G01BQ5WMR

## How has this been tested?

### DIFF
--- a/src/components/experiment-results/CondensedLatestAnalyses.tsx
+++ b/src/components/experiment-results/CondensedLatestAnalyses.tsx
@@ -9,6 +9,7 @@ import {
   TableContainer,
   TableRow,
   Theme,
+  Typography,
   useTheme,
 } from '@material-ui/core'
 import clsx from 'clsx'
@@ -33,6 +34,7 @@ import {
 import * as Variations from 'src/lib/variations'
 import * as Visualizations from 'src/lib/visualizations'
 import { createStaticTableOptions } from 'src/utils/material-table'
+import { formatIsoDate } from 'src/utils/time'
 
 import { MetricAssignmentAnalysesData } from './ExperimentResults'
 import RecommendationString from './RecommendationString'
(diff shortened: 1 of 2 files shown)


aaronyan (approved):
LGTM! Nice addition of some summary stats. I really like seeing the total number of users and the % of users finalized so clearly.

The % of users finalized may need an * or something to indicate that it's based on the primary metric. Other metrics may have different percentages because the attribution window may be different.

And I think we can keep the strategy in for now. But more likely than not, folks won't need to know or care about the strategy. I can't think of anything to put it in its place, so we can leave it in.

yanirs (changes requested):
> The % of users finalized may need an * or something to indicate that it's based on the primary metric. Other metrics may have different percentages because the attribution window may be different.

Yeah, I think it'd be better to drop it as it's too confusing.

> And I think we can keep the strategy in for now. But more likely than not, folks won't need to know or care about the strategy. I can't think of anything to put it in its place, so we can leave it in.

Yeah, I think we're better off leaving it out.

yanirs:
More details:
- The `finalized` figure is only for the primary metric. If we want a top-level `finalized` percentage, it should be based on the metric where it's lowest.
- Experimenters can't see other strategies. As far as they're concerned, there's only one strategy, and we will keep it this way as discrepancies between strategies need to be analysed manually.
- `X% of participants are finalized` is hard to understand. If we keep a percentage, we should be more specific on what this means. Also, % final isn't great to highlight for an experiment that's still running.
- The mixing of font sizes and boldface use makes it too busy in my opinion.
- If we want to highlight stuff at the top, I'd add the recommendation for the primary metric and whether the experiment is still running.

jessie-ross:
Ok, I might shelve this PR until later.

jessie-ross:
@yanirs I am going to have another go at this tomorrow.

jessie-ross:
Ok I have have given Experiment Results a general look over and this is what I am thinking of implementing (probably in separate PRs):
- More integration of Metrics into the MetricAssignments
   - The min diff, change expected, description etc
- Rest of #419s improvements
- Top Level stats panel
   - Participants as max over MetricAssignments
   - Stop/Continue indicator #419 
   - Primary metric Recommendation
   - (Debug Mode) Select Box for changing the strategy in use.

Future PR I might mention here:
- Traffic light panel with Red/Orange/Green, tooltip on highlight
   - SRM at p < 0.005 / p < 0.05 / p > 0.05
   - Spammer SRM at p < 0.005 / p < 0.05 / p > 0.05
   - Crossovers, n > 5% , n > 1%, n < 1%
   - Guardrail Metrics: practical significance / statistical significance / no significant change
- Advanced Mode (For current debug level features)
   - Button in the top level
   - Dialog with detailed warning and "Go Back" "Continue"

yanirs:
Yeah, as noted on #419, I think that some of these details are best hashed out in PRs, so separate PRs are the way to go.

For the traffic light, I think it's better to go with green/red for things like SRMs, as anything that looks suspicious is worth a deep look. Orange can be reserved for cases where it's more of a judgement call, like wide CIs or experiment duration warnings.

jessie-ross:
Replaced by #454

PR #373, opened 2020-10-27
Tags CRUD Base - Metrics Files copy

jessie-ross:
**This PR just copies the metrics files to act as a base for #370 so the diffs display better.**


jessie-ross:
@aaronyan This just needs a token review, it is exactly the same code as in #370 which has been approved.

aaronyan (approved):

withinboredom (approved):
I'd be the first to argue that this PR doesn't need reviewing since the PR upstream is reviewed (can just change the base to master and merge), but :rubber-stamp:ing this one :)

Issue #361, opened 2020-10-09
Add exclusion tag assignment to the creation/full-edit form


yanirs:
Once we have the tag API endpoints defined in Swagger,<sup>*</sup> we can extend the experiment creation/full-edit form to support assignment of exclusion tags (`namespace='exclusion_group'`) in debug-only mode. 

<sup>*</sup> We may want to hold off until the assignment API is fully implemented.

jessie-ross:
Posting this here after mistakenly putting it in #360.

**Here are my thoughts on this:**

I see three places where we will want to introduce tags:

- Experiment Creation Form: Basic Info
- Experiment: General Panel
- Experiment: General Panel: Edit

And for new UI components:
- A custom tag chip for display tags as a tag is two pieces of data, the namespace and the name.
- Tag input component: using our existing autocomplete but customising the rendering of the auto-complete item (using `renderOption`) as a tag chip. Seems like we just need to allow in custom `getOptionSelected` and `getOptionValue`.

There will also need to be some minor schema changes but that is all straightforward.

jessie-ross:
Yanir mentioned about the General Panel Edit in https://github.com/Automattic/abacus/issues/360#issuecomment-707564381:

> We won't need this initially, as the only tags are exclusion group tags that must be assigned ahead of the experiment (as part of creation or when editing staging experiments).

This is interesting, I didn't think of this and it complicates the tag behaviour quite a bit. For me this is a point towards using a seperate metadata field in the future.

yanirs:
> This is interesting, I didn't think of this and it complicates the tag behaviour quite a bit. For me this is a point towards using a seperate metadata field in the future.

Why does it complicate things more than any other field that can be changed on new experiments but not later (like segments)?

jessie-ross:
This is a bit different to segments as we have one field that behaves in seperate ways depending on what we are setting. 
An equivalent for segments would be if locales and countries behaved separately and could be edited at different times.

yanirs:
Ah, I see. Initially we're probably only going to have exclusion group tags (and maybe `internal/code_ready`), so I think it's fine for now. If we add tags that can be applied to non-staging experiments, we can figure out how to best approach it, as it really depends on the semantics of the tags. As noted before, I think we should avoid overusing tags.

[...]

PR #374, opened 2020-10-28
Add exclusion tags display

jessie-ross:
**This PR displays experiment exclusion tags.**

97146725-50583c00-17a3-11eb-9620-d4f4051fdae9

Part of #361 

## How has this been tested?

(Swagger needs to be updated for you to see the exclusion groups: 474-gh-experimentation-platform)

### DIFF
--- a/src/components/ExperimentPageView.tsx
+++ b/src/components/ExperimentPageView.tsx
@@ -212,7 +212,7 @@ export default function ExperimentPageView({
           </div>
         </div>
         {isLoading && <LinearProgress />}
-        {experiment && metrics && segments && analyses && (
+        {experiment && metrics && segments && analyses && tags && (
           <>
             {view === ExperimentView.Overview && (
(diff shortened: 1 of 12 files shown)


src/components/ExperimentPageView.tsx
```diff
@@ -208,10 +212,10 @@ export default function ExperimentPageView({
           </div>
         </div>
         {isLoading && <LinearProgress />}
-        {experiment && metrics && segments && analyses && (
+        {experiment && metrics && segments && analyses && tags && (
```
withinboredom, line 215: upon adding a 5<sup>th</sup> `&&`, it may be valuable to break into a function or a variable that self-documents this.

withinboredom (approved):

src/components/ExperimentPageView.tsx
jessie-ross: Unfortunately breaking this into a variable means this stops working as a type-guard, maybe with a future version of typescript.

PR #375, opened 2020-10-29
Add multi-entry autocomplete

jessie-ross:
**This PR allows Autocomplete to have multi entries.**
- In doing this I simplified the component by using separate inner data from outer.
- Fixed the existing tests to not rely on implementation details as much.

## How has this been tested?


src/components/Autocomplete.tsx
```diff
@@ -33,30 +18,86 @@ export const autocompleteInputProps = (params: AutocompleteRenderInputParams, lo
...
   } = props
 
-  // Can't destructure this as it is of type any
-  const value = props.field.value as unknown
+  // ## Translating OuterValues (value | value[]) <-> InnerValues (AutocompleteItem | AutocompleteItem[])
+  //
+  // Typescript and esline make this look much more complicated than it is, we are just keeping the form value type
+  // in the form (string | number, and array version for multiple), and inside here we are using Autocomplete.
+  //
+  // We transform on the way in and on the way out.
```
jessie-ross, line 40: These inner-value/outer-value transforms are super useful and clean ways to make custom input components (it is also used for segments).

We might even want to abstract this out as a hook.

aaronyan:
Adding @withinboredom as a reviewer here, since they will likely be more equipped to review this PR! 😄

src/components/ExperimentPageView.tsx
```diff
@@ -208,10 +212,10 @@ export default function ExperimentPageView({
           </div>
         </div>
         {isLoading && <LinearProgress />}
-        {experiment && metrics && segments && analyses && (
+        {experiment && metrics && segments && analyses && tags && (
```
withinboredom, line 215: Can we replace this entire chain with `!isLoading`?

src/components/ExperimentPageView.tsx
```diff
@@ -128,13 +129,16 @@ export default function ExperimentPageView({
...
+  useDataLoadingError(tagsError, 'Tags')
+
   const { isLoading: analysesIsLoading, data: analyses, error: analysesError } = useDataSource(
     () => (experimentId ? AnalysesApi.findByExperimentId(experimentId) : createUnresolvingPromise<Analysis[]>()),
     [experimentId],
   )
   useDataLoadingError(analysesError, 'Analyses')
 
-  const isLoading = or(experimentIsLoading, metricsIsLoading, segmentsIsLoading, analysesIsLoading)
+  const isLoading = or(experimentIsLoading, metricsIsLoading, segmentsIsLoading, tagsIsLoading, analysesIsLoading)
```
withinboredom, line 141: Outside the scope of this PR, but we may want to combine these API requests into a single API call at some point, instead of hammering the API with up to 5 API calls.

src/components/Autocomplete.tsx
```diff
@@ -33,30 +18,86 @@ export const autocompleteInputProps = (params: AutocompleteRenderInputParams, lo
...
+
+  type InnerValue = Multiple extends true ? InnerValueSingle[] : InnerValueSingle
+  type OuterValue = Multiple extends true ? OuterValueSingle[] : OuterValueSingle
```
withinboredom, line 43: Clever! I like it!

withinboredom (approved):

src/components/ExperimentPageView.tsx
jessie-ross: This acts as a typeguard, unfortunately `!isLoading` won't work for it.

src/components/Autocomplete.tsx
```diff
@@ -33,30 +18,86 @@ export const autocompleteInputProps = (params: AutocompleteRenderInputParams, lo
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- React strangely thinks the types should be dependecies, smh...
+    [],
+  )
 
+  const outerValue = props.field.value as OuterValue
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore; Typescript can't quite work this:
+  const innerValue: InnerValue = multiple
+    ? outerValue
+      ? (outerValue as OuterValueSingle[]).map(outerSingleToInnerSingle)
```
jessie-ross, line 60: Prettier thinks this is better without as many parens 🤦

src/components/ExperimentPageView.tsx
withinboredom: Bummer

PR #376, opened 2020-11-02
Add exclusion group tag editing to Experiment Wizard

jessie-ross:
~~**DO NOT MERGE YET**: This PR needs the respective backend diff to be merged, this is because it sends an empty array for exclusion groups by default, and the backend doesn't allow extensions to the experiment type being sent.~~

**This PR add exclusion group tags to the experiment wizard.**
- A lot of files have changed but most of them are quite small.
- I have decided to put the tags into the autocomplete system.
- With the parent #375, the front-end is basically also ready to add general tags if needed.
- It is not pretty just being tacked on to the end of "Audience" I will add a P2 soon about splitting the sections further.

Screen Shot 2020-11-02 at 11 55 11 pm

## How has this been tested?

- Manual testing with production data


src/pages/experiments/new.tsx
```diff
@@ -47,7 +48,28 @@ const ExperimentsNewPage = function (): JSX.Element {
...
+    userCompletionDataSource,
+    eventCompletionDataSource,
+    exclusionGroupCompletionDataSource,
+  }
+
+  const isLoading = or(
+    metricsIsLoading,
+    segmentsIsLoading,
+    ...Object.values(completionBag).map((dataSource) => dataSource.isLoading),
+  )
```
withinboredom, line 72: This appears to be duplicated from above. It might be out of scope of this PR, but might be worth consolidating these two components at some point :)

withinboredom (approved):

src/pages/experiments/new.tsx
jessie-ross: Yeah I am thinking so too, this is also easier with the CRA changes :)

jessie-ross: I am going to do so after this PR is merged.

.circleci/config.yml
```diff
@@ -96,7 +96,7 @@ jobs:
           command: npm run test:integration -- --ci
       - run:
           name: End-to-End Testing
-          command: npm run test:e2e -- --ci
+          command: GENERATE_SOURCEMAP=false npm run test:e2e -- --ci
```
jessie-ross, line 99: @yanirs This fixes CircleCI memory issues for now.

Linking #397

src/lib/schemas.ts
```diff
@@ -357,6 +361,8 @@ export const experimentFullNewOutboundSchema = experimentFullNewSchema
           props: event.props ? _.fromPairs(event.props.map(({ key, value }) => [key, value])) : undefined,
         }),
       ),
+      // The backend doesn't support exclusion groups (it will crash if this is an empty array) so we disable it here:
+      exclusion_group_tag_ids: undefined,
```
jessie-ross, line 365: I disable `exclusion_group_tag_ids` from being sent for now so that this PR can be merged.

.circleci/config.yml
yanirs: Thanks @jessie-ross! I added it to #401 to make the build pass there. I also removed the resource class setting since it's ignored.

yanirs:
I just looked at it on Vercel and I have a question: Why is `Exclusion Groups` marked as required? We want to it be optional.

jessie-ross:
> Why is Exclusion Groups marked as required?

I am assuming you are referring to this:
https://github.com/Automattic/abacus/blob/a980071b934db361dfcea9f18cf2eac6acbc3eee/src/components/experiment-creation/Audience.tsx#L310

Yeah this is in there by accident. I think it might be an accessibility error it but doesn't actually affect validation of the form which is done using:

https://github.com/Automattic/abacus/blob/a980071b934db361dfcea9f18cf2eac6acbc3eee/src/lib/schemas.ts#L287-L302

Where `exclusionGroupTagIds` isn't defined. 

I'll remove it from the PR.

yanirs:
@jessie-ross yeah, that's what I meant. I was just going by the UI without looking at the code. It might also be be worth adding the word _optional_ somewhere. Maybe `Optional: Add this experiment to a mutually exclusive experiment group.`

Anyway, I'll leave it to @withinboredom to do a final pass/test.

src/components/experiment-creation/Audience.tsx
```diff
@@ -276,6 +283,40 @@ const Audience = ({
...
+            <FormHelperText>Optionally add this experiment to a mutually exclusive experiment group.</FormHelperText>
+            <br />
+            <Field
+              component={AbacusAutocomplete}
+              name='experiment.exclusionGroupTagIds'
+              id='experiment.exclusionGroupTagIds'
+              fullWidth
+              options={
+                // istanbul ignore next; trivial
+                completionBag.exclusionGroupCompletionDataSource.data ?? []
```
withinboredom, line 299: We should probably still test the `null` case, otherwise removing `?? []` won't fail the test but will elicit warnings (IIRC) which may not be seen.

src/components/experiment-creation/ExperimentForm.tsx
```diff
@@ -5,16 +5,22 @@ import _ from 'lodash'
...
 import BasicInfo from './BasicInfo'
 import Beginning from './Beginning'
 import Metrics from './Metrics'
 
+export interface ExperimentFormCompletionBag {
+  userCompletionDataSource: DataSourceResult<AutocompleteItem[]>
+  eventCompletionDataSource: DataSourceResult<AutocompleteItem[]>
+  exclusionGroupCompletionDataSource: DataSourceResult<AutocompleteItem[]>
+}
+
```
withinboredom, line 23: Just out of curiosity, why did we move this here vs moving to its own file?

withinboredom (approved):
There is a minor issue and a curiosity, but LGTM.

src/components/experiment-creation/ExperimentForm.tsx
jessie-ross: `ExperimentFormCompletionBag` was with `AutocompleteApi` before.

The first part of it is that it is arguably UI code and doesn't belong in the `Api` layer (`ExperimentForm` is in its name). We want to keep this layer as simple as possible. These collections could also be fairly arbitrary, we might have several other interfaces containing permutations of 5 or so autocomplete endpoints.

The second part is it is pretty specifically tied to the Experiment Form, that is the "source" of the river that is this data structure flowing down, so it makes sense for it to belong there.

src/components/experiment-creation/Audience.tsx
jessie-ross: I don't think testing the null case is useful, it is pretty predictable what happens, I have tried it out and don't see any warnings.

Although looking at this again this is a case of full coverage getting in the way of real testing, for example a real problem here might be that a value is set (in the case of this form being updated) but the exclusion groups never load. This might be a problem if the user wants to change it and loses their original choice. In this case we do have the loading indicator and loading errors, as well as the user now and pre-experiment reviewing the checks as mitigating factors, which I think is enough. I would love to do some kind of proper risk analysis with the App.

PR #384, opened 2020-11-11
Use React Router (Redo): Pt 2/3: Use React-Router Page Params into /real-pages

jessie-ross:
**This PR alters pages to use react-router's page params system instead of NextJS.**

- Also clean up the parameter validation code by using yup.
- Remove the no longer needed `qc-to_int` and `qc-to_bool` libs

Part of a series on #212 

## How has this been tested?

### DIFF
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -1,9 +1,8 @@
 require('dotenv').config()
-const { toBool } = require('qc-to_bool')
 
 module.exports = {
   launch: {
-    headless: toBool(process.env.PUPPETEER_HEADLESS, true),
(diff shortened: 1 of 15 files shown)


yanirs:
@jessie-ross this looks good, but it'd be good to know what's supposed to work and not work with this PR. For example, navigation between experiment views doesn't seem to work. Is it coming in part 3?

jessie-ross:
> this looks good, but it'd be good to know what's supposed to work and not work with this PR. For example, navigation between experiment views doesn't seem to work. Is it coming in part 3?

Yep part three.

I intended to PR it immediately after this but got stuck in some more e2e testing issues.

yanirs (approved):
:tada:

PR #391, opened 2020-11-12
Add id-slugs in urls for experiments (permalinks with names in it)

### LINKED ISSUE #365: Use id-slugs instead of ids

jessie-ross:
**This PR adds id-slugs for experiments.**

Now we use urls like:  
`http://a8c-abacus-local:3000/experiments/20033-explat-test-aa-weekly-calypso-2020-week-46/overview`

This is a cinch with CRA 😄 

Resolves #365 

## How has this been tested?

### DIFF
--- a/src/utils/general.ts
+++ b/src/utils/general.ts
@@ -1,3 +1,5 @@
+import { idSchema } from 'src/lib/schemas'
+
 // TODO: Consider using a more extensive lib like Rambda that covers these
 /**
  * Coerces values to booleans and performs an `or` operation across them.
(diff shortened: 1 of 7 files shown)


src/utils/general.ts
```diff
@@ -30,3 +32,35 @@ export function toggleDebugMode(): void {
...
+ * The idea here is that we can use these for permalinks because even if the name changes
+ * all we care about is the ID.
+ *
+ * @param idSlug string The idSlug
+ * @returns integer The id from the id slug
+ */
+export function parseIdSlug(idSlug: string): number {
+  const results = /^\d+/.exec(idSlug)
+  const id = results && idSchema.defined().validateSync(results[0])
+  // istanbul ignore next; Should not occur
```
withinboredom, line 49: Since we test this functionality above, we can probably remove this

withinboredom (approved):
LGTM

PR #394, opened 2020-11-13
Fix docker port

jessie-ross:
**I changed the docker-start port to `3000` thinking I was changing the internal port, but it should actually be `8888` to match how it was before. This PR fixes that.**

## How has this been tested?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,8 @@ jobs:
     executor:
       name: node/default
       tag: '12.18.4'
+    # See https://circleci.com/docs/2.0/configuration-reference/#resource_class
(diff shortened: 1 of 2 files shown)


yanirs (approved):

jessie-ross:
@yanirs I have also slipped a CircleCI resource increase in here but it won't take effect until an admin on CircleCI upgrades us:

Screen Shot 2020-11-13 at 3 44 07 pm

yanirs:
Thanks! It's probably worth finding out who to ping to get us upgraded, as it should be faster and cheaper than trying to reduce the build's memory consumption. :slightly_smiling_face:

PR #395, opened 2020-11-13
Stop reloading data for overview tab

jessie-ross:
**This PR fixes the link for the overview tab, stopping it from causing the experiment data to reload.**

Since there is now a redirect from `/experiments/[id]` to `/experiments/[id]/overview`, clicking on the overview tab link would cause a refresh, this PR fixes that.

## How has this been tested?


aaronyan (approved):
Nice! LGTM.

PR #405, opened 2020-11-20
Relax manual analysis criteria to only check chosenVariationId

aaronyan:
While looking at some of our experiment results, I kept getting frustrated with the manual analysis required. In with hopefully a quick fix! :)

## How has this been tested?

- [x] Manual testing with production data (locally)

### DIFF
--- a/src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
+++ b/src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
@@ -85,7 +85,7 @@ export default function CondensedLatestAnalyses({
         .map(
           (analyses) =>
             //  istanbul ignore next; We don't need to test empty analyses as we filter out all undefined values
-            _.last(analyses)?.recommendation,
+            _.last(analyses)?.recommendation.chosenVariationId,
         )
         .filter((recommendation) => !!recommendation)
       const uniqueRecommendations = _.uniq(recommendations.map((recommendation) => JSON.stringify(recommendation)))
--- a/src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
+++ b/src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
@@ -85,10 +85,10 @@ export default function CondensedLatestAnalyses({
         .map(
           (analyses) =>
             //  istanbul ignore next; We don't need to test empty analyses as we filter out all undefined values
-            _.last(analyses)?.recommendation.chosenVariationId,
+            _.last(analyses)?.recommendation,
         )
         .filter((recommendation) => !!recommendation)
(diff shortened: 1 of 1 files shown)


src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
```diff
@@ -85,7 +85,7 @@ export default function CondensedLatestAnalyses({
         .map(
           (analyses) =>
             //  istanbul ignore next; We don't need to test empty analyses as we filter out all undefined values
-            _.last(analyses)?.recommendation,
+            _.last(analyses)?.recommendation.chosenVariationId,
         )
         .filter((recommendation) => !!recommendation)
       const uniqueRecommendations = _.uniq(recommendations.map((recommendation) => JSON.stringify(recommendation)))
```
aaronyan, line 91: Should we change the name throughout this file from `recommendations ` to `recommendedVariationIds`?

jessie-ross: Yep, I think this is needed to make this code clear.

We can also remove the `JSON.stringify` mapping.

yanirs, line 91: In addition to the naming, it looks like the `filter()` call won't work as expected. How about we change it to something like this?

```suggestion
            _.last(analyses)?.recommendation,
        )
        .filter((recommendation) => !!recommendation)
      const recommendationConflict = _.uniq(recommendations.map((recommendation) => recommendation.chosenVariationId)) > 1
```

That is, leave the `recommendations` array as it is, and change the `recommendationConflict` variable.

jessie-ross: I don't see why filter won't work?
Also before this PR goes out `chosenVariationId` will be optional chained, in case that is something you are missing?

jessie-ross: Ahh I didn't realise `chosenVariationId` set to `null` signified something.

aaronyan:
Changed to 'WIP' and I'll save this for next week to avoid further distraction. :) Thanks for the eyes folks!

jessie-ross:
This was so close to being done I thought I might just gently nudge it across the line 😄

yanirs (approved):
Pushed https://github.com/Automattic/abacus/pull/405/commits/e58581b6a858fb6eacc4be0fa51d8311d1ec533d to make it a true team effort. :smile: 

@jessie-ross you can merge if it looks good to you.

jessie-ross (approved):

aaronyan:
Haha thank you @jessie-ross and @yanirs for helping make this happen! 😄

PR #408, opened 2020-11-27
Add support for email and pipe platforms

yanirs:
Add `email` and `pipe` as platforms to the `Platform` enum and `PlatformToHuman` mapping.

**Note**: We can hold off on merging this until the backend supports the new platforms.

## How has this been tested?
- Manual testing with mock data (locally or on Vercel): Checked that the new platforms appear in the wizard

### DIFF
--- a/src/lib/experiments.ts
+++ b/src/lib/experiments.ts
@@ -39,8 +39,8 @@ export function getDefaultAnalysisStrategy(experiment: ExperimentFull): Analysis
 }
 
 export const PlatformToHuman: Record<Platform, string> = {
-  [Platform.Calypso]: 'Calypso',
-  [Platform.Email]: 'Email (Guides)',
-  [Platform.Pipe]: 'pipe (machine learning pipeline)',
-  [Platform.Wpcom]: 'WordPress.com Backend',
+  [Platform.Calypso]: 'Calypso front-end',
+  [Platform.Email]: 'Guides and other email systems',
+  [Platform.Pipe]: 'Machine learning pipeline',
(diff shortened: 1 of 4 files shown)


src/lib/experiments.ts
```diff
@@ -39,6 +39,8 @@ export function getDefaultAnalysisStrategy(experiment: ExperimentFull): Analysis
 }
 
 export const PlatformToHuman: Record<Platform, string> = {
-  [Platform.Wpcom]: 'WordPress.com Backend',
   [Platform.Calypso]: 'Calypso',
+  [Platform.Email]: 'Email (Guides)',
+  [Platform.Pipe]: 'pipe (machine learning pipeline)',
```
withinboredom, line 44: It's very odd to see this lower-case. Yes, I'm aware it's supposed to be lower case, but it looks like a typo. Perhaps consider making "Email" lower case.

withinboredom (approved):

src/lib/experiments.ts
yanirs: Maybe we can just change it to uppercase everywhere to make @demetd happy? :trollface: 

I realised that we're only using this mapping when the experiment is created, so it's worth specifying the constant name in the dropdown. This is what it looks like now:

Peek 2020-12-01 11-18

What do you think?

demetd: > Maybe we can just change it to uppercase everywhere to make @demetd happy? :trollface:

I prefer PiPe /piː-pæ/

withinboredom: That looks better!

PR #410, opened 2020-11-27
Enable saving exclusion groups

jessie-ross:
**_Do not merge until the backend supports saving exclusion groups or things will break._**
This PR enables saving exclusion groups.

## How has this been tested?

- Manual testing with production data (locally) against the updated endpoints

### DIFF
--- a/src/lib/schemas.ts
+++ b/src/lib/schemas.ts
@@ -361,8 +361,8 @@ export const experimentFullNewOutboundSchema = experimentFullNewSchema
           props: event.props ? _.fromPairs(event.props.map(({ key, value }) => [key, value])) : undefined,
         }),
       ),
-      // The backend doesn't support exclusion groups (it will crash if this is an empty array) so we disable it here:
-      exclusion_group_tag_ids: undefined,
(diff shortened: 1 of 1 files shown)


src/lib/schemas.ts
```diff
@@ -361,8 +361,8 @@ export const experimentFullNewOutboundSchema = experimentFullNewSchema
           props: event.props ? _.fromPairs(event.props.map(({ key, value }) => [key, value])) : undefined,
         }),
       ),
-      // The backend doesn't support exclusion groups (it will crash if this is an empty array) so we disable it here:
-      exclusion_group_tag_ids: undefined,
+      // // The backend doesn't support exclusion groups (it will crash if this is an empty array) so we disable it here:
```
withinboredom, line 364: Probably want to move the double `//`. In all actuality, you should probably just remove these two lines instead of commenting them out.

withinboredom (approved):

src/lib/schemas.ts
jessie-ross: 🤦 Yep it should be removed, don't know why I didn't delete this.

PR #411, opened 2020-11-30
Implement experiment cloning

### LINKED ISSUE #380: Implement copy experiments
When building the three experiments for the [Monthly Pricing experiment](https://growthchannels.wordpress.com/2020/09/29/project-thread-monthly-pricing/), it would have been nice to copy an experiment and have it loaded into the Experiment Creation Wizard. The biggest benefit is being able to copy over the metric assignments!

This feature might also be helpful for experimenters getting started with Abacus and ExPlat. 😄 

## The "Copy experiments" experience

When viewing the list of experiments, I think hovering over an experiment will display the ellipses on the right. After clicking on the ellipses, the only available option would be to "Copy experiment".  Clicking "Copy experiment would redirect to the Experiment Creation Wizard with all of the experiment details copied over.

## Action Items
- [ ] Discuss the "Copy experiments" experience
- [ ] Implement it or close issue

jessie-ross:
**This PR implements basic experiment cloning.**
This is quite simple to implement in this basic form.

Resolves #380 

## How has this been tested?

### DIFF
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -44,6 +44,9 @@ export default function Routes(): JSX.Element {
         <Route path='/experiments/:experimentIdSlug/wizard-edit' exact>
           <ExperimentWizard experimentWizardMode={ExperimentWizardMode.Edit} />
         </Route>
+        <Route path='/experiments/:experimentIdSlug/clone' exact>
+          <ExperimentWizard experimentWizardMode={ExperimentWizardMode.Clone} />
+        </Route>
         <Route path='/experiments/:experimentIdSlug/:view' exact>
           <Experiment />
         </Route>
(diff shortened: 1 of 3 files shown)


aaronyan (approved):
Yay, exciting stuff. LGTM! Should we have some e2e or unit tests for this (is is possible)? If not needed, then feel free to merge.

I removed the video link because of production data. :) Scrubbed the issue history too. It was helpful to watch though to see how it works on your end.

jessie-ross:
> I removed the video link because of production data. :) Scrubbed the issue history too. It was helpful to watch though to see how it works on your end.

Ah oops!! Thanks for that!!!

jessie-ross:
> Should we have some e2e or unit tests for this (is is possible)? If not needed, then feel free to merge.

I would say it is not needed as experiment creation/edit are well tested, and the experiment form is well tested internally. Really we are just switching the data source. But let's check with @yanirs to double check.

yanirs:
I think it'd be good to expand our e2e suite to cover more of these scenarios, but this case is simple enough to defer to when we do more work on general e2e testing.

PR #412, opened 2020-12-01
Fix experiment edit

jessie-ross:
**This PR fixes a bug caused by #411 where the experiment name goes missing on editing...**

## How has this been tested?

### DIFF
--- a/src/pages/experiments/ExperimentWizard.tsx
+++ b/src/pages/experiments/ExperimentWizard.tsx
@@ -12,7 +12,7 @@ import SegmentsApi from 'src/api/SegmentsApi'
 import TagsApi from 'src/api/TagsApi'
 import ExperimentForm from 'src/components/experiments/wizard/ExperimentForm'
 import Layout from 'src/components/page-parts/Layout'
-import { experimentToFormData } from 'src/lib/form-data'
+import { ExperimentFormData, experimentToFormData } from 'src/lib/form-data'
 import * as Normalizers from 'src/lib/normalizers'
 import { ExperimentFull, ExperimentFullNew, TagNamespace } from 'src/lib/schemas'
 import { useDataLoadingError, useDataSource } from 'src/utils/data-loading'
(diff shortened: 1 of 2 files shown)


src/pages/experiments/ExperimentWizard.tsx
```diff
@@ -147,7 +147,21 @@ export default function WizardEdit({
   }
   const onSubmit = onSubmitByExperimentWizardMode[experimentWizardMode]
 
-  const initialExperiment = experiment && experimentToFormData(clonedExperiment(experiment))
+  let initialExperiment: null | ExperimentFormData = null
```
yanirs, line 150: Looking at #41, I wonder if we make this simpler by doing something like this?

```typescript
const initialExperiment = experiment && experimentToFormData(experimentWizardMode === ExperimentWizardMode.Clone ? clonedExperiment(experiment) : experiment)
```

Also, do we really need `clonedExperiment()` if the goal is only to erase the name? There's no harm in leaving the existing name, as the backend will return an error if it isn't changed.

jessie-ross: > Also, do we really need clonedExperiment() if the goal is only to erase the name?

Probably don't need it at the moment.

> There's no harm in leaving the existing name, as the backend will return an error if it isn't changed.

We don't show backend errors yet and it can be quite confusing to the user when a save is rejected.

yanirs: I still don't see why we need the `switch`. I'll push a commit. :slightly_smiling_face:

src/pages/experiments/ExperimentWizard.tsx
```diff
@@ -147,7 +135,9 @@ export default function WizardEdit({
   }
   const onSubmit = onSubmitByExperimentWizardMode[experimentWizardMode]
 
-  const initialExperiment = experiment && experimentToFormData(clonedExperiment(experiment))
+  const initialExperiment =
+    experiment &&
+    experimentToFormData(experimentWizardMode === ExperimentWizardMode.Clone ? { ...experiment, name: '' } : experiment)
```
yanirs, line 140: @jessie-ross unless I missed something, this is a simpler solution than the `switch`, which gives us what we want.

yanirs (approved):
@jessie-ross unless I missed something with the last commit, I think this should be good to go.

src/pages/experiments/ExperimentWizard.tsx
jessie-ross: I rate them the same simplicity wise (using an "entanglement" definition of simplicity).

I think in the switch it is more obvious what is going on - for this version it is easy to miss that there are really three cases here.

yanirs: If two cases behave the same then it's really two cases. :slightly_smiling_face: 

I find the switch quite cumbersome and hard to follow, especially in the context of the PR and the code that was there before.

yanirs: Also, in the context of the entire function, it may be clearer if we only switched once based on the mode &ndash; with the mappings above and below, we're doing it three times. In general, this function is getting a bit too complicated at this point and is worth breaking up, but definitely not for this PR.

Issue #414, opened 2020-12-02
Clicking on some metrics causes an error


yanirs:
The error is: "Can't read property 'bottom' of undefined"

See p1606825441001400-slack-G01BQ5WMR8Q

jessie-ross:
Looks like this as we aren't currently accounting for `metricEstimates` being nullable.
Fix coming soon.

PR #417, opened 2020-12-02
Fix metric estimate errors

jessie-ross:
**This PR fixes https://github.com/Automattic/abacus/issues/414 where opening some metric results was causing the UI to crash.**

~~Seems like the root issue was that `metricEstimates` was incorrectly typed, this PR fixes that as well.~~

Nope...

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/ExperimentPageView.tsx
+++ b/src/components/experiments/single-view/ExperimentPageView.tsx
@@ -124,15 +124,15 @@ export default function ExperimentPageView({
       }
 
       return {
-      ...analysis,
-      metricEstimates: {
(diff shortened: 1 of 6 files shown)


jessie-ross:
Actually, after fixing this it might be that the schema was correct in the first place: https://app.swaggerhub.com/apis/yanir/experiments/0.1.0#/Analysis

jessie-ross:
Ignore the previous comments/commits.

**The real problem here was the diff inversion always created an object for `metricEstimates` even if it was null, this PR fixes that.**

yanirs (approved):

PR #424, opened 2020-12-04
Make accidental experiment control changes harder: Pt 2: The Dialogs

### LINKED ISSUE #420: Improve messaging around the Run and Disable buttons
From @jessie-ross on Slack:

> I have a feeling it may have been pressed to test the experiment out and instead of restricting to owner only, could just need:
> * A rename or clarification “Run (Deploy)”
> * Making the button red
> * Adding a danger sign in the confirm window
> * Add a warning message clearly explaining what the button does

I think we should make similar adjustments to the Disable button, as stopping an experiment that's already running is irreversible. Maybe change `Run` to `Deploy` and `Disable` to `Stop`, and also add icons to the buttons?

Another thing that may help is changing the order of the buttons to:

* Edit in Wizard
* Run
* Disable
* Clone

Since Clone is always enabled, it'd be nice to put it in the right-most position.

jessie-ross:
**This PR makes it harder to accidentally use the Experiment Controls by:**
- Adding better explanations to the dialog
- Swaps the buttons around
- Spreads the buttons apart
- Using a red button to confirm

Screen Shot 2020-12-08 at 1 32 09 am
Screen Shot 2020-12-08 at 1 31 14 am

Resolves #420 

## How has this been tested?


src/components/experiments/single-view/ExperimentDisableButton.tsx
```diff
@@ -72,20 +85,33 @@ const ExperimentDisableButton = ({
...
               variant='contained'
-              color='primary'
+              classes={{ contained: classes.buttonContained }}
               disabled={isSubmittingDisableExperiment}
               onClick={onConfirmDisableExperiment}
             >
               Disable
             </Button>
           </LoadingButtonContainer>
+          <Button variant='contained' color='primary' onClick={onCancelDisableExperiment}>
```
yanirs, line 112: Changing the order of the buttons may cause more confusion, as [most experimenters are not Windows users](https://ux.stackexchange.com/a/1077). [The Material Design guidelines](https://material.io/components/dialogs#specs) also show Cancel on the left.

Alternatively, we can stack the buttons and use more words, e.g., `Disable (irreversible)` and `Cancel (safe)`.

yanirs (commented):
The design and scary image look good, though I think the wording can be improved a bit. My main concern is around the order of the buttons, which may cause more confusion. It may be worth running this by a designer if you want another opinion.

jessie-ross:
## Update
- Put the buttons back in the prev order.
- Simplify the language.
- Don't show danger for disabling from staging.
- Adds red tinted backdrop to promote a sense of fear

Screen Shot 2020-12-10 at 3 06 38 am
Screen Shot 2020-12-10 at 3 06 56 am
Screen Shot 2020-12-10 at 3 07 34 am

src/components/experiments/single-view/ExperimentRunButton.tsx
```diff
@@ -76,16 +90,35 @@ const ExperimentRunButton = ({
           </Button>
         </span>
       </Tooltip>
-      <Dialog open={isAskingToConfirmRunExperiment} aria-labelledby='confirm-run-experiment-dialog-title'>
+      <Dialog
+        open={isAskingToConfirmRunExperiment}
+        aria-labelledby='confirm-run-experiment-dialog-title'
+        BackdropProps={{ className: classes.dangerBackdrop }}
+      >
+        <DialogTitle>
```
jessie-ross, line 98: I also improved the title spacing

src/components/experiments/single-view/ExperimentRunButton.tsx
```diff
@@ -23,6 +24,19 @@ const useStyles = makeStyles((theme: Theme) =>
...
+      color: theme.palette.error.contrastText,
+      '&:hover': {
+        background: theme.palette.error.light,
+      },
+    },
+    danger: {
+      textAlign: 'center',
+    },
+    dangerBackdrop: {
+      backgroundColor: 'rgb(195 61 61 / 62%)',
```
yanirs, line 38: There's quite a bit of duplication here. It'd be good to move these styles to a more central location so they can be reused.

yanirs (approved):
The changes look good! As noted inline, it'd be good to avoid duplicating the styles, especially given that it involves magic numbers.

PR #427, opened 2020-12-10
Let experimenters know dates are automatic

jessie-ross:
**This PR adds text underneath the start-date and end-date fields in the wizard to let Experimenters know that they are used for automation:**

Screen Shot 2020-12-11 at 2 15 25 am

Created this after seeing several experimenters surprised about it.

## How has this been tested?

### DIFF
--- a/src/components/experiments/wizard/BasicInfo.tsx
+++ b/src/components/experiments/wizard/BasicInfo.tsx
@@ -106,7 +106,13 @@ const BasicInfo = ({
           name='experiment.startDatetime'
           id='experiment.startDatetime'
           label='Start date'
-          helperText='Use the UTC timezone.'
+          helperText={
+            <>
+              Automatic start date.
+              <br />
+              Use the UTC timezone.
+            </>
+          }
           type='date'
           variant='outlined'
           required
@@ -125,7 +131,13 @@ const BasicInfo = ({
           name='experiment.endDatetime'
           id='experiment.endDatetime'
           label='End date'
-          helperText='Use the UTC timezone.'
+          helperText={
+            <>
+              Automatic end date.
+              <br />
+              Use the UTC timezone.
+            </>
+          }
           type='date'
           variant='outlined'
           required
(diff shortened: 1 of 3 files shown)


yanirs:
I think this can be made clearer. How about we change to something like this?

> :warning: The experiment will **start automatically** around midnight UTC on this date.
> :warning: The experiment will **end automatically** around midnight UTC on this date.

aaronyan:
> ⚠️ The experiment will start and end automatically around midnight UTC on these dates.

I like this version a lot! Is it possible to have this text horizontally span the width of the modal? Similar to the other fields? If that's too much work to make look nice, I vote for shortened but separate text with no caution sign:

> Starts automatically on this date around midnight UTC.
> Ends automatically on this date around midnight UTC.

jessie-ross:
**How is this looking:**

Screen Shot 2020-12-15 at 2 38 53 pm

- Decided to go with an info vibe instead of warning, but I think it is still pretty strong.

- I think we eventually need to move to actual labels for our controls - not the label inline with the outline, as they are a bit hard to see so aren't involved enough in creating a visual hierarchy.

yanirs:
The alert format looks good to me. We can also drop the `Use the UTC timezone.` help texts to make it less wordy. If we want to retain the reference to UTC around the fields, we can change the `Start date` and `End date` labels to `Start date (UTC)` and `End date (UTC)`.

I'm not too fussed if you want to keep it as it is, though.

jessie-ross:
Latest snap:
Screen Shot 2020-12-15 at 3 39 42 pm

jessie-ross:
Ah! Is it midnight at the start of end of these dates? I am guessing it is morning version? I think it would help to include that.

yanirs:
> Ah! Is it midnight at the start of end of these dates? I am guessing it is morning version? I think it would help to include that.

Good point. It's at the start. Maybe we can change `around midnight UTC` to `around 00:10 UTC` to make it less ambiguous?

jessie-ross:
Screen Shot 2020-12-15 at 4 35 51 pm

yanirs (approved):
LGTM, though @aaronyan might want to have another look.

aaronyan (approved):
Woo! Looks great to me. I really like the use of alerts throughout these PRs. 🚢

PR #430, opened 2020-12-15
Add inline info links

### LINKED ISSUE #426: Improve UX around "Change Expected" and "Minimum Difference" fields
From @demetd: 
> Would be good to have an info box next to "Change Expected" and "Minimum Difference" when adding new metrics, I don't know what the former is or what it impacts. Is it to allow different recommendations for A/A tests?
>
> p1607352709149500-slack-C7HH3V5AS

jessie-ross:
**This PR adds info boxes with links to the Wiki for Metrics and Exposure Events:**

Screen Shot 2020-12-15 at 3 28 53 pm

Fixes #426 

## How has this been tested?

### DIFF
--- a/src/components/experiments/wizard/Metrics.tsx
+++ b/src/components/experiments/wizard/Metrics.tsx
@@ -3,6 +3,7 @@ import {
   FormControl,
   IconButton,
   InputAdornment,
+  Link,
   MenuItem,
   Select as MuiSelect,
   Table,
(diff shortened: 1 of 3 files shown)


aaronyan (approved):
Really nice idea! Love that we're linking to our documentation. 

I would recommend making it so that we default to [open the links in new tabs](https://material-ui.com/components/links/#security) (`target="_blank"`) so that folks don't lose their progress in the experiment wizard.

After that, LGTM! Pre-approving for now.

PR #431, opened 2020-12-17
Update bookmarklet to use the new assignment endpoint

withinboredom:
Depends on D54296-code which uses the new dedicated assignment endpoint.

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/overview/VariationsTable.tsx
+++ b/src/components/experiments/single-view/overview/VariationsTable.tsx
@@ -44,33 +44,36 @@ function assignmentHref(variationName: string, experimentName: string) {
   nameSchema.validateSync(experimentName)
   return `javascript:(async () => {
     const response = await fetch(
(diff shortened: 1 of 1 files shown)


yanirs:
@aaronyan I made myself the reviewer since I'm already deep in this work :slightly_smiling_face:

yanirs (approved):
If you're happy with the changes I made in https://github.com/Automattic/abacus/pull/431/commits/4f2fd58eb0b77b7ee816acc1224185ae5f8aaa36?w=1, I'm happy to ship this and handle authorisation separately based on the discussion in pbmo2S-Dl-p2. It feels like an improvement since authorisation wasn't working before either. :slightly_smiling_face:

withinboredom:
> I'm happy to ship this and handle authorisation separately based on the discussion in pbmo2S-Dl-p2.

I missed this comment. But it seems like a simple enough solution for now. We can handle calypso separately.

Issue #415, opened 2020-12-02
Show backend errors to the user


jessie-ross:
> There's no harm in leaving the existing name, as the backend will return an error if it isn't changed.

We don't show backend errors yet and it can be quite confusing to the user when a save is rejected.

_Originally posted by @jessie-ross in https://github.com/Automattic/abacus/pull/412#discussion_r533915523_

This is also important in a number of other instances.

yanirs:
Making this a high priority since it's one of these annoying rough edges that we should probably address soon.

PR #438, opened 2021-01-08
Add generic HttpResponseErrors

### LINKED ISSUE #300: Replace our custom fetch wrapper with a library

jessie-ross:
This PR adds generic `HttpResponseError`:
- Cleans up `fetchApi` by using a fetch library.
- Consolidates http errors into a single HttpResponseError.
- `fetchApi` is now tested and the underlying code is battle-tested

Resolves #300 
Makes easy: #415 

## How has this been tested?

### DIFF
--- a/src/__tests__/api/AutocompleteApi.test.ts
+++ b/src/__tests__/api/AutocompleteApi.test.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/require-await */
 import { getEventNameCompletions, getPropNameCompletions, getUserCompletions } from 'src/api/AutocompleteApi'
-import NotFoundError from 'src/api/NotFoundError'
+import HttpResponseError from 'src/api/HttpResponseError'
 import * as Utils from 'src/api/utils'
 
 jest.mock('src/api/utils')
@@ -74,7 +74,7 @@ test('an empty event name returns a useful error message', async () => {
 
 test('a nonexistent event name returns a useful error message', async () => {
   mockedUtils.fetchApi.mockImplementation(async () => {
-    throw new NotFoundError()
+    throw new HttpResponseError(404)
   })
   expect(await getPropNameCompletions('no_exist')).toMatchInlineSnapshot(`null`)
 })
--- a/src/__tests__/api/AutocompleteApi.test.ts
+++ b/src/__tests__/api/AutocompleteApi.test.ts
@@ -1,4 +1,6 @@
 /* eslint-disable @typescript-eslint/require-await */
+import { StatusCodes } from 'http-status-codes'
+
 import { getEventNameCompletions, getPropNameCompletions, getUserCompletions } from 'src/api/AutocompleteApi'
 import HttpResponseError from 'src/api/HttpResponseError'
 import * as Utils from 'src/api/utils'
@@ -74,7 +76,7 @@ test('an empty event name returns a useful error message', async () => {
 
 test('a nonexistent event name returns a useful error message', async () => {
   mockedUtils.fetchApi.mockImplementation(async () => {
-    throw new HttpResponseError(404)
+    throw new HttpResponseError(StatusCodes.NOT_FOUND)
   })
   expect(await getPropNameCompletions('no_exist')).toMatchInlineSnapshot(`null`)
 })
--- a/src/__tests__/api/MetricsApi.test.ts
+++ b/src/__tests__/api/MetricsApi.test.ts
@@ -80,7 +80,7 @@ describe('MetricsApi.ts module', () => {
         expect(false).toBe(true) // This should never be reached.
       } catch (err) {
         expect(err).toBeInstanceOf(HttpResponseError)
-        expect(err).toBe(StatusCodes.NOT_FOUND)
+        expect((err as HttpResponseError).status).toBe(StatusCodes.NOT_FOUND)
       }
     })
   })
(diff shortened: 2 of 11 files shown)


src/__tests__/api/AutocompleteApi.test.ts
```diff
@@ -74,7 +74,7 @@ test('an empty event name returns a useful error message', async () => {
 
 test('a nonexistent event name returns a useful error message', async () => {
   mockedUtils.fetchApi.mockImplementation(async () => {
-    throw new NotFoundError()
+    throw new HttpResponseError(404)
```
yanirs, line 77: Can we use something like the constants from https://github.com/prettymuchbryce/http-status-codes instead of numbers?

src/__tests__/api/TagsApi.test.ts
```diff
@@ -58,7 +58,10 @@ describe('TagsApi.ts module', () => {
         await TagsApi.findById(0)
         expect(false).toBe(true) // This should never be reached.
       } catch (err) {
-        expect(err).toBeInstanceOf(NotFoundError)
+        expect(err).toBeInstanceOf(HttpResponseError)
+        if (err instanceof HttpResponseError) {
```
yanirs, line 62: Why is the `if` needed? We just checked the type on the previous line.

yanirs (commented):
Can you please explain more why `wretch` is needed? It looks like we're doing quite a lot to work around its limitations.

What's stopping us from adding our own `HttpResponseError` without the `wretch` dependency and extra conversions?

src/api/utils.ts
```diff
@@ -1,62 +1,81 @@
...
   })
 
-  // istanbul ignore next; branch can't be reached with the current tests.
-  if (response.status === 404) {
-    throw new NotFoundError()
-  }
-
-  // istanbul ignore next; branch can't be reached with the current tests.
-  if (response.status >= 300 || response.status < 200) {
-    throw new Error('Received a non-2XX response from server.')
```
yanirs, line 47: Can't we generalise lines 42-47 to throw our own `HttpResponseError`?

src/__tests__/api/TagsApi.test.ts
jessie-ross: This is necessary for typescript to be happy.

src/__tests__/api/AutocompleteApi.test.ts
jessie-ross: Sure

src/__tests__/api/TagsApi.test.ts
yanirs: Are you sure? Why is it not needed in `MetricsApi.test.ts`?

yanirs: Ah, in `MetricsApi.test.ts` we use `expect(err).toBe(404)` and the tests are skipped anyway. :man_facepalming: 

Maybe use `expect(err.status as HttpResponseError).toBe(404)` instead?

jessie-ross:
> Can you please explain more why wretch is needed? It looks like we're doing quite a lot to work around its limitations.

Ok, yep `wretch` is a small part of this PR and can easily be removed but I think it is a worthwhile addition especially since we are already working over and retesting this code.

Why `wretch`:
- Flexible interface without having to continue rolling our own for new features.
   An API function can use the `exPlatWretcher` directly if they need to e.g. change headers or have special processing.
   I think this is how it should be used.
- Prevents future funky introductions into `fetchAPI`.
   There has been funky stuff before that I have since removed.
- Provides some guide rails around web standards and conventions.
   E.g. a lot of the "working around limitations" is because we are returning text responses (empty strings) where we should be returning JSON or No Content responses.
- Battle tested, > 3 years old, high use, no dependencies.

The needing to wrap error messages is the only limitation I can see and but most of that code we would need to have for either solution.

yanirs:
Thanks for the explanation @jessie-ross! I'm happy to go with `wretch`, so it's just a matter of addressing the other comments and this PR will be good to go.

> E.g. a lot of the "working around limitations" is because we are returning text responses (empty strings) where we should be returning JSON or No Content responses.

I'm not sure if we have an issue for this, but we should probably fix it on the backend at some point.

src/__tests__/api/MetricsApi.test.ts
```diff
@@ -77,7 +79,8 @@ describe('MetricsApi.ts module', () => {
         await MetricsApi.findById(0)
         expect(false).toBe(true) // This should never be reached.
       } catch (err) {
-        expect(err).toBeInstanceOf(NotFoundError)
+        expect(err).toBeInstanceOf(HttpResponseError)
+        expect(err).toBe(StatusCodes.NOT_FOUND)
```
yanirs, line 83: This will fail once the test isn't skipped anymore, as lines 82-83 can't be true. :slightly_smiling_face:

src/__tests__/api/utils.test.ts
```diff
@@ -0,0 +1,60 @@
...
+          expect(error.name).toBe('Error')
+          expect(error.message).toBe('Error Message')
+        }
+      }
+    })
+
+    it('should return a correct HttpResponseError for a bad HTTP request', async () => {
+      expect.assertions(3)
+      fetchMock.once('*', {
+        status: 400,
```
yanirs, line 39: It'd be good to change the 400s here too.

yanirs (approved):
LGTM, but see a couple of minor comments inline.

PR #439, opened 2021-01-08
Add backend errors to Wizard

jessie-ross:
**This PR adds shows backend error messages to the Experimenter in the Experiment Wizard.**
- Adds ServerErrorAlert component

Screen Shot 2021-01-08 at 3 49 14 pm

Resolves: #415 

## How has this been tested?

To get a Server Error:
1. Spin up a production data Abacus
2. Clone an explat_test experiment
3. Copy the experiment_name from the original
4. (Update the start/end dates)
5. Try submit, see same error in screenshot above

To get a successful submit:
1. Create a normal experiment


src/components/general/ServerErrorAlert.tsx
```diff
@@ -0,0 +1,29 @@
...
+      margin: theme.spacing(2, 0),
+    },
+  }),
+)
+
+export default function ServerErrorAlert({ error }: { error?: Error }): JSX.Element {
+  const classes = useStyles()
+  return (
+    <>
+      {error && error instanceof HttpResponseError && (
```
yanirs, line 19: It's a bit odd that this doesn't render anything if it's not a `HttpResponseError`. I think it'd be better to change the type of the `error` parameter and move the type check to the calling code.

src/components/general/ServerErrorAlert.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      margin: theme.spacing(2, 0),
+    },
+  }),
+)
+
+export default function ServerErrorAlert({ error }: { error?: Error }): JSX.Element {
```
yanirs, line 15: This should be documented, especially as the behaviour is unintuitive. Also, the name `ServerErrorAlert` is a bit confusing &ndash; I think of server errors as 5xx status codes, not as "any error coming from the server".

Maybe change the signature and name to `HttpResponseErrorAlert({ error }: { error: HttpResponseError })`?

src/components/general/ServerErrorAlert.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import ServerErrorAlert from './ServerErrorAlert'
+
+test('renders nothing for no error', () => {
+  const { container } = render(<ServerErrorAlert />)
+  expect(container).toMatchSnapshot()
+})
+
+test('renders an HttpResponseError with server message in JSON', () => {
+  const error = new HttpResponseError(400)
```
yanirs, line 14: As in #438, it'd be good to use constants.

src/pages/experiments/ExperimentWizard.tsx
```diff
@@ -100,7 +101,8 @@ export default function WizardEdit({
         enqueueSnackbar('Experiment Created!', { variant: 'success' })
         history.push(`/experiments/${receivedExperiment.experimentId}/code-setup`)
       } catch (error) {
-        enqueueSnackbar('Failed to create experiment 😨 (Form data logged to console.)', { variant: 'error' })
+        setFormSubmissionError(error)
+        enqueueSnackbar('Failed to create experiment 😨', { variant: 'error' })
```
yanirs, line 105: This still leaves users in the dark on certain errors, e.g., on Vercel I went through the creation process and use uBlock to block `virtserver.swaggerhub.com`. All I get is the snackbar saying "Failed to create experiment 😨". Looking at the console, I see `TypeError: NetworkError when attempting to fetch resource` &ndash; this would be useful to surface.

yanirs (commented):
The general design looks good and is more informative, but I think we can improve clarity around the new component and handle more errors &ndash; see inline comments.

src/components/general/ServerErrorAlert.tsx
jessie-ross: I think it is better to have the type checking within as it simplifies the code outside and allows declarative usage.
I was also considering to later generalise this element to other errors.

src/components/general/ServerErrorAlert.tsx
jessie-ross: Might possibly change this to `GeneralErrorAlert` such that it handles all errors.

src/components/general/ServerErrorAlert.tsx
yanirs: Yeah, if this component is changed to handle any error then I think it'll be fine.

src/components/general/GeneralErrorAlert.tsx
```diff
@@ -0,0 +1,37 @@
...
+)
+
+export default function GeneralErrorAlert({ error }: { error?: Error }): JSX.Element | null {
+  const classes = useStyles()
+
+  if (error instanceof HttpResponseError) {
+    return (
+      <Alert severity='error' className={classes.root}>
+        <AlertTitle>
+          Server Error: {error.status} {error.response.statusText}
```
yanirs, line 22: Not critical, but it might be better to change `Server Error` to something like `Server returned error`, as non-5xx codes are technically not server errors.

yanirs (approved):
Looks good. Up to you if you want to change the wording a bit.

src/components/general/GeneralErrorAlert.tsx
jessie-ross: I've changed it to "Error Response".

PR #440, opened 2021-01-11
Add backend errors to Snackbars

jessie-ross:
**_STACKED PR_**

**This PR adds backend errors to Snackbars, giving full coverage of displaying backend errors.**

Screen Shot 2021-01-11 at 3 56 16 pm

Alongside #439 for resolving #415  

## How has this been tested?

To manually test: Alter the code so it allows an illegal server operation and send a request off.


src/api/HttpResponseError.ts
```diff
@@ -42,3 +42,13 @@ export function wretcherErrorToHttpResponseError(wretcherError: WretcherError):
...
 }
+
+export function serverErrorMessage(error: Error | undefined | null): string {
+  if (!(error instanceof HttpResponseError)) {
+    return ''
+  }
+
+  const message: string =
+    error.json && typeof error.json === 'object' ? (error.json as Record<string, string>).message : 'No server message'
+  return `Server Error: ${error.status}: ${message}`
```
withinboredom, line 53: It might look at little better to say `Server Error [${error.status}]: ${message}` instead of the double colons.

withinboredom (approved):

src/api/HttpResponseError.ts
jessie-ross: Ah nice! Yeah I was looking at the double colon too but didn't think of another way :)

PR #442, opened 2021-01-12
Fix ROPE scaling in plots

### LINKED ISSUE #435: Accidentally scaling rope up for conversion metrics
Looks like we accidentally are scaling ROPE up in graphs when we are scaling the Diff stats for conversion metrics. Simple fix.

jessie-ross:
**This PR fixes rope scaling in plots.**
- Also renames `%` to `Percentage Points` for the diff graph.

Screen Shot 2021-01-12 at 12 52 59 pm

Resolves #435 

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
+++ b/src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
@@ -352,7 +352,7 @@ function AnalysisDetailPanel({
     {
       name: 'ROPE: lower bound',
       x: dates,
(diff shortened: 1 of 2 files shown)


yanirs:
Is this a draft because we're not sure that we want to unscale on read? In any case, this won't solve the general problem since the ETL still assumes that the min difference isn't scaled, so the recommendations are often wrong.

yanirs:
P2ed: pbmo2S-Ez-p2

Issue #419, opened 2020-12-03
Improve result page wording and displayed data


yanirs:
The details may be best hashed out in the context of a PR, but here are some ideas:
* [x] The difference interval can be a sentence, e.g., _"There is a 95% probability that the value is between `<bottom>` and `<top>`"_.
* [x] For conversion metrics, the difference interval should be multiplied by 100 to get percent values (consistency with the plots).
* [x] Finish adding a top-level recommendation (#369)
* [x] Consider dropping the strategy.
* [x] Add the variation value ranges (in a sentence?) so that users won't need to hover over the graphs to get them.
* [x] Add details to _"manual analysis required"_, e.g., _"manual analysis required: contact @experimentation-platform-project on #a8c-experiments"_.
* [x] Move the recommendations to the top-level area.
* [x] Make the remaining warnings friendlier.
* [ ] Integrate feedback from Greg: pbmo2S-Fn-p2.

jessie-ross:
By "Replace Python duration warnings with client-side duration warnings and move them to the top-level area.", do you mean doing the calculation in the front-end?

yanirs:
Yeah, it's a trivial calculation (`end_datetime - start_datetime`). I specifically meant ignoring `RecommendationWarning.ShortPeriod` and `RecommendationWarning.LongPeriod` in favour of smarter date warnings. For example, duration warnings are irrelevant for email experiments. The original thinking behind having them in the Python code was to have a single source of truth (in Hypotheses the same duration warnings were calculated in two places), but it's too inflexible and a pain to maintain.

aaronyan:
> The difference interval can be a sentence, e.g., "There is a 95% probability that the value is between <bottom> and <top>".

This wording was helpful for some experimenters, and they also requested it to be added to Abacus result page (pbxNRc-De#comment-1404). Just adding this comment as a vote for this improvement!

yanirs:
Also edited to add a task item to integrate Greg's feedback from pbmo2S-Fn-p2

jessie-ross:
Oooh! After seeking @veekaybee's extended wording it might be good to do this for the difference interpretation:

<img width="875" alt="Screen Shot 2021-02-05 at 9 41 24 am" src="https://user-images.githubusercontent.com/971886/106978081-7922a100-6796-11eb-8552-854d610ecb36.png">

yanirs:
[...]

PR #443, opened 2021-01-13
Improve Metric Assignment Results 

### LINKED ISSUE #355: Trend plots are unclear when there are results from a single day

jessie-ross:
**This PR improves metric assignment results:**

(Edit: Scrubbed Photos)

With only one data point:

(Edit: Scrubbed Photos)

Changes:
- Adds sections to Metric Assignment Results
- Moves the graphs to the top
- Add variation estimates
- Add interpretation of estimates
- Changes diff graph for conversions to have "percentage points" as unit.
- Refactors metric formatting into `formatMetricValue`
- Additionally changes the Participant graph to use: "lines+markers" to partially fix issue #355 
   (Probably shouldn't be in this PR but leaving it in as it is so small.)

Partially resolves issue #419 

## How has this been tested?


jessie-ross:
This does reveal the x100 percentage points, so it does need to be done in time with other percentage point changes. It can be taken out pretty easily though.

src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
```diff
@@ -213,15 +213,10 @@ function MetricAssignmentsPanel({
...
-                    <>
-                      {resolvedMetricAssignment.minDifference}&nbsp;
-                      <Tooltip title='Percentage Points'>
-                        <span className={classes.tooltipped}>pp</span>
-                      </Tooltip>
-                    </>
+                  {formatMetricValue(
+                    resolvedMetricAssignment.minDifference,
+                    resolvedMetricAssignment.metric.parameterType,
+                    true,
```
yanirs, line 219: It's hard to tell what's `true` referring to in this context. It'd be easier to read if `formatMetricValue()` was changed to work with an object as a single parameter.

src/utils/formatters.ts
```diff
@@ -1,17 +1,59 @@
...
+ * Precision to be inputed into _.round, to be used outside of graphs.
+ */
+export const metricValueFormatPrecision = 4
+
+/**
+ * Metric Formatting Data
  */
-function formatUsCurrencyDollar(value: number): string {
-  return usCurrencyDollarFormatter.format(value)
+export const metricValueFormatData: Record<
```
yanirs, line 20: Why are these two consts exported?

yanirs:
This feels like a step in the right direction, but I'm not sure about some of these changes:

* Was the change to not always show dollar values with cents intentional (`$x.cc`)? If so, what's the reasoning?
* Always spelling out `pp` feels too verbose. I think that the tooltip approach is better in places where we use % or $. Experimenters may need to hover once or twice, but then they won't need to do it anymore.
* It'd be good to reorder things under the `Latest Estimates` heading, putting the difference first, as it's the most important estimate. It'll also work nicely when we add support for more than two variations. 
* Shall we hide the metric-level graphs in the single-point case? Or are you planning on doing it separately?
* Having the warnings under `Latest Estimates` is a bit unexpected. Maybe they should be under `Analysis Fine Print`?
* I find that my eyes jump a lot with the way the three tables are (mis)aligned. Maybe we can use a single table with subheadings for better alignment of the second column? Something like:
<table>
<tr><td><h3>Latest Estimates</h3></td><td></td></tr>
<tr><td>Difference</td><td>[...]</td></tr>
<tr><td>control</td><td>[...]</td></tr>
<tr><td><h3>Metric Assignment Details</h3></td><td></td></tr>
<tr><td>Metric Description</td><td>This is a metric...</td></tr>
</table>

By the way, I edited the PR description to add the word "issue" before the issue number so that they don't get closed by this PR (`partially fixes issue #xxx` doesn't close the issue, while `partially fixes #xxx` does). 

> This does reveal the x100 percentage points, so it does need to be done in time with other percentage point changes. It can be taken out pretty easily though.

I don't think it's an issue since it's already a problem in the graph and the recommendations. We should aim to fix it this week, though.

src/utils/formatters.ts
jessie-ross: Ah yeah I was going to use the transform for the graphs but I can unexport it.

jessie-ross:
> Was the change to not always show dollar values with cents intentional ($x.cc)? If so, what's the reasoning?

Yeah, since this is ARPU the level of precision you might need could be below 2 decimal places and then I figured we should just use the same level of rounding across Conversion and Revenue (4 decimal places).

> Always spelling out pp feels too verbose. I think that the tooltip approach is better in places where we use % or $. Experimenters may need to hover once or twice, but then they won't need to do it anymore.

Hmm 🤔 

> It'd be good to reorder things under the Latest Estimates heading, putting the difference first, as it's the most important estimate. It'll also work nicely when we add support for more than two variations.

:+1: 

> Shall we hide the metric-level graphs in the single-point case? Or are you planning on doing it separately?

Yeah separately, I accidentally let it creep in here for the participants.

> Having the warnings under Latest Estimates is a bit unexpected. Maybe they should be under Analysis Fine Print?

I think it is important to have warnings by the estimates, I don't think Experimenters are likely to look into the Fine Print.

> I find that my eyes jump a lot with the way the three tables are (mis)aligned. Maybe we can use a single table with subheadings for better alignment of the second column? Something like:

I can just fix the alignment

yanirs:
> Yeah, since this is ARPU the level of precision you might need could be below 2 decimal places and then I figured we should just use the same level of rounding across Conversion and Revenue (4 decimal places).

Hmm... Maybe we should just change the default to two decimal places? For percentages, four decimal places is actually six (1% = 0.01), and for dollar values, I believe people don't usually care about fractions of cents.

> I think it is important to have warnings by the estimates, I don't think Experimenters are likely to look into the Fine Print.

That's a good point. Maybe add a visual indicator to differentiate them? Something simple like ":warning: Warnings" may work well.

> I can just fix the alignment

That would work too :slightly_smiling_face:

jessie-ross:
## Update

- Changed rounding to 2 digits past the point
- Changed to "pp" with tooltip everywhere - replaced `formatMetricValue` with component `MetricValue`
- Put difference up top
- I've ended up hiding the plots for 1 data point:
Screen Shot 2021-01-14 at 4 10 00 pm

- Put the warning emoji in front of the warning line
- Align the table columns

src/components/general/MetricValue.tsx
```diff
@@ -0,0 +1,84 @@
...
+  isDifference = false,
+}: {
+  value: number
+  metricParameterType: MetricParameterType
+  isDifference?: boolean
+}): JSX.Element {
+  const format = metricValueFormatData[`${metricParameterType}${isDifference ? '_difference' : ''}`]
+  return (
+    <>
+      {format.prefix}
```
yanirs, line 79: Looking at https://abacus-git-add-improve-metric-assignments-results.a8c.vercel.app/experiments/2-experiment-name-123/results, negative dollar values look odd (`$-0.25` rather than `-$0.25`). If you don't want to use `Intl.NumberFormat` to work around it, maybe a good solution for now would be to make the prefix `USD ` (with a trailing space, perhaps nbsp)? This would also make it less ambiguous, and something like `USD -0.25` would probably be clear enough in a sentence.

src/components/general/MetricValue.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    render(<MetricValue value={0.01} metricParameterType={MetricParameterType.Revenue} isDifference={true} />)
+      .container,
+  ).toMatchInlineSnapshot(`
+    <div>
+      $
+      0.01
+      
+    </div>
+  `)
+})
```
yanirs, line 91: It'd be good to add two more test cases:
* Negative dollar value (see comment below)
* Many decimal points, e.g., 0.1232455 (validate the rounding)

src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
```diff
@@ -373,31 +386,208 @@ function AnalysisDetailPanel({
...
+                ? `Conversion rate difference estimates (percentage points)`
+                : `Revenue difference estimates ($)`,
+            }}
+            data={plotlyDataDifferenceGraph}
+            className={classes.metricEstimatePlot}
+          />
+        </div>
+      ) : (
+        <Typography variant='body1' className={classes.noPlotMessage}>
+          (Not enough data to plot.)
```
yanirs, line 421: I think we can just render nothing, or maybe change the wording to: `Past values will be plotted once we have more than one day of results.` (no need for parentheses).

src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
```diff
@@ -373,31 +386,208 @@ function AnalysisDetailPanel({
...
+  }
+
   return (
     <TableContainer className={clsx(classes.root, 'analysis-detail-panel')}>
+      {dates.length > 1 ? (
+        <div className={classes.metricEstimatePlots}>
+          <Plot
+            layout={{
+              ...Visualizations.plotlyLayoutDefault,
+              title: isConversion ? `Conversion rate estimates by variation (%)` : `Revenue estimates by variation ($)`,
```
yanirs, line 403: If we change the metric value prefix to USD (see below), we should also change it here and on line 413.

src/components/experiments/single-view/results/CondensedLatestAnalyses.tsx
```diff
@@ -373,31 +386,208 @@ function AnalysisDetailPanel({
...
+                  There is a 95% probability that the metric value for this variation is between{' '}
+                  <MetricValue
+                    value={latestEstimates[`variation_${variation.variationId}`].bottom}
+                    metricParameterType={metric.parameterType}
+                  />{' '}
+                  and{' '}
+                  <MetricValue
+                    value={latestEstimates[`variation_${variation.variationId}`].top}
+                    metricParameterType={metric.parameterType}
+                  />
```
yanirs, line 512: This should also have a full stop (like the difference sentence) :slightly_smiling_face:

yanirs (commented):
Looks great! See a few more inline comments.

yanirs (approved):
:ship:

PR #447, opened 2021-01-15
Implement min difference scaling

### LINKED ISSUE #377: Multiply conversion metric units by 100
**Q: Should we adjust the conversion metric units to "points" (from "pp")?**

Right now, we use `pp` to define "percentage points" for conversion metrics. We primarily use this unit when talking about the minimum difference value.

**Q: Should we multiple the conversion metric units by 100?**

The conversion metric units are also unchanged and a `0.02` value will appear as `0.02 pp`. I think we should be displaying this value as `2 pp`, which is basically x100 the original value. When reading metric assignments, we would x100. And when writing metric assignments, we would /100 before sending the request. 

Background to this issue: creating the experiments in Abacus for the monthly pricing experiment, and it got my thinking about this conversion metric unit experience.

## Action items

- [x] Discuss if we want to adjust conversion metric units from `pp` to `points`
- [x] Discuss if we want to multiply the values by 100 to show `2 points` instead of `0.02 points` (The actual value is 2/100).
- [x] Implement any of the changes discussed above

### LINKED ISSUE #435: Accidentally scaling rope up for conversion metrics

jessie-ross:
**This PR scales min difference x100 for editing.**

Resolves #377 (along with parent PR #443)
Resolves #435 

## How has this been tested?

   - Create new metric
   - Edit existing metric
   - Add metric assignment to running experiment, e.g. `/experiments/20073/overview`

### DIFF
--- a/src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
+++ b/src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
@@ -8,7 +8,6 @@ import {
   FormControl,
   FormHelperText,
   FormLabel,
-  InputAdornment,
   MenuItem,
   Toolbar,
   Tooltip,
@@ -23,13 +22,14 @@ import TableRow from '@material-ui/core/TableRow'
 import Typography from '@material-ui/core/Typography'
 import { Add } from '@material-ui/icons'
 import { ErrorMessage, Field, Formik } from 'formik'
-import { Select, Switch, TextField } from 'formik-material-ui'
+import { Select, Switch } from 'formik-material-ui'
 import { useSnackbar } from 'notistack'
 import React, { useMemo, useState } from 'react'
 import * as yup from 'yup'
 
 import ExperimentsApi from 'src/api/ExperimentsApi'
 import { serverErrorMessage } from 'src/api/HttpResponseError'
+import MetricDifferenceField from 'src/components/general/MetricDifferenceField'
 import MetricValue from 'src/components/general/MetricValue'
 import { AttributionWindowSecondsToHuman } from 'src/lib/metric-assignments'
 import * as MetricAssignments from 'src/lib/metric-assignments'
@@ -44,7 +44,6 @@ import {
   Status,
 } from 'src/lib/schemas'
 import { formatBoolean } from 'src/utils/formatters'
-import { formikFieldTransformer } from 'src/utils/formik'
 
 import LoadingButtonContainer from '../../../general/LoadingButtonContainer'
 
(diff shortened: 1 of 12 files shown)


src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
```diff
@@ -333,7 +340,13 @@ function MetricAssignmentsPanel({
                       Minimum Difference
                     </FormLabel>
                     <Field
-                      component={TextField}
+                      component={
```
yanirs, line 343: Looking at the other lines in this `Field`, it'd be nice to also add `max: '100'` to the `inputProps` when it's a conversion metric. Also, thinking of the changes in #443, the `startAdornment` should be `USD` for revenue metrics. To make the difference more explicit, we may also want to use different placeholders.

Therefore, it's probably worth doing the parameter type check on a higher level (`...parameterType === MetricParameterType.Conversion ? <Field with conversion params> : <Field with revenue params>`) instead of 2-4 times.

It may also be time to pull this out to a common component (`MetricMinDifferenceField`?), as we're starting to duplicate non-trivial code.

What do you think?

src/components/experiments/wizard/ExperimentForm.tsx
```diff
@@ -269,6 +269,7 @@ const ExperimentForm = ({
                 )}
                 {currentStageId === StageId.Submit && (
                   <div className={classes.formPart}>
+                    {console.log(formikProps.errors?.experiment?.metricAssignments)}
```
yanirs, line 272: Should this be here?

src/components/general/MetricDifferenceField.tsx
```diff
@@ -0,0 +1,84 @@
...
+        component={ConversionMetricTextField}
+        name={props.name}
+        id={props.id}
+        type='number'
+        variant='outlined'
+        placeholder='1.30'
+        inputProps={{
+          'aria-label': 'Minimum Difference',
+          min: '0',
+          max: 100,
```
yanirs, line 47: Should this be `max: '100'` to match the `min`?

src/components/general/MetricDifferenceField.tsx
```diff
@@ -0,0 +1,84 @@
...
+  if (props.metricParameterType === MetricParameterType.Conversion) {
+    return (
+      <Field
+        className={props.className}
+        component={ConversionMetricTextField}
+        name={props.name}
+        id={props.id}
+        type='number'
+        variant='outlined'
+        placeholder='1.30'
```
yanirs, line 43: It doesn't really matter, but we can also make the placeholder different.

yanirs (approved):
Looks good. Just a few small questions.

src/components/experiments/wizard/ExperimentForm.tsx
jessie-ross: Nope 👍

jessie-ross:
Merging this into `add/improve-metric-assignments-results`

PR #455, opened 2021-01-21
Remove not_final references

yanirs:
As part of addressing pbmo2S-F2-p2, stop displaying `not_final` counts so that we can stop calculating them on the backend.

## How has this been tested?


aaronyan (approved):
LGTM! This was much easier to remove than I had originally expected 😄  Nice work!

PR #449, opened 2021-01-19
Clean up data flow in Results

jessie-ross:
**This PR cleans up the data flow in Results.**
- Also allows debug mode users to switch between strategies. 
Screen Shot 2021-01-19 at 11 09 46 am

- Remove the logged data from the debug page 

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)

PR #456, opened 2021-01-21
Revert "Clean up data flow in Results"

aaronyan:
Reverts Automattic/abacus#449. I think another PR (#454) was stacked on this. Sorry about that!


yanirs:
As noted on Slack, there's no need to revert. @jessie-ross can move forward with #454 and resolve conflicts.

PR #418, opened 2020-12-03
Hack: Disable status change and editing buttons for nav_unification_v2

yanirs:
Following the discussion in #413, this is a temporary hack to disable the buttons for the `nav_unification_v2` experiment.

## How has this been tested?

* Run `REACT_APP_PRODUCTION_CONFIG_IN_DEVELOPMENT=true npm run dev`
* Navigate to the `nav_unification_v2` overview page and verify that the buttons are disabled:

image

* Navigate to a different `staging` experiment page and verify that the buttons are enabled

PR #458, opened 2021-01-27
Removes nav_unification hacks.

cpapazoglou:
This PR aims to reverse https://github.com/Automattic/abacus/pull/418. After the launch of nav-unification to all a12s this hacks are no longer needed (p7DVsv-ago-p2). I have also disabled the experiment which can now be safely deleted. There are no useful data about it.

## How has this been tested?

* Run `REACT_APP_PRODUCTION_CONFIG_IN_DEVELOPMENT=true npm run dev`
* Navigate to a `staging` experiment and make sure that buttons are still active
*


aaronyan (approved):
LGTM! Thank you for helping us remember and creating the PR to remove this from the code. 👏

PR #465, opened 2021-02-25
Add an assignment cache indicator to the experiment overview

yanirs:
Add a very basic indicator of an experiment's caching status to the overview page. It can probably benefit from more exciting styling, but I think it's good enough to get us started and aid with debugging once we change all the assigners to use the file system cache.

[This is what it looks like with dummy data](https://github.com/Automattic/abacus/pull/465#issuecomment-786466340).

## How has this been tested?

- Manual testing with mock data (locally and on Vercel)

### DIFF
--- a/src/components/experiments/single-view/overview/GeneralPanel.tsx
+++ b/src/components/experiments/single-view/overview/GeneralPanel.tsx
@@ -105,7 +105,18 @@ function GeneralPanel({
     {
       label: 'Assignment Cache Entry',
       value: (
-        <span className={classes.monospace}>{AssignmentCacheStatusToHuman[experiment.assignmentCacheStatus]}</span>
+        <span className={classes.monospace}>
+          {AssignmentCacheStatusToHuman[experiment.assignmentCacheStatus]}&nbsp;(
+          <Link
+            href="https://github.com/Automattic/experimentation-platform/wiki/Experimenter's-Guide#the-file-system-cache"
+            rel='noopener noreferrer'
+            target='_blank'
+            underline='always'
+          >
+            learn more
+          </Link>
+          )
+        </span>
       ),
     },
   ]
--- a/src/components/experiments/single-view/overview/GeneralPanel.tsx
+++ b/src/components/experiments/single-view/overview/GeneralPanel.tsx
@@ -29,6 +29,7 @@ import LabelValueTable from 'src/components/general/LabelValueTable'
 import { ExperimentFull, experimentFullSchema, Status, yupPick } from 'src/lib/schemas'
 import { formatIsoDate } from 'src/utils/time'
 
+import { AssignmentCacheStatusToHuman } from '../../../../lib/experiments'
 import LoadingButtonContainer from '../../../general/LoadingButtonContainer'
 
 const useStyles = makeStyles((theme: Theme) =>
@@ -99,6 +100,12 @@ function GeneralPanel({
         <Link href={experiment.p2Url} rel='noopener noreferrer' target='_blank' className={classes.monospace}>
           {experiment.p2Url}
         </Link>
+      ),
+    },
+    {
+      label: 'Assignment Cache Entry',
+      value: (
+        <span className={classes.monospace}>{AssignmentCacheStatusToHuman[experiment.assignmentCacheStatus]}</span>
       ),
     },
   ]
--- a/src/components/experiments/single-view/overview/GeneralPanel.test.tsx
+++ b/src/components/experiments/single-view/overview/GeneralPanel.test.tsx
@@ -189,6 +189,26 @@ test('renders as expected', () => {
                 </a>
               </td>
             </tr>
+            <tr
+              class="MuiTableRow-root"
+            >
+              <th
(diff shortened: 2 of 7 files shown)


src/components/experiments/single-view/overview/GeneralPanel.tsx
```diff
@@ -101,6 +102,12 @@ function GeneralPanel({
         </Link>
       ),
     },
+    {
+      label: 'Assignment Cache Entry',
```
yanirs, line 106: I didn't label it `Assignment Cache Status` to reduce confusion with the experiment status. I think it reads well enough as a sentence like this, e.g., `Assignment Cache Entry [...is...] Fresh: The cache matches the database for this experiment`.

jessie-ross:
Should this be a debug mode only feature, with alerts for us?

This is particularly with the prominence on the general panel page, maybe if we want to show it to general Experimenters perhaps we could add traffic light for it when they arrive?

yanirs:
> Should this be a debug mode only feature, with alerts for us?

I think it's worth exposing to experimenters because questions about cache propagation are inevitable, especially since sandbox updates require manual action. I just added [a wiki section](https://github.com/Automattic/experimentation-platform/wiki/Experimenter's-Guide#the-file-system-cache) and linked to it in https://github.com/Automattic/abacus/pull/465/commits/4ebe1dab205f38a8c0b663b71d1ab1ba7858ba23, which should cover the most common questions. 

> This is particularly with the prominence on the general panel page, maybe if we want to show it to general Experimenters perhaps we could add traffic light for it when they arrive?

Maybe emoji? Added in https://github.com/Automattic/abacus/pull/465/commits/90dc2becb84c07f9f105c1ae1027ce161a2be834. I think they draw attention to the cache status without making the view too busy:

image

image

jessie-ross:
Hmm, maybe could we reduce the messaging down to this form in the status line:

> 🟠 running  - ⚠️ [Recently changed, may take 10 minutes to propagate.](#)

- Next to the status - makes it clear what has been impacted.
- Where an Experimenter would check if their experiment variation isn't showing up.
- Out of sight when nothing is wrong - users don't need the fresh information around, especially with the improvements in #466

yanirs:
> Hmm, maybe could we reduce the messaging down to this form in the status line:
> 
> > :orange_circle: running  - :warning: [Recently changed, may take 10 minutes to propagate.](#)
> 
> * Next to the status - makes it clear what has been impacted.
> * Where an Experimenter would check if their experiment variation isn't showing up.
> * Out of sight when nothing is wrong - users don't need the fresh information around, especially with the improvements in #466
> * Experimenters don't need to know the state names.
> 
> Do Experimenters need to know the `Missing` state?

Thanks @jessie-ross. I think that would work well if the cache updated automatically on sandboxes, but given that it doesn't, it's useful to have a persistent indicator. For example, it may be the case that the production cache is fresh, but manual action is needed to update it on the experimenter's sandbox. With the freshness indicator and the "learn more" link, they have a chance of successful troubleshooting even if they forget about the file system cache.

I think that there's no harm in letting experimenters know the status names, as they're consistent with the code and wiki description, and the assignment cache status is mostly aimed at developers.

I just deployed the API change and tested the implementation manually with production data, so it's good to go if you're happy with it. Otherwise, we can continue the discussion. :slightly_smiling_face:

jessie-ross:
> Thanks @jessie-ross. I think that would work well if the cache updated automatically on sandboxes, but given that it doesn't, it's useful to have a persistent indicator. For example, it may be the case that the production cache is fresh, but manual action is needed to update it on the experimenter's sandbox. With the freshness indicator and the "learn more" link, they have a chance of successful troubleshooting even if they forget about the file system cache.

That's a good point, what about something like this?:

> 🟠 running - [⚠️ Recently changed, may take 10 minutes to propagate.]()

And when that isn't showing, show this:

> 🟠 running - [Changes not showing? Troubleshooting instructions are available!]()

Both with de-emphasised text.

If you still aren't feeling it, I am happy to go with the cache line in the table, but it might be important here to change the `✅ Fresh` text to something like "Changes not showing? Find out more here.", otherwise they might see the fresh and assume that the caching is not the problem.

jessie-ross (approved):

withinboredom:
FWIW, I like Jessie's copy better. The vast majority of people running experiments aren't devs, and just want to know the status. For devs implementing an experiment, having some links to docs makes a lot sense.

yanirs:
Hmm... I tried putting it on the same row but I think it makes it a bit too wordy, so it interferes with seeing the experiment status (I also tried Jessie's wording):

image

> If you still aren't feeling it, I am happy to go with the cache line in the table, but it might be important here to change the ✅ Fresh text to something like "Changes not showing? Find out more here.", otherwise they might see the fresh and assume that the caching is not the problem.

Yeah, that's a good point. I changed it to "✅ Fresh: Production cache is up to date, but manual sandbox updates may be needed ([learn more]())"

I'll go with this for now, but you can play with the styling or moving the text in a follow up PR if you want.

PR #467, opened 2021-03-04
Add Akismet and LOHP as supported platforms

yanirs:
Does what it says in the title.

This is a part of addressing 588-gh-Automattic/experimentation-platform and should only be merged once the backend changes are deployed.

## How has this been tested?

Ran the dev server and manually checked the drop-down menu when creating an experiment:

image


jessie-ross (approved):
LGTM :)

PR #468, opened 2021-03-04
Fix tab history behaviour

jessie-ross:
**This PR fixes the Experiment tab history behaviour.**

By adding `replace` to the links in the tab it stops flooding the history causing you to need multiple back presses to get back to the experiment list.

## How has this been tested?


yanirs:
Hmm... This behaviour is a bit surprising now: Clicking on an experiment, then Results, then Code Setup, then pressing back, I get back to the experiment list. This isn't what I expect as a user when I see the URL change. 

Going back to the experiment list always easy, as it's a matter of clicking on the logo or Experiments link, so I'm not sure that this is solving a problem that we actually have.

jessie-ross:
> Is this something that someone requested?

This is my own personal annoyance. Now that I think about it the root cause might be the inability to open experiments in a new tab.

PR #469, opened 2021-03-09
Implement AggregateRecommendation

jessie-ross:
**This PR implements AggregateRecommendation.**
- See the next PR in stack for context
- Consolidates all the recommendation code in one place

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/AggregateRecommendationDisplay.test.tsx
+++ b/src/components/experiments/single-view/results/AggregateRecommendationDisplay.test.tsx
@@ -1,0 +1,94 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+import React from 'react'
+
+import { AggregateRecommendationType } from 'src/lib/analyses'
+import { render } from 'src/test-helpers/test-utils'
+
+import AggregateRecommendationDisplay from './AggregateRecommendationDisplay'
+
+test('renders NotAnalyzedYet correctly', () => {
+  const { container } = render(
+    <AggregateRecommendationDisplay
+      aggregateRecommendation={{
+        type: AggregateRecommendationType.NotAnalyzedYet,
+      }}
+      // @ts-ignore
+      experiment={{}}
+    />,
+  )
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      Not Analyzed Yet
+    </div>
+  `)
+})
+
+test('renders ManualAnalysisRequired correctly', () => {
+  const { container } = render(
+    <AggregateRecommendationDisplay
+      aggregateRecommendation={{
+        type: AggregateRecommendationType.ManualAnalysisRequired,
+      }}
+      // @ts-ignore
+      experiment={{}}
+    />,
+  )
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      Manual Analysis Required
+    </div>
+  `)
+})
+
+test('renders Inconclusive correctly', () => {
+  const { container } = render(
+    <AggregateRecommendationDisplay
+      aggregateRecommendation={{
+        type: AggregateRecommendationType.Inconclusive,
+      }}
+      // @ts-ignore
+      experiment={{}}
+    />,
+  )
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      Inconclusive
+    </div>
+  `)
+})
+
+test('renders DeployEither correctly', () => {
+  const { container } = render(
+    <AggregateRecommendationDisplay
+      aggregateRecommendation={{
+        type: AggregateRecommendationType.DeployEither,
+      }}
+      // @ts-ignore
+      experiment={{}}
+    />,
+  )
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      Deploy Either Variation
+    </div>
+  `)
+})
+
+test('renders Deploy correctly', () => {
+  const { container } = render(
+    <AggregateRecommendationDisplay
+      aggregateRecommendation={{
+        type: AggregateRecommendationType.Deploy,
+        variationId: 123,
+      }}
+      // @ts-ignore
+      experiment={{ variations: [{ variationId: 123, name: 'variation_name_123' }] }}
+    />,
+  )
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      Deploy 
+      variation_name_123
+    </div>
+  `)
+})
--- a/src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
+++ b/src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
@@ -1,0 +1,40 @@
+import React from 'react'
+
+import { AggregateRecommendation, AggregateRecommendationType } from 'src/lib/analyses'
+import { ExperimentFull } from 'src/lib/schemas'
+
+/**
+ * Displays an AggregateRecommendation.
+ */
+export default function AggregateRecommendationDisplay({
+  aggregateRecommendation,
+  experiment,
+}: {
+  aggregateRecommendation: AggregateRecommendation
+  experiment: ExperimentFull
+}): JSX.Element {
+  switch (aggregateRecommendation.type) {
+    case AggregateRecommendationType.ManualAnalysisRequired:
+      return <>Manual Analysis Required</>
+    case AggregateRecommendationType.NotAnalyzedYet:
+      return <>Not Analyzed Yet</>
+    case AggregateRecommendationType.Inconclusive:
+      return <>Inconclusive</>
+    case AggregateRecommendationType.DeployEither:
+      return <>Deploy Either Variation</>
+    case AggregateRecommendationType.Deploy: {
+      const chosenVariation = experiment.variations.find(
+        (variation) => variation.variationId === aggregateRecommendation.variationId,
+      )
+      // istanbul ignore next; Typeguard
+      if (!chosenVariation) {
+        throw new Error('No match for chosenVariationId among variations in experiment.')
+      }
+
+      return <>Deploy {chosenVariation.name}</>
+    }
+    // istanbul ignore next; Shouldn't occur
+    default:
+      throw new Error('Missing AggregateRecommendationType.')
+  }
+}
--- a/src/lib/analyses.ts
+++ b/src/lib/analyses.ts
@@ -1,4 +1,6 @@
-import { AnalysisStrategy, RecommendationWarning } from './schemas'
+import _ from 'lodash'
+
+import { Analysis, AnalysisStrategy, RecommendationWarning } from './schemas'
 
 /**
  * Mapping from AnalysisStrategy to human-friendly descriptions.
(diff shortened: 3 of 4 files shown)


jessie-ross:
AggregateRecommendation is also going to be extended to contain `endExperiment` and `warnings` for recommendations on ending the experiment.

src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
```diff
@@ -0,0 +1,40 @@
...
+      // istanbul ignore next; Typeguard
+      if (!chosenVariation) {
+        throw new Error('No match for chosenVariationId among variations in experiment.')
+      }
+
+      return <>Deploy {chosenVariation.name}</>
+    }
+    // istanbul ignore next; Shouldn't occur
+    default:
+      throw new Error('Missing AggregateRecommendationType.')
```
yanirs, line 38: Is this needed at all? Doesn't TypeScript verify that the `case`s cover all options?

src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
```diff
@@ -0,0 +1,40 @@
...
+    case AggregateRecommendationType.DeployEither:
+      return <>Deploy Either Variation</>
+    case AggregateRecommendationType.Deploy: {
+      const chosenVariation = experiment.variations.find(
+        (variation) => variation.variationId === aggregateRecommendation.variationId,
+      )
+      // istanbul ignore next; Typeguard
+      if (!chosenVariation) {
+        throw new Error('No match for chosenVariationId among variations in experiment.')
+      }
```
yanirs, line 32: Can we replace this with casting since it should never happen?

src/lib/analyses.ts
```diff
@@ -19,3 +21,59 @@ export const RecommendationWarningToHuman = {
...
+  NotAnalyzedYet = 'NotAnalyzedYet',
+  Inconclusive = 'Inconclusive',
+  DeployEither = 'DeployEither',
+  Deploy = 'Deploy',
+}
+
+export interface AggregateRecommendation {
+  type: AggregateRecommendationType
+  variationId?: number
+}
```
yanirs, line 36: Looking at the full stack, I don't see the need to add these new types. It feels like a leaky abstraction where `AggregateRecommendationType` is essentially a string, and the new levels of indirection seem to complicate things.

I think it'd be better with a flat interface like:

```typescript
interface AggregateRecommendation {
  description: string,
  manualAnalysisRequired: boolean,
  chosenVariation: Variation | null,
}
```

This would obviate the need for `AggregateRecommendationDisplay` and encode the only semantic question that we seem to care about (manual analysis required?) as a boolean.

What do you think?

src/lib/analyses.ts
```diff
@@ -19,3 +21,59 @@ export const RecommendationWarningToHuman = {
...
+
+/**
+ * Returns the aggregate recommendation over analyses of different analysis strategies.
+ * @param analyses Analyses of different strategies for the same day.
+ */
+export function getAggregateRecommendation(analyses: Analysis[]): AggregateRecommendation {
+  const recommendationChosenVariationIds = analyses
+    .map((analysis) => analysis.recommendation?.chosenVariationId)
+    .filter(Number)
+  const recommendationConflict = _.uniq(recommendationChosenVariationIds).length > 1
```
yanirs, line 46: Doesn't this change the current semantics where we consider `null` to be in conflict with a recommendation to deploy a specific variation?

yanirs (commented):
See inline comments. My two main concerns are:

* This refactor adds levels of indirection over what's essentially a mapping from `analyses` to a string. I think it can be simplified.
* `@ts-ignore` should be used sparingly, not for cases where we can easily produce the real types.

src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
jessie-ross: Typescript will not ensure that the `case`s cover all options. 
This would fail silently if `AggregateRecommendationType` was extended and there was no default case.

src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
jessie-ross: Ah this isn't a type-guard actually (more like a 'shouldn't occur') as we have nothing in the front-end guaranteeing the variationId in the recommendation exists in the experiment variations, e.g. from a bug in the server.

src/lib/analyses.ts
```diff
@@ -19,3 +21,59 @@ export const RecommendationWarningToHuman = {
...
+  if (!recommendation.chosenVariationId) {
+    return {
+      type: AggregateRecommendationType.DeployEither,
+    }
+  }
+
+  return {
+    type: AggregateRecommendationType.Deploy,
+    variationId: recommendation.chosenVariationId,
+  }
```
withinboredom, line 78: This fall-through style if-statement smells ripe for some pattern matching.

src/lib/analyses.ts
jessie-ross: Intuitively I much prefer the "type" pattern, partly it helps separate UI and core logic, partly I like derived data to be reusable, partly it feels extensible is several directions. 

Practically I don't see it adding complexity (in the intertwined sense), and don't see any particular reason it would be a leaky abstraction (that would mean needing to reach into the Analysis data or the Recommendation data, which is what this abstracts away).

From my perspective the downside is needing an extra component to display the data, and I think the benefit of separating UI from core functionality is a worthy tradeoff here.

src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
yanirs: Why not test it then? :slightly_smiling_face:

src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
yanirs: OK, then shall we test it?

src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
```diff
@@ -0,0 +1,40 @@
...
+}): JSX.Element {
+  switch (aggregateRecommendation.type) {
+    case AggregateRecommendationType.ManualAnalysisRequired:
+      return <>Manual Analysis Required</>
+    case AggregateRecommendationType.NotAnalyzedYet:
+      return <>Not Analyzed Yet</>
+    case AggregateRecommendationType.Inconclusive:
+      return <>Inconclusive</>
+    case AggregateRecommendationType.DeployEither:
+      return <>Deploy Either Variation</>
```
yanirs, line 24: Why were the phrases changed to _Title Case_ from _Sentence case_?

src/lib/analyses.ts
yanirs: Hmm... Yeah, maybe calling it a leaky abstraction was the wrong term. I see the type as more of an ill-fitting abstraction because it's tightly coupled to the display string and logic. Also, the "type" names are pretty ambiguous, e.g., I doubt that anyone can infer the difference between `Deploy` and `DeployEither` without reading all the code.

The extra component and the type enum can be easily avoided by making the `description` a field that'd be populated by `getAggregateRecommendation`.

The point about extensibility makes sense, though. Maybe it can be improved with better choice of names? Specifically:

* `type` and `AggregateRecommendationType` can be named `summary` and `AggregateRecommendationSummary` or something similar (conclusion/suggestion/advice/guidance?) &ndash; it's not really the type of the aggregate recommendation
* `variationId` should be named `chosenVariationId` for consistency with the `Recommendation` field
* `NotAnalyzedYet` can be named `MissingAnalysis`
* `DeployEither` can be named `DeployAnyVariation`
* `Deploy` can be named `DeployChosenVariation`

src/lib/analyses.ts
```diff
@@ -19,3 +21,59 @@ export const RecommendationWarningToHuman = {
...
+  if (recommendationConflict) {
+    return {
+      type: AggregateRecommendationType.ManualAnalysisRequired,
+    }
+  }
+
+  let recommendation = analyses.find((analysis) => analysis.recommendation?.chosenVariationId)?.recommendation
+  if (!recommendation) {
+    recommendation = analyses.find((analysis) => analysis.recommendation)?.recommendation
+  }
```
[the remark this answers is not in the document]
jessie-ross: I think I had the wrong behaviour here, see the new code

src/lib/analyses.ts
jessie-ross: Went with `decision: AggregateRecommendationDecision`

src/components/experiments/single-view/results/AggregateRecommendationDisplay.test.tsx
```diff
@@ -0,0 +1,140 @@
+import React from 'react'
+
+import { AggregateRecommendationDecision } from 'src/lib/analyses'
+import Fixtures from 'src/test-helpers/fixtures'
+import { render } from 'src/test-helpers/test-utils'
+
+import AggregateRecommendationDisplay from './AggregateRecommendationDisplay'
+
+test('renders NotAnalyzedYet correctly', () => {
```
yanirs, line 9:
```suggestion
test('renders MissingAnalysis correctly', () => {
```

src/components/experiments/single-view/results/AggregateRecommendationDisplay.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    />,
+  )
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      Inconclusive
+    </div>
+  `)
+})
+
+test('renders DeployEither correctly', () => {
```
yanirs, line 57:
```suggestion
test('renders DeployAnyVariation correctly', () => {
```

src/components/experiments/single-view/results/AggregateRecommendationDisplay.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    />,
+  )
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      Deploy either variation
+    </div>
+  `)
+})
+
+test('renders Deploy correctly', () => {
```
yanirs, line 73:
```suggestion
test('renders DeployChosenVariation correctly', () => {
```

src/components/experiments/single-view/results/AggregateRecommendationDisplay.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    <AggregateRecommendationDisplay
+      aggregateRecommendation={{
+        decision: AggregateRecommendationDecision.DeployChosenVariation,
+        chosenVariationId: 123,
+      }}
+      experiment={Fixtures.createExperimentFull({
+        variations: [
+          {
+            variationId: 123,
+            name: 'varition_name_123',
```
yanirs, line 84:
```suggestion
            name: 'variation_name_123',
````

File: src/components/experiments/single-view/results/AggregateRecommendationDisplay.test.tsx [hunk 12]
```diff
@@ -0,0 +1,140 @@
...
+            isDefault: false,
+          },
+        ],
+      })}
+    />,
+  )
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      Deploy 
+      varition_name_123
```
yanirs, line 95:
```suggestion
      variation_name_123
```

src/components/experiments/single-view/results/AggregateRecommendationDisplay.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    </div>
+  `)
+})
+
+test('throws error for missing varitionId', () => {
```
yanirs, line 100:
```suggestion
test('throws error for missing chosenVariationId', () => {
```

yanirs (approved):
Thanks for addressing all the comments!

LGTM with the minor inline suggestions.

PR #471, opened 2021-03-09
Add top level result panel

jessie-ross:
This PR adds top level panel including:
- Total participants
- AggregateRecommendation

Screen Shot 2021-03-09 at 12 59 54 pm

## How has this been tested?


src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -129,6 +171,31 @@ export default function ActualExperimentResults({
...
+  // ### Top Level Stats
+
+  const primaryMetricLatestAnalysesByStrategy = _.mapValues(
+    primaryMetricAssignmentAnalysesData.analysesByStrategyDateAsc,
+    _.last.bind(null),
+  )
+
+  const totalParticipants = Object.values(primaryMetricLatestAnalysesByStrategy)
+    .map((primaryMetricLatestAnalysis) => {
+      // istanbul ignore next; trivial: If there is a missing strategy, count it as zero.
```
yanirs, line 183: I think it's worth testing.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -129,6 +171,31 @@ export default function ActualExperimentResults({
...
   ]
 
+  // ### Top Level Stats
+
+  const primaryMetricLatestAnalysesByStrategy = _.mapValues(
+    primaryMetricAssignmentAnalysesData.analysesByStrategyDateAsc,
+    _.last.bind(null),
+  )
+
+  const totalParticipants = Object.values(primaryMetricLatestAnalysesByStrategy)
```
yanirs, line 181: Can you please explain why this is done this way? It's unclear why we need all this code to get the total participants.

jessie-ross: Ah, this should actually just be assignments, as it is the max of all participant totals.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -130,6 +172,21 @@ export default function ActualExperimentResults({
...
 
+  // ### Top Level Stats
+
+  const primaryMetricLatestAnalysesByStrategy = _.mapValues(
+    primaryMetricAssignmentAnalysesData.analysesByStrategyDateAsc,
+    _.last.bind(null),
+  )
+  // We want the max participants so we use IttPure
+  const totalParticipants =
+    primaryMetricLatestAnalysesByStrategy[AnalysisStrategy.IttPure]?.participantStats['total'] ?? 0
```
yanirs, line 183: Why do we want the `IttPure` count? It seems likely to be confusing that the number doesn't match the latest variation sum, especially given that there's often a large difference between assigned and exposed numbers. 

If we change the wording below from `total participants` to `analyzed participants` and show the total from the `latestPrimaryMetricAnalysis` then everything should be consistent.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -218,16 +274,54 @@ export default function ActualExperimentResults({
...
+                    />
+                  </Typography>
+                  <Typography variant='subtitle1'>
+                    <strong>primary metric</strong> recommendation
+                  </Typography>
+                </div>
+              </>
+            )}
+          </Paper>
+          {/* <Paper className={classes.summaryHealthPaper}></Paper> */}
```
withinboredom, line 322: Can we remove this?

withinboredom (approved):

src/components/experiments/single-view/results/ActualExperimentResults.tsx
jessie-ross: I've switched to analysed participants

src/components/experiments/single-view/results/ActualExperimentResults.tsx
jessie-ross: Well it get's uncommented in the next PR in the stack

jessie-ross: Oh actually it doesn't, I'll just remove it for now

PR #473, opened 2021-03-10
Add Experiment Health Utils

jessie-ross:
**This PR adds utils for Experiment Health:**
- `getParticipantCounts`
- `getExperimentHealthStats`
   - Useful Ratios
   - Probabilities

This will be used for:
- Displaying the stats in debug mode
- Displaying "traffic lights"

Part of #419

## How has this been tested?


src/lib/analyses.ts
```diff
@@ -77,3 +81,115 @@ export function getAggregateRecommendation(
...
+    variations: Object.fromEntries(
+      experiment.variations.map(({ variationId }) => [
+        variationId,
+        getParticipantCountsSetForParticipantStatsKey(`variation_${variationId}`, analysesByStrategy),
+      ]),
+    ),
+  }
+}
+
+export function getExperimentHealthStats(experiment: ExperimentFull, analysesByStrategy: AnalysesByStrategy) {
```
yanirs, line 130: Looks like the build is failing because this function doesn't declare its return value.

src/lib/analyses.ts
```diff
@@ -77,3 +81,115 @@ export function getAggregateRecommendation(
...
+    assigned: assigned,
+    assignedCrossovers:
+      assigned - (analysesByStrategy[AnalysisStrategy.MittNoCrossovers]?.participantStats[participantStatsKey] ?? 0),
+    assignedSpammers:
+      assigned - (analysesByStrategy[AnalysisStrategy.MittNoSpammers]?.participantStats[participantStatsKey] ?? 0),
+    exposed: analysesByStrategy[AnalysisStrategy.PpNaive]?.participantStats[participantStatsKey] ?? 0,
+  }
+}
+
+export function getParticipantCounts(
```
yanirs, line 115: Can you please document the exported functions?

src/lib/analyses.ts
```diff
@@ -77,3 +81,115 @@ export function getAggregateRecommendation(
...
+  experiment: ExperimentFull,
+  analysesByStrategy: AnalysesByStrategy,
+): { total: CountsSet; variations: Record<number, CountsSet> } {
+  return {
+    total: getParticipantCountsSetForParticipantStatsKey('total', analysesByStrategy),
+    variations: Object.fromEntries(
```
yanirs, line 121: Maybe change `variations` to `byVariation` or something similar? It sounds like it'd be the actual variation numbers.

src/utils/math.ts
```diff
@@ -0,0 +1,19 @@
+import { abs, erf } from 'mathjs'
+
+/**
+ * Assuming a distribution of X ~ Binomial(n, p), returns the probability of the number of trials being equal to x or more extreme than x.
+ *
+ * @param successfulTrials number of successful trials
+ * @param totalTrials number of total trials
+ * @param probabilityOfSuccess probability of success
+ */
+export function binomialProbValue(successfulTrials: number, totalTrials: number, probabilityOfSuccess: number): number {
```
yanirs, line 10: Can you please change this to accept an object? It's easy to get the parameter order wrong when it's three numbers.

src/utils/math.ts
```diff
@@ -0,0 +1,19 @@
...
+  const mean = totalTrials * probabilityOfSuccess
+  const variance = totalTrials * probabilityOfSuccess * (1 - probabilityOfSuccess)
+  // By the CLT, B ~ Binomial(n, p) is approximated well enough by X ~ N(mu, sigma)) for n > 30
+  // We don't care about the accuracy for n <= 30 so we let them be.
+  // And if Y ~ N(0, 1/2) then Y = (X - mu)/(2 * sigma)^(1/2)
+  const y = abs(successfulTrials - mean) / Math.sqrt(2 * variance)
+  // Since erf gives the probabilty of [-y, y] on Y, and pValue is the union of [-inf, -y] and [y, inf]:
+  return 1 - erf(y)
```
yanirs, line 18: Do you have a citation for this approximation?

src/utils/math.ts
```diff
@@ -0,0 +1,19 @@
+import { abs, erf } from 'mathjs'
```
yanirs, line 1: From a quick search, I wonder if we can use `stdlib` instead? It has full support for [the binomial distribution](https://stdlib.io/docs/api/v0.0.90/@stdlib/stats/base/dists/binomial), and I'm guessing we can only use the parts we need (though I haven't checked).

jessie-ross: Ah, I didn't see that when I last searched 😮

jessie-ross: Hmm, seems like there are some subtle differences between them

jessie-ross: Looks like it is just the approximation is different for smaller numbers, it's good to use :)

package.json
```diff
@@ -18,6 +18,8 @@
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.56",
     "@rehooks/component-size": "^1.0.3",
+    "@stdlib/stdlib": "0.0.93",
+    "@types/mathjs": "^6.0.11",
```
yanirs, line 22: Do we still need this?

src/lib/analyses.ts
```diff
@@ -77,3 +81,158 @@ export function getAggregateRecommendation(
...
+      assigned - (analysesByStrategy[AnalysisStrategy.MittNoSpammers]?.participantStats[participantStatsKey] ?? 0),
+    exposed: analysesByStrategy[AnalysisStrategy.PpNaive]?.participantStats[participantStatsKey] ?? 0,
+  }
+}
+
+/**
+ * Gets participant counts for an Experiment
+ *
+ * @param experiment
+ * @param analysesByStrategy
```
yanirs, line 119: Can drop the param descriptions if have nothing interesting to say about them. :slightly_smiling_face: 

```suggestion
```

src/utils/binomial-test.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '@stdlib/stats/binomial-test' {
```
yanirs, line 1: Can we get these types from a package? `stdlib` has TypeScript docs, so I assumed it has something we can use.

src/utils/math.ts
```diff
@@ -0,0 +1,20 @@
...
+export function binomialProbValue({
+  successfulTrials,
+  totalTrials,
+  probabilityOfSuccess,
+}: {
+  successfulTrials: number
+  totalTrials: number
+  probabilityOfSuccess: number
+}): number {
+  return binomialTest(successfulTrials, totalTrials, { p: probabilityOfSuccess }).pValue
```
yanirs, line 19: How about we just inline the call (i.e., use `binomialTest` directly)? Then we won't need to worry about `math.test.ts` either. :slightly_smiling_face:

src/lib/analyses.test.ts
```diff
@@ -304,3 +304,205 @@ describe('getAggregateRecommendation', () => {
...
+              variation_2: 27,
+            },
+          }),
+        },
+      ),
+    ).toEqual({
+      probabilities: {
+        byVariationId: {
+          '1': {
+            assignedDistributionMatchingAllocated: 0.000011583130623216142,
```
yanirs, line 468: Looks like the tests are now failing on the comparisons here. If this varies across environments, it's worth looking into the solution suggested in https://github.com/facebook/jest/issues/3654 (https://github.com/maasencioh/jest-matcher-deep-close-to).

jessie-ross: ...I just forgot to push my last commit

src/utils/binomial-test.d.ts
jessie-ross: Yeah I spent a little while looking for how to do this but think it is time to give up for now, it looks like they are incredibly non-standard although they are working at fixing things.

yanirs (approved):
:ship:

src/lib/analyses.ts
```diff
@@ -1,4 +1,14 @@
-import { Analysis, AnalysisStrategy, RecommendationWarning } from './schemas'
+import _ from 'lodash'
+
+import { Analysis, AnalysisStrategy, ExperimentFull, RecommendationWarning } from './schemas'
+
+// I can't get stdlib to work as an import...:
```
kgryte, line 5: @jessie-ross Heya! stdlib maintainer here. We have an experimental ESM distribution, which may be of interest (see https://github.com/stdlib-js/esm), if you are needing full ESM compatibility. Otherwise, the default `@stdlib/stdlib` distribution is CommonJS, and thus requires the use of `require`.

jessie-ross: Thanks @kgryte for the heads up and impressive lib! We like to stay on stables and we will switch over when it is ready :-)

kgryte: Sounds good!

PR #477, opened 2021-03-17
Add health stats in debug mode

jessie-ross:
**This PR adds health stats at the bottom of results in debug mode.**

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/ActualExperimentResults.tsx
+++ b/src/components/experiments/single-view/results/ActualExperimentResults.tsx
@@ -16,8 +16,8 @@ import MaterialTable from 'material-table'
 import { PlotData } from 'plotly.js'
 import React, { useState } from 'react'
 import Plot from 'react-plotly.js'
-import DebugOutput from 'src/components/general/DebugOutput'
 
+import DebugOutput from 'src/components/general/DebugOutput'
 import {
   AggregateRecommendation,
   AggregateRecommendationDecision,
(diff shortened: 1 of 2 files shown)


yanirs:
Hmm... I think it'd be better to add these stats to the debug page rather than the result page, and it'd also be good to have the cover the new code with tests.

jessie-ross:
> Hmm... I think it'd be better to add these stats to the debug page rather than the result page, and it'd also be good to have the cover the new code with tests.

I'd eventually like to remove the Debug page, I think it can all be in one page with the Advanced button behind it.

The only thing really missing is providing a table for each MetricAssignment pivoting on Strategy.

yanirs:
Taking another look at the JSON, I think it'd be even hard for us to parse it, as it relies on variation IDs. It's also hard to judge the necessity of all the ratios without knowing where it's going. Maybe instead of showing it on the page, we can `console.log()` it?

jessie-ross:
Yeah it is not meant to be a permanent solution, just a quick cheap way of presenting data for us now. 
I am using it when I look at and compare the A/As

yanirs:
> Yeah it is not meant to be a permanent solution, just a quick cheap way of presenting data for us now.
> I am using it when I look at and compare the A/As

OK, if it's a temporary solution I think it's all good. Before merging, can you please add an inline comment that it's temporary?

yanirs (approved):

PR #479, opened 2021-03-23
Implement experiment health indicators

jessie-ross:
This PR implements experiment health indicators: taking stats, making them user friendly and adds indications. 

In a subsequent PR it will look like this:
Screen Shot 2021-03-24 at 2 20 29 am

Part of #419 

## How has this been tested?


yanirs:
I need to step out for a couple of hours, but I left a few suggestions in #481. I'll finish reviewing when I get back.

src/lib/analyses.ts
```diff
@@ -237,3 +239,171 @@ export function getExperimentHealthStats(
...
+    link?: string
+    indicationBrackets: Array<IndicationBracket>
+  }
+
+  const indicatorDefinitions: IndicatorDefinition[] = [
+    {
+      name: 'Assignment distribution matching allocated',
+      value: minVariationProbabilities.assignedDistributionMatchingAllocated,
+      unit: HealthIndicatorUnit.Pvalue,
+      link: '',
```
yanirs, line 305: Why do we set the links to empty strings? Can't we just omit them since `link` is optional?

jessie-ross: Ah I was going to address this in the next PR

yanirs: In general, it's usually better to leave out unused fields. :slightly_smiling_face:

yanirs (approved):
Approving to unblock this since I'll be AFK for the rest of the week. We can discuss my comments and tweaks on #481 next week.

PR #481, opened 2021-03-24
Suggestions for PR #479

yanirs:
Suggestions I've played with as part of reviewing PR #479. The main change is to simplify things by using ordinal enums as array indices. See inline comments for details.

## How has this been tested?


src/lib/analyses.test.ts
```diff
@@ -1,6 +1,7 @@
 import Fixtures from 'src/test-helpers/fixtures'
 
 import * as Analyses from './analyses'
+import { HealthIndicatorUnit, HealthLevel } from './analyses'
```
yanirs, line 4: Changed to use the enums rather than their string values, but other than that the tests are conceptually the same.

src/lib/analyses.ts
```diff
@@ -240,15 +240,15 @@ export function getExperimentHealthStats(
   }
 }
 
-export enum HealthIndication {
-  Nominal = 'Nominal',
-  PossibleIssue = 'PossibleIssue',
-  ProbableIssue = 'ProbableIssue',
+export enum HealthLevel {
```
yanirs, line 243: Reordered from worst to best, renamed, and made this an ordinal enum.

I'm not sure why we don't default to ordinals in general. Any reason for that?

src/lib/analyses.ts
```diff
@@ -259,7 +259,7 @@ export interface HealthIndicator {
   value: number
   unit: HealthIndicatorUnit
   link?: string
-  indication: HealthIndication
+  indication: HealthLevel
```
yanirs, line 262: If we go with `HealthLevel` as the enum name, the key should also be changed to `level` or `healthLevel`.

src/lib/analyses.ts
```diff
@@ -284,126 +284,60 @@ export function getExperimentHealthIndicators(experimentHealthStats: ExperimentH
...
-        {
-          max: 0.05,
-          indication: HealthIndication.PossibleIssue,
-        },
-        {
-          max: 1,
-          indication: HealthIndication.Nominal,
-        },
-      ],
+      sortedLevelThresholds: probabilityLevelThresholds,
```
yanirs, line 302: This is the main part of the PR: I find the flat threshold arrays easier to parse, and they're definitely less verbose. :slightly_smiling_face: 

If we go with it, we should add some inline documentation to explain how it works.

src/lib/analyses.ts
```diff
@@ -284,126 +284,60 @@ export function getExperimentHealthIndicators(experimentHealthStats: ExperimentH
...
-  return indicatorDefinitions.map(({ value, indicationBrackets, ...rest }) => ({
+  return indicatorDefinitions.map(({ value, sortedLevelThresholds, unit, ...rest }) => ({
     value,
-    indication: (_.sortBy(indicationBrackets, 'max').find((bracket) => value <= bracket.max) as IndicationBracket)
-      .indication,
+    unit,
+    indication:
+      unit === HealthIndicatorUnit.Ratio
+        ? sortedLevelThresholds.length - _.sortedIndex(sortedLevelThresholds, value)
+        : _.sortedIndex(sortedLevelThresholds, value),
```
yanirs, line 340: I'm not a fan of the way the ordering works different for ratios and probabilities.

I wanted to keep this suggestion PR focused, but we can simplify things by returning the complementary probabilities (i.e., `1 - p_value`). What do you think?

src/lib/analyses.ts
jessie-ross: > I'm not sure why we don't default to ordinals in general. Any reason for that?

It shows up better in debugging and errors, instead of:

```
{
   indication: 2
}
```

We get:

```
{
   indication: 'Healthy',
}
```

src/lib/analyses.ts
jessie-ross: Hmm, not a fan of the enum being used like this if we do go this way.

src/lib/analyses.ts
jessie-ross: I like the factorisation of the common `probabilityLevelThresholds`.

I was thinking about using an array like this but I think the brackets is simplest here, especially when the threshold's meaning changes depending on the unit. 

I would be looking to expand this in the future to e.g.:
- LogStash error rates
- MC Stat flow rate
- Warnings

And we could work out a better system then.

src/lib/analyses.ts
```diff
@@ -240,15 +240,15 @@ export function getExperimentHealthStats(
...
 }
 
-export enum HealthIndication {
-  Nominal = 'Nominal',
-  PossibleIssue = 'PossibleIssue',
-  ProbableIssue = 'ProbableIssue',
+export enum HealthLevel {
+  Unhealthy,
+  Suspicious,
+  Healthy,
```
jessie-ross, line 246: Ehh, I am ok with this, but it is a bit confusing to me as I'd think Unhealthy and Suspicious would be swapped.

Maybe Suspicious -> PossiblyUnhealthy?

I was thinking about using "Health" but liked Nominal

src/lib/analyses.ts
yanirs: Hmm... The debugging consideration makes sense, though it would come with a bit of a performance hit in production. I don't mind it, though, as it's likely to be negligible.

However, in this case, I think that an ordinal enum makes more sense because the items are ordered.

src/lib/analyses.ts
yanirs: The single-dimensional array approach would work for any expansion, as we'd be mapping a sorted array of thresholds to an ordinal enum of health levels. 

I find the brackets more complicated and error-prone because they're not actually brackets (which would imply min/max) &ndash; they map thresholds to levels, but there's no guarantee that the mapping is one-to-one. With the threshold array we get a "natural" mapping from ordinal indices to thresholds.

I agree that the threshold meaning changes are not great, but that's also a problem with the bracket approach. I had to actually change the code to realise that the ratios are defined in reverse. I think that's better addressed by using the complements of the probabilities, so we can assume that higher values are always worse.

src/lib/analyses.ts
yanirs: Why not? It's fairly common to use ordinal enums to index mappings.

src/lib/analyses.ts
yanirs: `PossiblyUnhealthy` sounds good.

> I was thinking about using "Health" but liked "Nominal" as these shouldn't be used to rule things out.

In practice, I think it'd be used to rule things out. :slightly_smiling_face: 

It was actually the "nominal" bit that confused me the most because of the other definitions:
image

jessie-ross: > because of the other definitions

Ah true

src/lib/analyses.ts
jessie-ross: > The single-dimensional array approach would work for any expansion, as we'd be mapping a sorted array of thresholds to an ordinal enum of health levels.

E.g. for flow rates we might want to have:
- No flow: Error
- Little flow: Warning
- Normal flow: Nominal
- Too high flow: Warning
- Extreme flow: Error

(but not limited to that.)

We also might want to have other states (e.g. `too low`, `too high`), and making it ordinal and using it like this really limits it extensibility wise. If we still want to use an array approach like this we could use an array of Indications `[Unhealthy, Suspicious, Healthy]` and use the array order, I find it very strange that we are doing calculations and getting 

> I find the brackets more complicated and error-prone because they're not actually brackets (which would imply min/max) – they map thresholds to levels, but there's no guarantee that the mapping is one-to-one. With the threshold array we get a "natural" mapping from ordinal indices to thresholds.

I don't see this being a problem in practice. Also it is only not one-to-one if two brackets have the exactly the same `max`. We could also always add in a check to ensure it is so.

> I had to actually change the code to realise that the ratios are defined in reverse.

I wouldn't say your code is less tricky, perhaps a comment would help here?

src/lib/analyses.ts
jessie-ross: I looked this up a bit more and nominal is used generally in engineering and comes from the 3rd definition you mention: "(of a quantity or dimension) stated or expressed but not necessarily corresponding exactly to the real value." 
See also https://english.stackexchange.com/questions/184876/how-did-nominal-come-to-mean-within-acceptable-tolerances

Still don't mind changing it.

jessie-ross: Actually since this is an internal enum and we have separated UI for it I think it might be beneficial to use more technical terms as they are a bit more self-documenting.

src/lib/analyses.ts
yanirs: Hmm... This is a good point. Experiment duration is also a good example, as we want to warn for short and long durations, but not for anything in the middle.

src/lib/analyses.ts
yanirs: Yeah, not too fussed since the UI is separate. Thanks for the etymology reference! :slightly_smiling_face:

yanirs:
Thanks for the discussion @jessie-ross. Closing this as the code already moved on and we will be iterating on the warnings and their presentation.

PR #482, opened 2021-03-24
Update the bookmarklet alert to handle multiple non-cookie options

yanirs:
While manually testing D59227-code, I noticed that the alert message from the bookmarklet is wrong when the returned storage method is `sqooped_out_table`. This fixes it by flipping the condition.

## How has this been tested?
Manual testing as described in D59227-code.


aaronyan (approved):
LGTM! Nice catch!

PR #485, opened 2021-03-25
Extend inputs to getAggregateRecommendation

jessie-ross:
**This PR extends the inputs to getAggregateRecommendation in preparation for moving the stopping/warning logic from the ETL to Abacus**

Part of #419 

## How has this been tested?

### DIFF
--- a/src/lib/analyses.ts
+++ b/src/lib/analyses.ts
@@ -1,6 +1,13 @@
 import _ from 'lodash'
 
-import { Analysis, AnalysisStrategy, ExperimentFull, RecommendationWarning } from './schemas'
+import {
+  Analysis,
+  AnalysisStrategy,
+  ExperimentFull,
+  MetricAssignment,
+  MetricBare,
+  RecommendationWarning,
+} from './schemas'
 
 // I can't get stdlib to work as an import...:
 // eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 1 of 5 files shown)


src/lib/analyses.ts
```diff
@@ -49,10 +56,16 @@ export interface AggregateRecommendation {
...
-export function getAggregateRecommendation(
-  analyses: Analysis[],
-  defaultStrategy: AnalysisStrategy,
-): AggregateRecommendation {
+export function getAggregateRecommendation({
+  analyses,
+  defaultStrategy,
+}: {
+  experiment: ExperimentFull
+  metric: MetricBare
```
jessie-ross, line 64: Unfortunately we only receive MetricBare, we would need to change this to MetricFull or extend MetricBare as we need to have `higherIsBetter` available to make a recommendation.

jessie-ross: But these PRs will show the shape of this work and allow us to swap over the reliance on server data to client data at a later point.

src/lib/analyses.ts
```diff
@@ -49,10 +56,16 @@ export interface AggregateRecommendation {
...
+  metricAssignment: MetricAssignment
+  analyses: Analysis[]
+  defaultStrategy: AnalysisStrategy
```
yanirs, line 67: Should this be named `defaultStrategy`? It doesn't seem to be used as the default any more.

Also, all the parameters should be documented, and it's worth explaining how they affect the output of this function, as it's unclear from the documentation.

jessie-ross: > Should this be named defaultStrategy? It doesn't seem to be used as the default any more.

It is used in line 80 to pick a strategy to get the default recommendation off of.

yanirs: Yeah, I meant that it's inconsistent with other/previous uses of `defaultStrategy`. For example, `Experiments.getDefaultAnalysisStrategy()` always returns the same default strategy. 

In the context of this function, it's just the strategy that's used to choose the recommendation.

jessie-ross:
I don't think this is necessary for the moment.

PR #487, opened 2021-03-25
Add stopping recommendation to aggregate recommendation decision

jessie-ross:
**This PR extends aggregate recommendation decision to include the stopping recommendation:**

We have an `endExperiment` field as well the warnings which all relate to experiment length, `experiments/recommender.py#L40-L47` explains that the warnings take precedence. We then get three stopping recommendation states: `TooShort`, `GoodLength`, `TooLong`.

- `Inconclusive` and `TooShort` are the same, but `TooShort` is actionable and more understandable so we change the name and messaging:

Screen Shot 2021-03-25 at 10 04 08 pm

- `GoodLength` means we can just display the recommendation.

- `TooLong` is a soft cutoff (I am assuming, @yanirs do we want a hard cut off?) where we tell experimenters to stop the experiment:

Screen Shot 2021-03-25 at 11 18 00 pm

Yanir also mentioned in https://github.com/Automattic/abacus/issues/419#issuecomment-758314643 that email-like experiments shouldn't have `TooShort`, I will be addressing that in the next PR to keep this one smallish.

Part of #419 

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
+++ b/src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
@@ -12,7 +12,7 @@ const useStyles = makeStyles((theme: Theme) =>
       borderBottomStyle: 'dashed',
       borderBottomColor: theme.palette.grey[500],
     },
-    shouldStopAsterisk: {
+    shouldStop: {
       color: theme.palette.error.main,
     },
   }),
@@ -34,7 +34,7 @@ export default function AggregateRecommendationDisplay({
     <>
       <br />
       <Tooltip title='This experiment has run for too long.'>
-        <span className={clsx(classes.shouldStopAsterisk, classes.tooltipped)}>Stop Experiment</span>
+        <span className={clsx(classes.shouldStop, classes.tooltipped)}>Stop Experiment</span>
       </Tooltip>
     </>
   )
--- a/src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
+++ b/src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
@@ -1,4 +1,5 @@
 import { createStyles, makeStyles, Theme, Tooltip } from '@material-ui/core'
+import clsx from 'clsx'
 import React from 'react'
 
 import { AggregateRecommendation, AggregateRecommendationDecision } from 'src/lib/analyses'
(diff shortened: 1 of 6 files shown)


src/components/experiments/single-view/results/AggregateRecommendationDisplay.tsx
```diff
@@ -25,6 +29,16 @@ export default function AggregateRecommendationDisplay({
   experiment: ExperimentFull
 }): JSX.Element {
   const classes = useStyles()
+
+  const maybeShouldStopAsterisk = aggregateRecommendation.shouldStop && (
+    <>
+      <br />
+      <Tooltip title='This experiment has run for too long.'>
```
aaronyan, line 36: Maybe adjust to "This experiment has run for over 28 days. Consider stopping."

aaronyan (commented):
Thanks for researching and getting the stopping recommendation started @jessie-ross! A couple of thoughts:

I think we should add some context to the "Stop Experiment" in the top level panel. Consider putting it below the primary metric recommendation (in the same font size, unbolded), and have it say something like "And has a _long analysis_ warning" With the italics having the tooltip underline.

And I'm not the biggest fan of having "Stop Experiment" in the metrics panel, because it looks a bit redundant. I think keeping the only call to action in the top level panel will help keep things nice and clean. :)

I'll take a deeper look at the implementation tomorrow!

jessie-ross:
> I think we should add some context to the "Stop Experiment" in the top level panel. Consider putting it below the primary metric recommendation (in the same font size, unbolded), and have it say something like "And has a long analysis warning" With the italics having the tooltip underline.
> 
> And I'm not the biggest fan of having "Stop Experiment" in the metrics panel, because it looks a bit redundant. I think keeping the only call to action in the top level panel will help keep things nice and clean. :)

Yeah this is actually quite off the rails now that I am looking at it, your suggestions sound fine :-)

I have also been thinking about putting Inconclusive back in with a better tooltip ("More data needed") and display it for stopped experiments + email-like experiments instead of "More data". That is, it switches from "More data needed" to "Inconclusive" when stopped.

aaronyan:
> I have also been thinking about putting Inconclusive back in with a better tooltip ("More data needed") and display it for stopped experiments + email-like experiments instead of "More data". That is, it switches from "More data needed" to "Inconclusive" when stopped.

Yeah, this could be good! There was some confusion around the recommendation "inconclusive" recently (pbxNRc-D0-p2#comment-1892).

jessie-ross:
I'm splitting this up a bit differently, continues in https://github.com/Automattic/abacus/pull/488

PR #488, opened 2021-03-28
Integrate stopping info into AggregateRecommendation

jessie-ross:
_I've decided to split this up differently, with the presentation seperate to the logic changes as much as I can._

**This PR integrates stopping info into AggregateRecommendation:**
- Added "MoreDataNeeded" decision for running experiments that need more data.
  - In the ETL code `experiments/recommender.py#L40-L47` we find that warnings take precedence.
  - We want to integrate that into the top level decision, otherwise it will just get ignored as a warning or experimenters won't know what to do with it.
  - This does cause some "Deploy" recommendations to become "Inconclusive" but we will rebalance this by adding _lift_ as a column to the top level metric assignment line.
  - As a decision we switch it to "Inconclusive" once the experiment has stopped running and running it longer is no longer a choice.
- Added `shouldStop` as a soft stopping recommendation.
- Decided that we don't need to move the python code here yet, we have isolated all our recommendation in one pure function with a nice recommendation interface forming. It should be simple to do so later if we need and won't particularly add anything at the moment.
- Refactored the test code as it was getting quite large

## How has this been tested?

### DIFF
--- a/src/lib/analyses.ts
+++ b/src/lib/analyses.ts
@@ -7,6 +7,7 @@ import {
   MetricAssignment,
   MetricBare,
   RecommendationWarning,
+  Status,
 } from './schemas'
 
 // I can't get stdlib to work as an import...:
@@ -40,6 +41,7 @@ export const RecommendationWarningToHuman = {
 export enum AggregateRecommendationDecision {
   ManualAnalysisRequired = 'ManualAnalysisRequired',
   MissingAnalysis = 'MissingAnalysis',
+  MoreDataNeeded = 'MoreDataNeeded',
   Inconclusive = 'Inconclusive',
   DeployAnyVariation = 'DeployAnyVariation',
   DeployChosenVariation = 'DeployChosenVariation',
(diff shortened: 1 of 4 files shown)


src/lib/analyses.ts
```diff
@@ -84,21 +88,37 @@ export function getAggregateRecommendation({
...
+    recommendation.warnings.includes(RecommendationWarning.WideCi)
+  ) {
+    if (experiment.status === Status.Running) {
+      return {
+        decision: AggregateRecommendationDecision.MoreDataNeeded,
+      }
+    } else {
+      return {
+        decision: AggregateRecommendationDecision.Inconclusive,
+      }
```
yanirs, line 104: Let's discuss this behaviour change on P2. I don't think that combining the warnings and the decisions is where we want to go, as we kept them separate on purpose. It may just confuse and annoy experimenters, as the warnings are based on somewhat-arbitrary thresholds that don't always generalise well.

jessie-ross:
Closing in favour of a new direction: pbmo2S-Ii-p2#comment-1789

PR #489, opened 2021-03-28
Alter recommendation decision for experiments that are assigned upfront

jessie-ross:
**This PR:**
- Introduces the idea of experiments assigned up front.
- Ignores the time based warning for these experiments.

## How has this been tested?

### DIFF
--- a/src/lib/experiments.test.ts
+++ b/src/lib/experiments.test.ts
@@ -1,50 +1,59 @@
 import Fixtures from 'src/test-helpers/fixtures'
 
 import * as Experiments from './experiments'
-import { AnalysisStrategy } from './schemas'
+import { AnalysisStrategy, Platform } from './schemas'
 
-describe('lib/experiments.ts module', () => {
-  describe('getDeployedVariation', () => {
-    it('should return null when no deployed variation declared', () => {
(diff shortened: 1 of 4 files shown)


src/lib/experiments.test.ts
```diff
@@ -1,50 +1,59 @@
 import Fixtures from 'src/test-helpers/fixtures'
```
jessie-ross, line 1: Mostly whitespace changes here, not sure why they occurred.

yanirs:
@jessie-ross can you please provide more context and an explanation of what this PR does? The assumption that email experiments are all assigned at once isn't necessarily correct, as some email experiments are for drip campaigns.

jessie-ross:
> @jessie-ross can you please provide more context and an explanation of what this PR does? The assumption that email experiments are all assigned at once isn't necessarily correct, as some email experiments are for drip campaigns.

Ahh I thought that was the case, you mention in https://github.com/Automattic/abacus/issues/419#issuecomment-758314643:
> duration warnings are irrelevant for email experiments.

yanirs:
Yeah, I was wrong because I was only thinking of mass emails that typically run over a short period. It may be better to keep things simple and _not_ suppress the time-based warnings based on the platform, as experimenters are free to ignore warnings anyway.

PR #491, opened 2021-03-28
Fix recommendation shouldStop

jessie-ross:
This PR displays `shouldStop` under decisions except for `MissingAnalysis`.
- Makes `shouldStop` not optional.
- Moves Manual Analysis after Missing analysis.

## How has this been tested?

### DIFF
--- a/src/lib/analyses.ts
+++ b/src/lib/analyses.ts
@@ -51,7 +51,7 @@ export enum AggregateRecommendationDecision {
 export interface AggregateRecommendation {
   decision: AggregateRecommendationDecision
   chosenVariationId?: number
-  shouldStop?: boolean
+  shouldStop: boolean
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/lib/analyses.ts
```diff
@@ -76,16 +76,23 @@ export function getAggregateRecommendation({
     .filter((x) => x)
     .map((recommendation) => recommendation?.chosenVariationId)
   const recommendationConflict = [...new Set(recommendationChosenVariationIds)].length > 1
```
aaronyan, line 78: I like having this `const recommendationConflict` closer to where it's being used (the same way `const recommendation` is being used). Consider moving it down next to `if (recommendationConflict)`

aaronyan (approved):
LGTM! I like the call on making `shouldStop` not optional. Only a minor comment on moving some code around.

jessie-ross:
Closing due to change in direction: pbmo2S-Ii-p2#comment-1789

PR #497, opened 2021-04-08
Demo: Experiments AgGrid

jessie-ross:
**This PR adds a demo Experiments table using AgGrid:**
- Better status sorting
- Better default sort - status then start date
- Advanced column filtering
- Using links for the experiment name, can open them in a new tab

See p1617909144001200-slack-G01BQ5WMR8Q for a demo vid.

Ultimately it makes visiting the experiments page much more useful dashboard of running experiments, so experimenters are more likely to use it. Also the lack of open experiment in a new tab was quite aggravating.

Notably it is missing the search, I want to re-add the search to the right of the nav bar at a later point.

## How has this been tested?

It actually shouldn't require that much to get it to full coverage, I think just a snapshot and unit testing the custom sort

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10930,6 +10930,14 @@
       "resolved": "https://registry.npmjs.org/ag-grid-community/-/ag-grid-community-25.1.0.tgz",
       "integrity": "sha512-sKJp6SZG8J50OlOeECtrLFSp/iwdx1mcG0+BmIkHsXmWoyXv5ZaTJyVg1/hMk6s3JVTm1ipg6LepbTaigWEL2A=="
     },
+    "ag-grid-react": {
+      "version": "25.1.0",
+      "resolved": "https://registry.npmjs.org/ag-grid-react/-/ag-grid-react-25.1.0.tgz",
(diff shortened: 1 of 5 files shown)


jessie-ross:
AgGrid is the main library in the open source grid game:
- Extremely reputable
- Been around for a long time, battle tested
- Very well maintained, massive community
- Heaps of features
- Open-source level should cover everything we need but has option to add on "enterprise" features at a price

yanirs (approved):
Looks good!

I'm happy with adding it under `/experiments-ag` and graduating to `/experiments` when it's fully tested and ready.

withinboredom:
Do we want to add a banner to the top of `/experiments` saying something like: "try out the new experience [here] that allows better filtering. Leave feedback [on this issue]"?

jessie-ross:
I was thinking of showing it to some experimenters, how about a message in #a8cexperiments:

> We are working on a new experiments grid for Abacus, with better sorting, proper linking, and column filtering, it isn't fully tested yet but feel free to [try it out](https://experiments.a8c.com/experiments-ag).

To me it is miles away from the current grid, so I have a feeling I know what they would say, but this gives experimenters a chance to use it with a warning.

Issue #493, opened 2021-03-30
Discuss and fix issues with top-level warnings


yanirs:
See pbmo2S-Ii-p2 &ndash; there are a bunch of problems with the top-level warnings that we need to address to ensure that they are clear and not misleading.

jessie-ross:
I think this is closed now with the health report table.

yanirs:
Yep, thanks for addressing all the concerns. It's so much better now! :tada:

Issue #436, opened 2021-01-07
Positively Signal when an Experiment is ready to be stopped


jessie-ross:
Currently we only show warnings when an experiment needs more time, I think it would be helpful to also display when it is ready to finish.

jessie-ross:
Oops accidentally closed this one (#436) instead of #463

PR #498, opened 2021-04-13
Implement getExperimentDurationDays

jessie-ross:
**This PR adds a function to get an experiment duration in days or part thereof.**

Part of #436 #419 and #493

## How has this been tested?

### DIFF
--- a/src/lib/experiments.test.ts
+++ b/src/lib/experiments.test.ts
@@ -76,11 +76,11 @@ describe('lib/experiments.ts module', () => {
         Experiments.getExperimentRunHours(
           Fixtures.createExperimentFull({
             status: Status.Running,
-            startDatetime: new Date('2021-04-01T00:00:00Z'),
-            endDatetime: new Date('2021-04-05T00:00:00Z'),
+            startDatetime: new Date('2021-04-01T04:05:06Z'),
+            endDatetime: new Date('2021-04-05T03:02:01Z'),
           }),
         ),
-      ).toBe(84)
+      ).toBe(79)
       expect(
         Experiments.getExperimentRunHours(
           Fixtures.createExperimentFull({
--- a/src/lib/experiments.test.ts
+++ b/src/lib/experiments.test.ts
@@ -53,7 +53,7 @@ describe('lib/experiments.ts module', () => {
   describe('getExperimentDurationDays', () => {
     it('returns the correct number of days', () => {
       expect(
-        Experiments.getExperimentDurationDays(
+        Experiments.getExperimentRunHours(
           Fixtures.createExperimentFull({
             status: Status.Staging,
             startDatetime: new Date('2021-04-01T00:00:00Z'),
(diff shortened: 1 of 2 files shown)


src/lib/experiments.ts
```diff
@@ -56,3 +58,16 @@ export const AssignmentCacheStatusToHuman: Record<AssignmentCacheStatus, string>
...
+/**
+ * Returns an experiment's duration in days, with partial days as fractions.
+ */
+export function getExperimentDurationDays(experiment: ExperimentFull): number {
+  if (experiment.status === Status.Staging) {
+    return 0
+  }
+
+  const maybeEndDate = experiment.status === Status.Running ? new Date() : experiment.endDatetime
+  // We use differenceInHours rather than differenceInDays so we can get partial days, useful for experimenters and debugging.
```
yanirs, line 71: Any reason not to return the value in hours and convert it as needed? It may be better for debugging, as it avoids floating point precision issues like:

image

jessie-ross: I didn't see hours being useful for us, but changing it to hours does seem nice now that you mention it.

src/lib/experiments.ts
```diff
@@ -56,3 +58,16 @@ export const AssignmentCacheStatusToHuman: Record<AssignmentCacheStatus, string>
   [AssignmentCacheStatus.Missing]: '❌️ Missing: Expected for new, renamed, or disabled experiments',
   [AssignmentCacheStatus.Stale]: '❌️ Stale: Recent changes take up to ten minutes to propagate',
 }
+
+/**
+ * Returns an experiment's duration in days, with partial days as fractions.
+ */
+export function getExperimentDurationDays(experiment: ExperimentFull): number {
+  if (experiment.status === Status.Staging) {
```
yanirs, line 66: This is a bit confusing, since it makes it the experiment's duration to now (or days ran). Using it to get the planned duration of staging experiments seems intuitive. 

Maybe renaming the function to `getExperimentDurationToNow` or `getExperimentRunHours/Days`would help?

jessie-ross: Another problem in this area is using it for disabled experiments that haven't run, it shouldn't be an issue for my planned use as we only show inidactors for experiments that have run. It is a pity we don't have a `discarded` experiment status.

yanirs: Yeah, maybe we thinking about adding a `discarded` status when we add the `draft` status, or we can implement full deletion of experiments that never ran.

src/lib/experiments.test.ts
```diff
@@ -47,4 +49,56 @@ describe('lib/experiments.ts module', () => {
...
+      expect(
+        Experiments.getExperimentRunHours(
+          Fixtures.createExperimentFull({
+            status: Status.Staging,
+            startDatetime: new Date('2021-04-01T00:00:00Z'),
+            endDatetime: new Date('2021-04-05T00:00:00Z'),
+          }),
+        ),
+      ).toBe(0)
+      MockDate.set('2021-04-04T00:00:00Z')
```
yanirs, line 64: To make this a bit more realistic, it'd be good if not all test datetimes were in round hours.

yanirs (approved):
LGTM, but see inline comment on making the tests a bit more realistic before merging.

PR #499, opened 2021-04-13
Refactor health indicators

jessie-ross:
Best viewed commit by commit.

**This PR refactors health indicators:**
- Renames `ParticipantStats` to `HealthStats`
   I think it is simpler data flow wise if we aren't trying to get everything in one stats object.
- Renames `ExperimentHealthIndicators` to `ExperimentParticipantStatsHealthIndicator`
   This allows adding additional functions that have different dependencies.
- Turns `HealthIndication` from enum into an object
   Allows extensions to the indication, such as `severity`, `reason` and `recommendedAction`.
- Refactors the indicationBracket code, allowing it for use in other indicators.

Part of #436 #419 and #493

See also pbmo2S-Ii-p2#comment-1788

## How has this been tested?

### DIFF
--- a/src/lib/analyses.ts
+++ b/src/lib/analyses.ts
@@ -246,9 +246,16 @@ export enum HealthIndicationCode {
   ProbableIssue = 'ProbableIssue',
 }
 
+export enum HealthIndicationSeverity {
+  Clear = 'Clear',
+  Warning = 'Warning',
+  Error = 'Error',
+}
+
 interface HealthIndication {
   code: HealthIndicationCode
   reason: string
+  severity: HealthIndicationSeverity
 }
 
 export enum HealthIndicatorUnit {
@@ -328,18 +335,21 @@ export function getExperimentParticipantHealthIndicators(
           max: 0.001,
           indication: {
             code: HealthIndicationCode.ProbableIssue,
+            severity: HealthIndicationSeverity.Error,
           },
         },
         {
           max: 0.05,
           indication: {
             code: HealthIndicationCode.PossibleIssue,
+            severity: HealthIndicationSeverity.Warning,
           },
         },
         {
           max: 1,
           indication: {
             code: HealthIndicationCode.Nominal,
+            severity: HealthIndicationSeverity.Clear,
           },
         },
       ],
@@ -355,18 +365,21 @@ export function getExperimentParticipantHealthIndicators(
           max: 0.001,
           indication: {
             code: HealthIndicationCode.ProbableIssue,
+            severity: HealthIndicationSeverity.Error,
           },
         },
         {
           max: 0.05,
           indication: {
             code: HealthIndicationCode.PossibleIssue,
+            severity: HealthIndicationSeverity.Warning,
           },
         },
         {
           max: 1,
           indication: {
             code: HealthIndicationCode.Nominal,
+            severity: HealthIndicationSeverity.Clear,
           },
         },
       ],
@@ -382,18 +395,21 @@ export function getExperimentParticipantHealthIndicators(
           max: 0.001,
           indication: {
             code: HealthIndicationCode.ProbableIssue,
+            severity: HealthIndicationSeverity.Error,
           },
         },
         {
           max: 0.05,
           indication: {
             code: HealthIndicationCode.PossibleIssue,
+            severity: HealthIndicationSeverity.Warning,
           },
         },
         {
           max: 1,
           indication: {
             code: HealthIndicationCode.Nominal,
+            severity: HealthIndicationSeverity.Clear,
           },
         },
       ],
@@ -408,18 +424,21 @@ export function getExperimentParticipantHealthIndicators(
           max: 0.01,
           indication: {
             code: HealthIndicationCode.Nominal,
+            severity: HealthIndicationSeverity.Clear,
           },
         },
         {
           max: 0.05,
           indication: {
             code: HealthIndicationCode.PossibleIssue,
+            severity: HealthIndicationSeverity.Warning,
           },
         },
         {
           max: 1,
           indication: {
             code: HealthIndicationCode.ProbableIssue,
+            severity: HealthIndicationSeverity.Error,
           },
         },
       ],
@@ -434,18 +453,21 @@ export function getExperimentParticipantHealthIndicators(
           max: 0.075,
           indication: {
             code: HealthIndicationCode.Nominal,
+            severity: HealthIndicationSeverity.Clear,
           },
         },
         {
           max: 0.3,
           indication: {
             code: HealthIndicationCode.PossibleIssue,
+            severity: HealthIndicationSeverity.Warning,
           },
         },
         {
           max: 1,
           indication: {
             code: HealthIndicationCode.ProbableIssue,
+            severity: HealthIndicationSeverity.Error,
           },
         },
       ],
--- a/src/lib/analyses.ts
+++ b/src/lib/analyses.ts
@@ -240,16 +240,21 @@ export function getExperimentParticipantStats(
   }
 }
 
-export enum HealthIndication {
+export enum HealthIndicationCode {
   Nominal = 'Nominal',
   PossibleIssue = 'PossibleIssue',
   ProbableIssue = 'ProbableIssue',
+}
+
+interface HealthIndication {
+  code: HealthIndicationCode,
 }
 
 export enum HealthIndicatorUnit {
   Pvalue = 'P-Value',
   Ratio = 'Ratio',
 }
+
 
 /**
  * Indicators are the important stats that give us clear direction on how an experiment is going.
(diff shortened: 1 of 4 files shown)


src/lib/analyses.ts
```diff
@@ -240,12 +240,24 @@ export function getExperimentHealthStats(
...
 
-export enum HealthIndication {
+export enum HealthIndicationCode {
   Nominal = 'Nominal',
   PossibleIssue = 'PossibleIssue',
   ProbableIssue = 'ProbableIssue',
 }
 
+export enum HealthIndicationSeverity {
+  Clear = 'Clear',
```
yanirs, line 250: Maybe call this `Info` to match common logging level names?

src/lib/analyses.ts
```diff
@@ -262,12 +274,35 @@ export interface HealthIndicator {
   indication: HealthIndication
```
yanirs, line 274: How about we flatten this by removing the `HealthIndication` and the indication term? All this nesting and terminology makes the code pretty hard to follow.

```suggestion
  code: HealthIndicatorCode
  reason: string
  severity: HealthIndicatorSeverity
```

src/lib/analyses.ts
```diff
@@ -240,12 +240,24 @@ export function getExperimentHealthStats(
...
 }
 
-export enum HealthIndication {
+export enum HealthIndicationCode {
   Nominal = 'Nominal',
   PossibleIssue = 'PossibleIssue',
   ProbableIssue = 'ProbableIssue',
 }
 
+export enum HealthIndicationSeverity {
```
yanirs, line 249: Having both severity and code feels redundant. What scenarios do you have in mind where this would be useful?

src/lib/analyses.ts
yanirs: Also, is there a reason to keep `link` optional?

src/lib/analyses.ts
jessie-ross: See the example in pbmo2S-Ii-p2#comment-1787

Also flow (24 hour participant diff while running or average):
- Very Low (error)
- Low (warning)
- Ok (nominal)
- High (warning)
- Very High (error)

And I am thinking of possible adding conflict indication codes:

MITT vs ITT
- No conflict (nominal)
- Some inconclusive (nominal)
- Conflicting recommendation (warning)

MITT vs PPNaive
- No conflict (nominal)
- Some inconclusive (nominal)
- Conflicting recommendation (warning)

(I am also thinking of adding the same expandable health report table within each metricAssignment for reviewers, it should basically come free.)

src/lib/analyses.ts
jessie-ross: Yeah I was thinking about that but I feel this is cleaner, these all belong to the indication so we should at least be prefixing them:

`
  indicationCode: HealthIndicatorCode
  indicationReason: string
  indicationSeverity: HealthIndicatorSeverity
`

I think it is cleaner this way and gives us a better pattern with the bracketing code.

src/lib/analyses.ts
jessie-ross: Well this isn't a log and I don't think info fits here as we are reporting on the severity of a value, we want to explicitly say it is of an expected range, I would prefer `nominal` here really but went with `clear` as a compromise.

src/lib/analyses.ts
yanirs: I don't see why these examples need both a code an severity. Can you give an example where the mapping isn't one to one?

src/lib/analyses.ts
yanirs: The logging severity levels are [well established](https://en.wikipedia.org/wiki/Syslog#Severity_level) and likely to be known by anyone reading this code. They fit very well here, as we're indicating the severity of a message. I don't see why "clear" is a better severity level descriptor than "info" in this context, but perhaps the best solution is to drop this enum, as noted above.

src/lib/analyses.ts
yanirs: Given the large number of interfaces and redundant terminology introduced here, I very strongly feel that the extra nesting and the subtle distinction between indicator and indication over-complicate things. It's especially confusing with `value` being a part of the indicator, but `severity` and `code` being a part of the indication. With a flat structure and fewer interfaces, there will be less to keep in mind, and I think that it can actually improve the bracketing code.

src/lib/analyses.ts
jessie-ross: I might get a third opinion on this @withinboredom

I would prefer `nominal` here but went with clear [after looking up what other systems use for indicators](https://www.google.com/search?q=severity+level&tbm=isch&ved=2ahUKEwj886qXyv3vAhVgnksFHW_KAvEQ2-cCegQIABAA&oq=severity+level&gs_lcp=CgNpbWcQAzICCAAyAggAMgIIADICCAAyAggAMgIIADICCAAyAggAMgIIAFC1MFjjNWCgN2gAcAB4AIAB0gGIAZADkgEFMC4xLjGYAQCgAQGqAQtnd3Mtd2l6LWltZ8ABAQ&sclient=img&ei=_sp2YPz8AeC8rtoP75SLiA8&bih=986&biw=1792&safe=active&hl=en). How about `satisfactory` or `ok`?

> likely to be known by anyone reading this code

I don't see anyone being tripped up by this, particularly if we go `nominal`, `satisfactory` or `ok`.

> They fit very well here, as we're indicating the severity of a message.

An indicator is different to a log as indicators are persistent and in-place, they are not messages. I think seeing an `info` here would be confusing actually.

src/lib/analyses.ts
jessie-ross: - The duration in the P2 comment, e.g. `long` and `short` are both warnings.
- Flow mentioned above is similar for `low` and `high`.
- The conflict indication codes examples above both have two nominal severities

src/lib/analyses.ts
jessie-ross: I strongly feel that the indication is a seperate derivation of the value and the indicator type, particularly the `code`, `reason`, `severity`, and the possible future `recommendation` should all be grouped together. 

To me this is a case of "as simple as possible but no simpler".

> so we should at least be prefixing them

What did you think about this? I'm also happy to change the names around.

src/lib/analyses.ts
withinboredom: The code and severity separation makes the most sense and is the most flexible. My 2¢

src/lib/analyses.ts
withinboredom: Yeah. This isn't a log :).

The ones I've seen today said ok, a green light, or ready. Personally, I like ok. But any of those work and I'd actually recommend picking the one that looks best. For example if "ok" looks too short, go with something longer or if we want the "ok" shortness compared to others to also be an indication, then that's fine too.

withinboredom (approved):

src/lib/analyses.ts
yanirs: Yeah, OK is better. Can spell it out as `Okay` if we want a longer word.

src/lib/analyses.ts
yanirs: I still don't see it, but maybe I'm missing something, so I'll rephrase: Is there a case where these pairs don't go together?

* Severity: `Okay` + Code: `Nominal`
* Severity: `Warning` + Code: `PossibleIssue`
* Severity: `Error` + Code: `ProbableIssue`

src/lib/analyses.ts
yanirs: Prefixing them would be better, though I still don't think that the concept of an indication is worth keeping.

src/lib/analyses.ts
jessie-ross: It's is because the code isn't just those three which are for probabilistic indicators, for example for experiment duration:
- Severity: Error + Code: Very Short
- Severity: Warning + Code: Short
- Severity: Okay + Code: Nominal
- Severity: Warning + Code: Long
- Severity: Error + Code: Very Long

yanirs: Ah, I see. I think we need to rename the code and its items then, or drop the code.

jessie-ross: I see the code as very important, the point of these indications is to abstract the figure behind them - so the Experimenters don't need to know what a p-value or ratio is and get an easily digestible message like "Short" experiment duration or "High" crossovers or "PossibleIssue" with the Assignment distribution.

yanirs: Ah ok, so you mean that Code would have something like `VeryShort` as a possible value in addition to the current values? This makes sense to me. :slightly_smiling_face:

yanirs: Hmm... Not for this PR, but should we change the codes used for ratios? Unlike the p-value thresholds, we don't really know that certain ratios are indicative of probable or possible issues.

src/lib/analyses.ts
jessie-ross: After trying it out I am not convinced the flat structure is any better and it seems to me like the only difference is we are using `indicationCode` instead of `indication.code`, I am going to go forward with the nested indication - it is a relatively local decision and we can always change it later.

PR #502, opened 2021-04-15
Health Report

jessie-ross:
**This PR adds the Health Report to experiment results:**

Screen Shot 2021-04-16 at 4 04 50 am

- No Information hidden on hover
- Clearer icon semantics - emojis go alongside the color and the indication code.
- Sentence case
- Links open in a new tab
- Plenty of room to expand.

Part of #436 #419 and #493

See also pbmo2S-Ii-p2#comment-1788

**Future directions:**
- Putting Health report under a show-hide
- Health report for each metric assignment

**Todo:**
- [x] Testing
- [x] Remove the top level indicators
- [x] Add overall report severity with link to the top level display

## How has this been tested?


yanirs:
Woohoo! This is definitely going in the right direction. :tada:

A few comments based on the screenshot:

* We should round the values when displaying them. Rounding to four decimal places would be good enough.
* I think it'd look better to have the icons to the left of the code.
* Looking at [the dictionary](https://www.merriam-webster.com/dictionary/indication), I still think that the word `Indication` isn't right here. `Health Level` or `Status` would be better.
* The indicator names can be improved along the lines of my suggestion on pbmo2S-Ii-p2, but this can come later.

jessie-ross:
> Looking at the dictionary, I still think that the word Indication isn't right here. Health Level or Status would be better.

To me it seems right, it is the first definition "1a: something that serves to indicate", although it is a brief definition. Thankfully https://english.stackexchange.com/a/73099  clears up this exact definition.

> To see how well an economy is performing, we look at economic indicators, for example. That apart, an indication is indeed what an indicator provides. Indicator is the means, indication the result.

E.g. the unemployment rate is 5.6% (the indicator), this indicates that the economy is performing ok (the indication).

The basic idea is that an indicator gives an indication, in this case the indicator itself is a measure.

> Health Level

I don't think this fits, we will have e.g. "High" and "Low" as indications. "High" health level would be confusing.

> Status

We are already using Status elsewhere and we don't want to overload it, and we would have the same confusion as before, "Very High" status might seem good.

---

Ultimately I think the important part is the indication code and the highlighting/emoji, having a header that doesn't completely fit isn't going to break anything, especially when compared to ones that can cause actual confusion.

jessie-ross:
Some fresh screenshots:


Screen Shot 2021-04-20 at 4 16 16 am

Screen Shot 2021-04-20 at 4 19 04 am
 
Screen Shot 2021-04-20 at 4 17 23 am

Screen Shot 2021-04-20 at 4 25 57 am

aaronyan:
LGTM so far! I'm a fan of the simple top-level summary of the health metrics and the new way to display the health indicators. 

We should consider reducing the amount of color shown for the health indicator, so it's not so alarming. It's good to bring attention to potential issues, but we don't want to create anxiety for the experimenter. Consider using a linear bar with some color to indicate health issues instead of coloring the entire card. Something like this:

subtle-health-indicator

And just to cast my vote, I like "indicator" the most atm. I'll wait until the tests pass before diving deeper in the review.

yanirs:
> Ultimately I think the important part is the indication code and the highlighting/emoji, having a header that doesn't completely fit isn't going to break anything, especially when compared to ones that can cause actual confusion.

The confusing use of indication isn't limited to the header, though. The confusion is also in the code. I agree that it's not the most important part, but names are important. As per previous PRs, I believe that removing the "indication" terms can also improve the code.

Since we are borrowing from medicine with _Health Report_, it's worth looking at [the medical definition of the word](https://en.wikipedia.org/wiki/Indication_(medicine)):

> In medicine, an indication is a valid reason to use a certain test, medication, procedure, or surgery. There can be multiple indications to use a procedure or medication. An indication can commonly be confused with the term diagnosis. A diagnosis is the assessment that a particular [medical] condition is present while an indication is a reason for use. The opposite of an indication is a contraindication, a reason to withhold a certain medical treatment because the risks of treatment clearly outweigh the benefits. 

It seems to me like "diagnosis" is what's really meant here, with the "indication" being the measured value. If we believe Wikipedia, the confusion between a diagnosis and an indication is common, though. :slightly_smiling_face: 

As further evidence that it's the wrong word, it's worth noting that none of the common synonyms of _indication_ make sense here (e.g., see https://www.thesaurus.com/browse/indication and https://www.merriam-webster.com/thesaurus/indication).

> And just to cast my vote, I like "indicator" the most atm.

@aaronyan where do you see _indicator_? :slightly_smiling_face: What do you think of _diagnosis_ instead of _indication_ as the table header? I'll just let it go if you also think that _indication_ is better.

Other than that, the new screenshots look great, though I agree with Aaron on the reduced use of colour.

aaronyan:
> @aaronyan where do you see indicator?

Ah sorry for the confusion. I was referring to voting for "indicator" in the code with `experimentHealthIndicators` and `HealthIndicatorTable`. I now see "indication" in the table header, and I think it's decent (not my favorite, but passable imo). 

> It seems to me like "diagnosis" is what's really meant here, with the "indication" being the measured value. If we believe Wikipedia, the confusion between a diagnosis and an indication is common, though. 🙂

I think "diagnosis" sounds great for the table header!  I'll formally change my vote for "diagnosis/ diagnoses" in the header and code. It's a lot to replace though, and the changes may extend past this PR. It may be worth adjusting that naming for a future PR.

jessie-ross:
Thanks for the input @aaronyan, I will stick the terminology as-is for this PR.

For the future, as far as diagnosis is concerned, I don't see it as a fit - a "diagnosis" is a suspected cause of an issue (what is wrong), while this is more like a symptom (whether something is wrong). However there is the term "Diagnostics", "that which helps you determine a diagnosis", which has prior art in the land of computer systems and might be a better fit for "indicator" and would be paired with "Result" for the "indication".

jessie-ross:
The new look:

Screen Shot 2021-04-21 at 3 47 06 pm

Screen Shot 2021-04-21 at 4 11 37 pm

aaronyan (approved):
LGTM! Thank you @jessie-ross for fielding and discussing all of the iterations for the health report. This looks great!

PR #503, opened 2021-04-15
Improve health indicators pt 1

jessie-ross:
**This PR improves Health Indicators:**
- Hide exposure event distribution when unnecessary
- Gracefully handle indication errors, fixing a recent bug where a value was `NaN` and didn't match any brackets and caused a crash
- Adds proportional indications for ratios

**Next PR:**
- Add seperate MittNoCrossoverNoSpammer distribution prob

## How has this been tested?


src/lib/analyses.ts
```diff
@@ -242,8 +242,17 @@ export function getExperimentParticipantStats(
 
 export enum HealthIndicationCode {
   Nominal = 'Nominal',
+  IndicationError = 'IndicationError',
```
yanirs, line 245: I think that `ValueError` would be a better name (or `InvalidValue`/`UnexpectedValue`). Repeating the word `Indication` when it's already in the enum name looks a bit odd.

src/lib/analyses.ts
```diff
@@ -323,38 +340,41 @@ export function getExperimentParticipantHealthIndicators(
...
+          severity: HealthIndicationSeverity.Ok,
         },
-      ],
-    },
-    {
+      },
+    ],
+  })
+
+  experimentParticipantStats.ratios.overall.exposedToAssigned !== 0 &&
```
yanirs, line 376: This would be more readable as an `if`.

src/lib/analyses.ts
```diff
@@ -383,96 +403,97 @@ export function getExperimentParticipantHealthIndicators(
...
-      indicationBrackets: [
-        {
-          max: 0.001,
-          indication: {
-            code: HealthIndicationCode.ProbableIssue,
-            severity: HealthIndicationSeverity.Error,
-          },
+    })
+
+  indicatorDefinitions.push({
```
yanirs, line 408: Any reason not to call `push()` once with multiple arguments here?

yanirs (commented):
Since we're touching the indicator definitions, this PR may be a good opportunity for also improve their names as noted in pbmo2S-Ii-p2.

Can you please also extend the tests to reproduce the crash so we have it covered in the future?

src/lib/analyses.ts
jessie-ross: This seems like a nice pattern?

yanirs: Seems like unnecessary repetition to me. :slightly_smiling_face:

src/components/experiments/single-view/results/HealthIndicators.tsx
```diff
@@ -19,6 +19,11 @@ const indicationCodeToMessage: Record<HealthIndicationCode, React.ReactNode> = {
       🆗
     </span>
   ),
+  [HealthIndicationCode.ValueError]: (
```
jessie-ross, line 22: Yeah I am trying to avoid stacked PRs and I have ended up with parallel PRs, this is of course temporary and will be removed once the table is added.

jessie-ross:
Looks like Vercel is breaking due to stdlib:


> 00:36:16.801 | npm ERR! Could not install from "node_modules/@stdlib/node_modules/@stdlib/stdlib/lib/node_modules/@stdlib/array" as it does not contain a package.json file.

jessie-ross:
I am a bit worried about the maturity of Stdlib - and the security of it since it is so much un audited code, it doesn't pass my sniff test.
Since it can be replaced but one very well unit tested function that's already been written, why not replace it and save us the hassle?

yanirs:
> I am a bit worried about the maturity of Stdlib - and the security of it since it is so much unaudited code, it doesn't pass my sniff test.
> Since it can be replaced but one very well unit tested function that's already been written, why not replace it and save us the hassle?

Yeah, I agree. It's unfortunate since it looked good on paper, but having to work hard to use a third-party library defeats the purpose of using a library. Can you please open a separate PR to remove it?

By the way, do you know why we're getting failures on Vercel but not locally or on CircleCI?

jessie-ross:
> By the way, do you know why we're getting failures on Vercel but not locally or on CircleCI?

I think it is to do with caching, if I create a new repo it the stdlib issue occurs.

yanirs (approved):

PR #504, opened 2021-04-19
Improve health indicators pt 2

jessie-ross:
**This PR:**
- Adds seperate MittNoCrossoverNoSpammer distribution prob
- Changes the names as per the P2 pbmo2S-Ii-p2 

## How has this been tested?

### DIFF
--- a/src/lib/analyses.test.ts
+++ b/src/lib/analyses.test.ts
@@ -353,28 +353,33 @@ describe('getParticipantCounts', () => {
           }),
         },
       ),
-    ).toEqual({
-      total: {
-        assigned: 100,
-        assignedCrossovers: 10,
-        assignedSpammers: 15,
-        exposed: 40,
-      },
-      byVariationId: {
-        '1': {
-          assigned: 40,
-          assignedCrossovers: 5,
-          assignedSpammers: 0,
-          exposed: 15,
+    ).toEqual(
+       {
+        "byVariationId":  {
+          "1":  {
+            "assigned": 40,
+            "assignedCrossovers": 5,
+            "assignedNoSpammersNoCrossovers": 25,
+            "assignedSpammers": 0,
+            "exposed": 15,
+          },
+          "2":  {
+            "assigned": 70,
+            "assignedCrossovers": 15,
+            "assignedNoSpammersNoCrossovers": 35,
+            "assignedSpammers": 25,
+            "exposed": 25,
+          },
         },
-        '2': {
-          assigned: 70,
-          assignedCrossovers: 15,
-          assignedSpammers: 25,
-          exposed: 25,
+        "total":  {
+          "assigned": 100,
+          "assignedCrossovers": 10,
+          "assignedNoSpammersNoCrossovers": 60,
+          "assignedSpammers": 15,
+          "exposed": 40,
         },
-      },
-    })
+      }
+    )
   })
 
   it('should work correctly without any analyses', () => {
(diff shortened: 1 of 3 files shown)


src/lib/analyses.ts
```diff
@@ -101,6 +101,7 @@ interface CountsSet {
   assigned: number
   assignedCrossovers: number
   assignedSpammers: number
+  assignedNoSpammersNoCrossovers: number
```
withinboredom, line 104: nit: a better name might be `assignedWithoutSpammersOrCrossovers`

src/lib/analyses.ts
```diff
@@ -330,6 +345,10 @@ export function getExperimentParticipantHealthIndicators(
         acc.assignedSpammersDistributionMatchingAllocated,
         cur.assignedSpammersDistributionMatchingAllocated,
       ),
+      assignedNoSpammersNoCrossoversDistributionMatchingAllocated: Math.min(
+        acc.assignedNoSpammersNoCrossoversDistributionMatchingAllocated,
+        cur.assignedNoSpammersNoCrossoversDistributionMatchingAllocated,
+      ),
       exposedDistributionMatchingAllocated: Math.min(
         acc.exposedDistributionMatchingAllocated,
         cur.exposedDistributionMatchingAllocated,
```
withinboredom, line 354: This seems ripe for a wee bit of refactoring to keep it DRY.

withinboredom (approved):
I think some of this could benefit from some dynamic programming, but otherwise, it looks good and isn't a blocker.

src/lib/analyses.ts
```diff
@@ -343,44 +362,43 @@ export function getExperimentParticipantHealthIndicators(
...
-  })
-
-  if (experimentParticipantStats.ratios.overall.exposedToAssigned) {
-    indicatorDefinitions.push({
-      name: 'Exposure event distribution matching allocated',
-      value: minVariationProbabilities.exposedDistributionMatchingAllocated,
+      ],
+    },
+    {
+      name: 'Assignment without crossovers and spammers distribution',
```
yanirs, line 397: This is a bit awkward. Maybe `Assignment distribution without crossovers and spammers`?

src/lib/analyses.ts
```diff
@@ -404,16 +422,16 @@ export function getExperimentParticipantHealthIndicators(
...
+    },
+  )
 
-  indicatorDefinitions.push(
-    {
-      name: 'Spammer distribution matching allocated',
-      value: minVariationProbabilities.assignedSpammersDistributionMatchingAllocated,
+  if (experimentParticipantStats.ratios.overall.exposedToAssigned) {
+    indicatorDefinitions.push({
+      name: 'Exposure distribution',
```
yanirs, line 430: In line with the renaming above, how about `Assignment distribution of exposed participants`?

src/lib/analyses.ts
yanirs: This name is consistent the name of the analysis strategy (`MittNoSpammersNoCrossovers = 'mitt_no_spammers_no_crossovers'`).

src/lib/analyses.test.ts
```diff
@@ -562,57 +573,57 @@ describe('getExperimentHealthIndicators', () => {
         indication: {
           code: 'ProbableIssue',
           reason: '−∞ < x ≤ 0.001',
-          severity: Analyses.HealthIndicationSeverity.Error,
+          severity: 'Error',
```
yanirs, line 576: Why were the constants removed?

src/lib/analyses.ts
withinboredom: Yeah, I know. I'd love to see it renamed to something better. There's no reason the name has to stay the same.

src/lib/analyses.test.ts
withinboredom: I wonder if webpack interns strings... or the javascript runtime. :thinking:

src/lib/analyses.ts
yanirs: Yeah, it may not be the best name, but in my opinion it's worth keeping things consistent to make the relationship between the two clear. It's not critical, though. :slightly_smiling_face:

src/lib/analyses.test.ts
jessie-ross: Oh, I have been using toMatchInlineSnapshot and converting back to toEqual so I don't have to manually do all these test changes, and I forgot to change these back, in #505 I am switching just to the snapshot method.

src/lib/analyses.ts
jessie-ross: Maybe if it gets larger than this, I think it is fine as it is so localised

PR #505, opened 2021-04-20
Remove Stdlib

jessie-ross:
This PR removes stdlib.
- See https://github.com/Automattic/abacus/pull/503#issuecomment-823006014
- Replaces it with the previous implementation: https://github.com/Automattic/abacus/pull/473/commits/8317c9d43f044dc30bd7c3e97c79105888786dfc

## How has this been tested?


src/utils/math.ts
```diff
@@ -0,0 +1,31 @@
+import { abs, erf } from 'mathjs'
+
+/**
+ * Assuming a distribution of X ~ Binomial(n, p), returns the probability of the number of trials being equal to x or more extreme than x.
```
yanirs, line 4: Can you please rephrase this in terms of the parameters passed into the function?

src/utils/math.ts
```diff
@@ -0,0 +1,31 @@
...
+  // By the CLT, B ~ Binomial(n, p) is approximated well enough by X ~ N(mean, variance) for n > 30
+  // See also: https://en.wikipedia.org/wiki/Central_limit_theorem
+  //           https://en.wikipedia.org/wiki/Binomial_distribution#Normal_approximation
+  // (We don't care about the accuracy for n <= 30 so we let them be.)
+  // And if Y ~ N(0, 1/2) then Y = (X - mean)/(2 * variance)^(1/2)
+  // See also: https://en.wikipedia.org/wiki/Normal_distribution#Symmetries_and_derivatives
+  const y = (successfulTrials - mean) / Math.sqrt(2 * variance)
+  // Since erf(abs(y)) gives the probabilty of [-y, y] on Y, and pValue is the union of [-inf, -y] and [y, inf]:
+  // See also: https://en.wikipedia.org/wiki/Error_function
+  return 1 - erf(abs(y))
```
yanirs, line 39: I don't understand where `Y ~ N(0, 1/2)` comes from. Isn't this just [the complement of the Normal CDF](https://en.wikipedia.org/wiki/Normal_distribution#Cumulative_distribution_function) multiplied by two here?

yanirs: The last commit moved the comment, by the question still stands. :slightly_smiling_face:

src/lib/analyses.test.ts
```diff
@@ -461,49 +465,51 @@ describe('getExperimentParticipantStats', () => {
...
+      Object {
+        "probabilities": Object {
+          "byVariationId": Object {
+            "1": Object {
+              "assignedDistributionMatchingAllocated": 0.000011583130623216142,
+              "assignedSpammersDistributionMatchingAllocated": 1.970346108493004e-11,
+              "exposedDistributionMatchingAllocated": 0.11384629800665802,
+            },
+            "2": Object {
+              "assignedDistributionMatchingAllocated": 0.3804551252503884,
```
yanirs, line 478: These differences aren't negligible. Are you confident that they're not indicative of an implementation issue?

src/utils/math.ts
jessie-ross: `Y` is introduced for the purpose of the `erf` which I went with as it was easy to find a good quality lib for and it has a fast and simple implementation.

I think this is fine, but in the last two months [this normal distribution lib has been created](https://www.npmjs.com/package/gaussian), also [using erf in basically the same way](https://github.com/errcw/gaussian/blob/67296bb5fd476a08a3634aefc0931db2f63c7925/lib/gaussian.js#L56), and I could switch it out. Although it is quite young.

src/lib/analyses.test.ts
jessie-ross: This is for a relatively small `n` which and even then is approximate enough for our usage.

I've added some more tests and it looks right.

src/utils/math.test.ts
```diff
@@ -0,0 +1,85 @@
...
+        Math.abs(binomialProbValue({ successfulTrials: 15, totalTrials: 30, probabilityOfSuccess: 0.5 }) - 1),
+      ).toBeLessThan(aSmallValue)
+      expect(
+        Math.abs(
+          binomialProbValue({ successfulTrials: 60, totalTrials: 1000, probabilityOfSuccess: 0.000008 }) - 0.000002,
+        ),
+      ).toBeLessThan(aSmallValue)
+      expect(
+        Math.abs(
+          binomialProbValue({ successfulTrials: 70, totalTrials: 1000, probabilityOfSuccess: 0.001069 }) - 0.000002,
```
jessie-ross, line 33: Oops actually I messed these up

src/utils/math.ts
yanirs: I still don't understand this bit: `if Y ~ N(0, 1/2) then Y = (X - mean)/(2 * variance)^(1/2)` &ndash; how does defining Y in this way create a relationship with X?

How about something like this explanation?

> From symmetry of the normal distribution, P(not -x < X < x) = P(X > x or X < -x) = 2 * P(X > x)
> From https://en.wikipedia.org/wiki/Normal_distribution#Cumulative_distribution_function, 2 * P(X > x) = 2 - 2 * CDF(x) = 2 - 2 * 0.5 * (1 + erf(x - mean / (sqrt(2 * variance)))) = 1 - erf(x - mean / (sqrt(2 * variance)))

src/utils/math.test.ts
```diff
@@ -0,0 +1,85 @@
+import { binomialProbValue } from './math'
+
+describe('utils/math.ts module', () => {
+  describe('binomialProbValue', () => {
+    it('is approximately correct', () => {
```
yanirs, line 5: Can't we use [`toBeCloseTo`](https://jestjs.io/docs/expect#tobeclosetonumber-numdigits) in this test?

src/lib/analyses.test.ts
jessie-ross: scratch that

jessie-ross: Using the gaussian npm package I get basically the same approximation as this implementation, assuming there is nothing wrong with the mean, variance and usage of everything it looks like this is just the difference between using the normal approximation vs without approximation.

**Using gaussian package:**

Screen Shot 2021-04-21 at 5 10 33 am

**Using current implementation:**

Screen Shot 2021-04-21 at 5 12 19 am

The tests aren't great as the "toBeCloseTo" type check can't scale according to the value - it checks fixed digits while the smaller the pValue the smaller the error will be, we should in fact be checking that the ratio of approximation to actual is close.

Using higher values of `n` in the test makes them more accurate, but I have still had to use a low `digitsPrecision` (but which is plenty for us).

Since `gaussian` is a bit too new, without typescript, and I can't find any other sane CDF lib, I think it is good to go with the current implementation.

src/utils/math.ts
jessie-ross: > I still don't understand this bit: if Y ~ N(0, 1/2) then Y = (X - mean)/(2 * variance)^(1/2) – how does defining Y in this way create a relationship with X?

The transformation comes from the standard way to transform normal distributions mentioned in the linked wiki section: `Z = (X - mu)/sigma` - this makes it easy to relate any two normal distributions.

Maybe a bit too mathy (although to me it seems more intuitive than a formula), so I will include your formula.

> Where does the abs come from?

From the relation mentioned in the Erf wikipedia article:
> In statistics, for **non-negative values of x**, the error function has the following interpretation: for a random variable Y that is normally distributed with mean 0 and variance 1/2, erf x is the probability that Y falls in the range **[−x, x]**.

Because the probability range flips around if x < 0 it simplifies things to use the `abs`.

yanirs (approved):
:ship:

PR #509, opened 2021-04-28
Reword the existing users only switch

jessie-ross:
**This PR renames "existing user" terminology in the UI:**
- "New users only" becomes "New logged-in users only".
- Emphasise that all users includes anonymous participants.

Screen Shot 2021-04-28 at 12 16 05 pm
Screen Shot 2021-04-28 at 12 15 18 pm
Screen Shot 2021-04-28 at 12 15 06 pm

Resolves 621-gh-Automattic/experimentation-platform

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/overview/AudiencePanel.tsx
+++ b/src/components/experiments/single-view/overview/AudiencePanel.tsx
@@ -146,7 +146,7 @@ function AudiencePanel({
       label: 'User Type',
       value: (
         <span className={classes.monospace}>
-          {experiment.existingUsersAllowed ? 'All users (new + existing)' : 'New users only'}
+          {experiment.existingUsersAllowed ? 'All users (new + existing + anonymous)' : 'New logged-in users only'}
         </span>
       ),
     },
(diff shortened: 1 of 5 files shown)


src/components/experiments/single-view/overview/AudiencePanel.tsx
```diff
@@ -146,7 +146,7 @@ function AudiencePanel({
       label: 'User Type',
```
yanirs, line 146: While you're here, I noticed that this is `User types` in the creation form and `User Type` here. We can change this to `User Types` for now and worry about sentence-case/title-case consistency later. :slightly_smiling_face:

yanirs (approved):
LGTM

src/components/experiments/single-view/overview/AudiencePanel.tsx
jessie-ross: It makes sense to me that having a list of types to choose from is plural while displaying the current type is singular?

yanirs: By that logic we should also have `Platforms` rather than `Platform` in the creation form, but since an experiment can only have a single platform it makes sense to stick to the singular form. Therefore, changing the creation form makes more sense to me now. :slightly_smiling_face:

jessie-ross: Actually I think it should be singular for both

yanirs: :+1: Agreed

PR #510, opened 2021-04-29
Bookmarklet: Set assignment in LocalStorage

jessie-ross:
**This PR sets the assignment in LocalStorage in the bookmarklet.**

Also improves the bookmarklet alert:

Screen Shot 2021-04-29 at 1 17 17 pm

Resolves 614-gh-Automattic/experimentation-platform

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/overview/VariationsTable.tsx
+++ b/src/components/experiments/single-view/overview/VariationsTable.tsx
@@ -89,9 +89,9 @@ function assignmentHref(variationName: string, experimentName: string) {
             )
           );
 
-          alert('ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\nExperiment: ${experimentName}\\nVariation: ${variationName}\\nTTL: Expires in ' + duration + ' hours\\n\\nClient-side: Applies to the current domain (LocalStorage).\\nServer-side: Applies to current session (Cookie).');
+          alert('ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\nExperiment: ${experimentName}\\nVariation: ${variationName}\\n\\nMethod: Logged-out assignment, expires in ' + duration + ' hours\\nClient-side: Applies to the current domain (LocalStorage).\\nServer-side: Applies to current session (Cookie).');
         } else {
-          alert('ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\nExperiment: ${experimentName}\\nVariation: ${variationName}\\n\\nApplies to the current logged-in user.');
+          alert('ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\nExperiment: ${experimentName}\\nVariation: ${variationName}\\n\\nMethod: Logged-in assignment\\nApplies to the current logged-in user.');
         }
     }
 })()`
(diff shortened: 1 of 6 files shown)


src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -75,10 +75,23 @@ function assignmentHref(variationName: string, experimentName: string) {
...
+            }
+          )
+        );
+
         if (responseBody.storage_method === 'cookie') {
-          alert('Your current session has been assigned to ' + responseBody.variations.${experimentName} + ' for ' + duration + ' hours via cookie.');
+          alert('ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\nExperiment: ${experimentName}\\nVariation: ${variationName}\\nTTL: Expires in 1 hour\\n\\nClient-side: Applies to the current domain (LocalStorage).\\nServer-side: Applies to current session (Cookie).');
         } else {
-          alert('Your logged in user has been assigned to ' + responseBody.variations.${experimentName} + '. If you want to assign a different user, run this bookmarklet outside of Abacus.');
+          alert('ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\nExperiment: ${experimentName}\\nVariation: ${variationName}\\nTTL: Expires in 1 hour\\n\\nClient-side: Applies to the current domain (LocalStorage).\\nServer-side: Applies to current logged-in user (User attribute).');
```
yanirs, line 94: This is incorrect, as the server-side assignment doesn't necessarily use a user attribute and its TTL is not an hour. It won't change on the server side unless there's another bookmarklet call.

src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -75,10 +75,23 @@ function assignmentHref(variationName: string, experimentName: string) {
...
+              'variationName': '${variationName}',
+              'retrievedTimestamp': Date.now(),
+              'ttl': 60*60,
+            }
+          )
+        );
+
         if (responseBody.storage_method === 'cookie') {
-          alert('Your current session has been assigned to ' + responseBody.variations.${experimentName} + ' for ' + duration + ' hours via cookie.');
+          alert('ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\nExperiment: ${experimentName}\\nVariation: ${variationName}\\nTTL: Expires in 1 hour\\n\\nClient-side: Applies to the current domain (LocalStorage).\\nServer-side: Applies to current session (Cookie).');
```
yanirs, line 92: This is inaccurate given the situation with first/third-part cookies and the cookie's expiry, which is two hours.

src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -75,10 +75,23 @@ function assignmentHref(variationName: string, experimentName: string) {
         }
         const duration = responseBody.duration === 'unlimited' ?
           responseBody.duration : Math.ceil(responseBody.duration / 60 / 60);
+
+        window.localStorage.setItem(
```
yanirs, line 79: For now, it probably makes more sense to only store in localStorage when the storage method is cookie, as it implies that this was run outside Abacus. Storing in the localStorage of Abacus will only become useful when we start A/B testing it. :smile:

jessie-ross:
**New alerts:**

Screen Shot 2021-04-29 at 9 02 17 pm

Screen Shot 2021-04-29 at 9 02 46 pm

withinboredom (approved):

PR #511, opened 2021-04-29
Improve bookmarklet UX

jessie-ross:
**While I am here I have improved the bookmarklet UX:**
- Makes it more obvious that manual assignment is available
- Fixes some UI bugs

Screen Shot 2021-04-29 at 2 18 02 pm

Screen Shot 2021-04-29 at 2 21 51 pm

This PR also fixes 3 unit test `console.error`s, leaving one remaining.

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/overview/VariationsTable.tsx
+++ b/src/components/experiments/single-view/overview/VariationsTable.tsx
@@ -1,4 +1,4 @@
-import { Chip, Tooltip, Typography } from '@material-ui/core'
+import { Chip, Link, Tooltip, Typography } from '@material-ui/core'
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
 import Table from '@material-ui/core/Table'
 import TableBody from '@material-ui/core/TableBody'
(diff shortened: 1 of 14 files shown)


src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -40,8 +40,11 @@ const useStyles = makeStyles((theme: Theme) =>
 )
 
 function assignmentHref(variationName: string, experimentName: string) {
-  nameSchema.validateSync(variationName)
+  if (!experimentName) {
+    return ''
+  }
```
jessie-ross, line 45: This fixes a a unit test console.error where a page can be rendered initially without an experiment name while it is loading.

aaronyan: Thanks for the context! It may be valuable to add a comment here sharing this context.

aaronyan (approved):
LGTM! Thanks for making this experience easier for experimenters.

src/components/experiments/single-view/overview/VariationsTable.tsx
jessie-ross: I think it should be fine to go without a comment as returning nothing on no experimentName isn't out of the ordinary for Abacus.

PR #507, opened 2021-04-21
Add experiment run time indicator

jessie-ross:
**This PR adds the experiment run time indicator:**

Screen Shot 2021-04-21 at 10 15 07 pm

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)

PR #513, opened 2021-04-30
Add/experiment length indicator

jessie-ross:
Redo #507 which I accidentally merged into the wrong branch.


withinboredom (commented):
This doesn't look like the same PR :) the branch might need a `git rebase --onto` to remove the misc commits.

PR #515, opened 2021-05-03
Health: Soften indicators and add recommendations

jessie-ross:
**This PR:**
- Downgrades Kruschke and experiment duration errors to warnings
- Adds recommendations

## How has this been tested?

### DIFF
--- a/src/lib/analyses.test.ts
+++ b/src/lib/analyses.test.ts
@@ -580,6 +580,7 @@ describe('getExperimentParticipantStatHealthIndicators', () => {
           "indication": Object {
             "code": "probable issue",
             "reason": "−∞ < x ≤ 0.001",
+            "recommendation": "Contact @experimentation-platform-project",
             "severity": "Error",
           },
           "link": "https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#assignment-distribution-matching-allocated",
@@ -591,6 +592,7 @@ describe('getExperimentParticipantStatHealthIndicators', () => {
           "indication": Object {
             "code": "probable issue",
             "reason": "−∞ < x ≤ 0.001",
+            "recommendation": "Contact @experimentation-platform-project",
             "severity": "Error",
           },
           "link": "https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#assigned-no-spammers-no-crossovers-distribution-matching-allocated",
@@ -602,6 +604,7 @@ describe('getExperimentParticipantStatHealthIndicators', () => {
           "indication": Object {
             "code": "possible issue",
             "reason": "0.001 < x ≤ 0.05",
+            "recommendation": "Ensure exposure event fires regardless of which variation a participant receives.",
             "severity": "Warning",
           },
           "link": "https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#exposure-event-distribution-matching-allocated-sample-ratio-mismatch",
@@ -622,9 +625,10 @@ describe('getExperimentParticipantStatHealthIndicators', () => {
         },
         Object {
           "indication": Object {
-            "code": "very high",
-            "reason": "0.3 < x ≤ 1",
-            "severity": "Error",
+            "code": "high",
+            "reason": "0.1 < x ≤ 0.4",
+            "recommendation": "Spammers don't affect experiments, but high numbers could indicate other problems.",
+            "severity": "Warning",
           },
           "link": "https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#total-spammers",
           "name": "Ratio of spammers to assigned",
@@ -691,6 +695,7 @@ describe('getExperimentParticipantStatHealthIndicators', () => {
           "indication": Object {
             "code": "value error",
             "reason": "Unexpected value",
+            "recommendation": "Contact @experimentation-platform-project",
             "severity": "Error",
           },
           "link": "https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#total-crossovers",
@@ -702,6 +707,7 @@ describe('getExperimentParticipantStatHealthIndicators', () => {
           "indication": Object {
             "code": "value error",
             "reason": "Unexpected value",
+            "recommendation": "Contact @experimentation-platform-project",
             "severity": "Error",
           },
           "link": "https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#total-spammers",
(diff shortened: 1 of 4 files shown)


src/lib/analyses.test.ts
```diff
@@ -752,11 +758,12 @@ describe('getExperimentAnalysesHealthIndicators', () => {
           "indication": Object {
             "code": "very high",
             "reason": "1.5 < x ≤ ∞",
-            "severity": "Error",
+            "recommendation": "Results are very imprecise, be careful about drawing conclusions. Extend for more precision",
```
withinboredom, line 761:
```suggestion
            "recommendation": "Results are very imprecise, be careful about drawing conclusions. Continue running the experiment for more precision",
```

maybe? I just feel like "extend" is vague... this is really up to you though.

src/lib/analyses.ts
```diff
@@ -380,13 +387,15 @@ export function getExperimentParticipantHealthIndicators(
...
             severity: HealthIndicationSeverity.Error,
+            recommendation: contactUsRecommendation,
           },
         },
         {
           max: 0.05,
           indication: {
             code: HealthIndicationCode.PossibleIssue,
             severity: HealthIndicationSeverity.Warning,
+            recommendation: `Check daily ratio patterns for anomalies, contact @experiment-platform-project`,
```
withinboredom, line 398: Might not be for this PR, but I wonder if providing a list of recommendations or making them composable would be a little more maintainable? For example, this is a mix of "check daily ratio patterns for anomalies" and `contactUsRecommendation`.

src/lib/analyses.ts
```diff
@@ -443,13 +454,15 @@ export function getExperimentParticipantHealthIndicators(
           indication: {
             code: HealthIndicationCode.ProbableIssue,
             severity: HealthIndicationSeverity.Error,
+            recommendation: `If not in combination with other distribution issues, exposure event being fired is linked to variation causing bias. Choose a different exposure event or use assignment analysis.`,
```
withinboredom, line 457: > Choose a different exposure event or use assignment analysis.

I wish we could change things after an experiment starts running -- especially since the analysis gets regenerated for the past 14 days anyway. Restarting the experiment and changing the code just to tweak the assignment event seems like overkill.

src/lib/analyses.ts
```diff
@@ -563,7 +580,7 @@ export function getExperimentAnalysesHealthIndicators(
       name: 'Kruschke Precision (CI to ROPE ratio)',
       value: diffCiWidth / ropeWidth,
       unit: HealthIndicatorUnit.Ratio,
-      link: 'https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#ci-width-to-rope-ratio',
+      link: 'https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#kruschke-precision',
```
withinboredom, line 583: btw, this link is broken for me.

src/lib/analyses.ts
```diff
@@ -577,13 +594,76 @@ export function getExperimentAnalysesHealthIndicators(
...
+            code: HealthIndicationCode.Nominal,
+            severity: HealthIndicationSeverity.Ok,
+          },
+        },
+        {
+          max: 31,
+          indication: {
+            code: HealthIndicationCode.High,
+            severity: HealthIndicationSeverity.Warning,
+            recommendation: 'Experiment is running long, stop experiment now.',
```
withinboredom, line 658:
```suggestion
            recommendation: 'Experiment is running long, stop experiment soon.',
```

maybe?

src/lib/analyses.ts
```diff
@@ -577,13 +594,76 @@ export function getExperimentAnalysesHealthIndicators(
...
+      value: Experiments.getExperimentRunHours(experiment) / 24,
+      unit: HealthIndicatorUnit.Days,
+      link: 'https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#experiment-run-time',
+      indicationBrackets: [
+        {
+          max: 3,
+          indication: {
+            code: HealthIndicationCode.VeryLow,
+            severity: HealthIndicationSeverity.Warning,
+            recommendation: 'Less than 3 days is generally too short to run an experiment.',
```
withinboredom, line 635:
```suggestion
            recommendation: 'Experiments should generally run at least 7 days before drawing conclusions.',
```

Maybe?

src/lib/analyses.ts
```diff
@@ -577,13 +594,76 @@ export function getExperimentAnalysesHealthIndicators(
...
+/**
+ * Get experiment health indicators for a experiment.
+ */
+export function getExperimentHealthIndicators(experiment: ExperimentFull): HealthIndicator[] {
+  const indicatorDefinitions = [
+    {
+      name: 'Experiment Run Time',
+      value: Experiments.getExperimentRunHours(experiment) / 24,
+      unit: HealthIndicatorUnit.Days,
+      link: 'https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#experiment-run-time',
```
withinboredom, line 628: This link is broken for me as well. FWIW, this is what the page looks like for me:

image

withinboredom (approved):
LGTM, but I did bikeshed a bit on your wording. :)

I did notice that the links don't have matching sections in the wiki. Is that on purpose?

src/lib/analyses.ts
withinboredom: That was my thought as well. The code won't benefit from it right now.

src/lib/analyses.ts
jessie-ross: Ah I meant manually changing the exposure event for this experiment - I might add `contact @experiment-platform-group to do so.` at the end of it.

As an aside, I would love some remodelling of the data where an experiment slug has multiple analyses - you could theoretically run the same experiment with different targeting and exposure groups all at the same time.

src/lib/analyses.ts
withinboredom: The reason I think saying 7 days is because it would set people's expectations. If they were planning to run it for two days and see "<3 days" they might wait an extra day and then be annoyed to see "<7 days"

src/lib/analyses.ts
jessie-ross: Yep I need to update the wiki to fit these.

src/lib/analyses.ts
withinboredom: The issue we had with Hypothesis doing that is that different analyses weren't necessarily comparable. But if we could add new/change metrics/exposure events during the first 48 hours or something, that'd be :100:. Those things don't really affect the runtime (like segments).

jessie-ross:
Nice job picking up those links, I was planning to update the wiki but I never got around to it. I will add an issue for that, in the mean-time I think the recommendations are doing some lifting there.

PR #517, opened 2021-05-03
Use sentence case for indicator names

jessie-ross:
**Use sentence case for indicator names:**

- Consistent
- Yanir approved

## How has this been tested?


withinboredom (approved):
nice catch

PR #518, opened 2021-05-04
Fix health report issues and tweak messaging

yanirs:
Fix a bunch of issues I noticed with the health report and messaging:

* Bug fix: Change `Assignment distribution without crossovers and spammers` to actually show that rather than the overall assignment distribution
* Replace `@experimentation-review-guild` and `@experiment-review-guild` with `@experiment-review` (the correct Slack group)
* Change "Kruschke precision" to "Kruschke uncertainty", as it aligns more with what it is (high number = high uncertainty), and change the recommendation wording accordingly. 
* Change the "high" experiment run time threshold from 31 to 42, as it's generally better to have measurements for whole weeks and 31 is too close to 28.
* Add periods at the end of recommendations (some had them and some didn't) and tweak some other phrasing.
* Remove `assignedSpammersDistributionMatchingAllocated` references and calculations since it doesn't get used.

## How has this been tested?


jessie-ross (approved):
LGTM and thanks for picking these up!

PR #520, opened 2021-05-04
Add early monitoring and improve empty results page

jessie-ross:
**This PR:**
- Adds early monitoring query
- Improves no results information

Screen Shot 2021-05-04 at 4 28 57 pm

Resolves 534-gh-Automattic/experimentation-platform

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/ActualExperimentResults.tsx
+++ b/src/components/experiments/single-view/results/ActualExperimentResults.tsx
@@ -119,6 +119,9 @@ const useStyles = makeStyles((theme: Theme) =>
     accordionDetails: {
       flexDirection: 'column',
     },
+    noAnalysesPaper: {
+      padding: theme.spacing(2),
+    },
     pre: {
       background: '#f5f5f5',
       padding: theme.spacing(3),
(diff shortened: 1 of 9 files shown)


src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -307,97 +326,155 @@ export default function ActualExperimentResults({
...
-        Health Report
-      </Typography>
-      <Paper id='health-report'>
-        <HealthIndicatorTable indicators={experimentHealthIndicators} />
-      </Paper>
+select
+  experiment_variations.name as variation_name,
+  raw_number_of_assignments
+from tracks_counts
+inner join experiment_variations on cast(tracks_counts.experiment_variation_id as bigint) = experiment_variations.experiment_variation_id`}
```
aaronyan, line 473: I think we'll want to specify the database `wpcom.experiment_variations` in case they are not already in the `wpcom` database when executing the query.

aaronyan (approved):
Woo love the improved and more informative page. LGTM!

I really like the call to action for monitoring assignments using that query. I tested the query, and works as expected!

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -307,97 +326,155 @@ export default function ActualExperimentResults({
...
+      )}
+
+      <Accordion className={classes.accordion}>
+        <AccordionSummary expandIcon={<ExpandMoreIcon />}>
+          <Typography variant='h5'>Early Monitoring - Live Assignment Count</Typography>
+        </AccordionSummary>
+        <AccordionDetails className={classes.accordionDetails}>
+          <Typography variant='body1'>
+            For early monitoring, you can run this query in Hue to retrieve raw unfiltered assignment counts from
+            Tracks.
```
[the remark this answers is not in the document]
yanirs: These results can be quite alarming since data gets deleted from `etl_events`. This needs further clarification on how it works.

Perhaps add something like "This query should be used to monitor event flow. The best way to use it is to run it multiple times and ensure that counts go up and are roughly distributed as expected. Counts may also go down as events are moved to `prod_events` every day."

src/components/experiments/single-view/results/ActualExperimentResults.tsx
yanirs: Definitely, it'll break otherwise.

PR #525, opened 2021-05-07
Change the early monitoring query to count uniques

yanirs:
Following the confusion overnight, change the early monitoring query to count unique users so that we don't get false alarms on duplicate assignment events. Also tweak the query a bit to make it less wide. :slightly_smiling_face: 

## How has this been tested?

- Additional manual testing instructions: Ran the query on a couple of different experiments to verify that the results make sense


jessie-ross (approved):
LGTM!

TIL about the `distinct` keyword

PR #527, opened 2021-05-07
Fix missing analysis crash

### LINKED ISSUE #436: Positively Signal when an Experiment is ready to be stopped

jessie-ross:
**This PR fixes the missing analysis crash:**

Screen Shot 2021-05-07 at 10 21 25 pm

Resolves #436

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
+++ b/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
@@ -71,6 +71,19 @@ const useStyles = makeStyles((theme: Theme) =>
   }),
 )
 
+function MissingAnalysisMessage() {
+  const classes = useStyles()
+  return (
+    <div className={classes.root}>
+      <Typography variant='h5' gutterBottom>
+        {' '}
+        No Analysis Data Found{' '}
+      </Typography>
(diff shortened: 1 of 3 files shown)


src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -71,6 +71,19 @@ const useStyles = makeStyles((theme: Theme) =>
   }),
 )
 
+function MissingAnalysisMessage() {
+  const classes = useStyles()
+  return (
+    <div className={classes.root}>
+      <Typography variant='h5' gutterBottom>
+        {' '}
+        No Analysis Data Found{' '}
```
aaronyan, line 80: I'd prefer sentence case, but looking around at the other headers, it looks like this is the style! :) Just noting it for now.

aaronyan (approved):
LGTM. Nice and simple!

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: This one is a header

PR #529, opened 2021-05-07
Hide top level issue display when primary analyses are missing

jessie-ross:
**This PR hides top level issues when analyses are missing** 

In the future we can look into conditionally displaying the indicators as a more long term fix.

From p1620396357160000-slack-C7HH3V5AS

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/ActualExperimentResults.tsx
+++ b/src/components/experiments/single-view/results/ActualExperimentResults.tsx
@@ -223,13 +223,16 @@ export default function ActualExperimentResults({
     ...Analyses.getExperimentHealthIndicators(experiment),
   ]
 
-  const maxIndicationSeverity = experimentHealthIndicators
-    .map(({ indication: { severity } }) => severity)
-    .sort(
-      (severityA, severityB) =>
-        Analyses.healthIndicationSeverityOrder.indexOf(severityB) -
-        Analyses.healthIndicationSeverityOrder.indexOf(severityA),
-    )[0]
+  const maxIndicationSeverity =
+    primaryMetricAggregateRecommendation.decision === Analyses.AggregateRecommendationDecision.MissingAnalysis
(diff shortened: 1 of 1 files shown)


src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -223,13 +223,16 @@ export default function ActualExperimentResults({
...
 
-  const maxIndicationSeverity = experimentHealthIndicators
-    .map(({ indication: { severity } }) => severity)
-    .sort(
-      (severityA, severityB) =>
-        Analyses.healthIndicationSeverityOrder.indexOf(severityB) -
-        Analyses.healthIndicationSeverityOrder.indexOf(severityA),
-    )[0]
+  const maxIndicationSeverity =
+    primaryMetricAggregateRecommendation.decision === Analyses.AggregateRecommendationDecision.MissingAnalysis
```
aaronyan, line 227: I wasn't too sure about the boundaries of this condition, but [it is being used here](https://github.com/Automattic/abacus/blob/d737dcc99fca0ec86735279d9859a45f95aebfe0/src/components/experiments/single-view/results/ActualExperimentResults.tsx#L298) to disable the detail panel.

aaronyan (commented):
LGTM! Thanks for coming in with a quick fix. We should have a test for this, right?

yanirs:
> We should have a test for this, right?

+1

jessie-ross:
This is a bit annoying to add a specific unit test for, it is however tested as part of the general missing analysis flow.

yanirs:
> This is a bit annoying to add a specific unit test for, it is however tested as part of the general missing analysis flow.

Maybe move some of the logic to fully-covered functions (`getMaxIndicationSeverity()` in this case)? It feels like these edge cases keep slipping through the cracks...

jessie-ross:
Superceded by #543

PR #534, opened 2021-05-13
Undo diff inversion hack

jessie-ross:
**This PR removes the diff inversion hack.**

Needs to be merged in coordination with the ETL fix.

Part of Automattic/experimentation-platform#465

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)

PR #536, opened 2021-05-20
Implement top level diff changes

jessie-ross:
**This PR implements top level absolute and relative diffs.**

As seen in slack

Part of #419 

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/ActualExperimentResults.tsx
+++ b/src/components/experiments/single-view/results/ActualExperimentResults.tsx
@@ -24,6 +24,7 @@ import { PlotData } from 'plotly.js'
 import React, { useState } from 'react'
 import Plot from 'react-plotly.js'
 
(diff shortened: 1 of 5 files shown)


aaronyan (approved):
I think the sign on the absolute change needs to be adjusted to be consistent with the relative change. I would expect the absolute change to be `treatment - control` since the lift is `(treatment - control)/control`. Here's a screenshot of `control` winning in the later metrics, showing a positive absolute change but negative lift.

Other than that, LGTM! I'm excited for this PR to land. 🚀 

Screen Shot 2021-05-20 at 11 02 38 AM

jessie-ross:
> I think the sign on the absolute change needs to be adjusted to be consistent with the relative change (both top level and in expanded view). I would expect the absolute change to be treatment - control since the lift is (treatment - control)/control. Here's a screenshot of control winning in the later metrics, showing a positive absolute change but negative lift.

Ah that is because this PR is seperate to #534

PR #539, opened 2021-05-26
Wizard: Improve metric input

### LINKED ISSUE #378: Improve metric assignments in experiment creation wizard
Expanded this issue to go beyond ordering metric assignments alphabetically during metric assignment in the experiment creation wizard. We can also consider enhancing the metric discovery and assignment process!

Instead of a dropdown menu, we can consider having 1/3 of the page be a section dedicated to searching and looking for metrics. Once we find a metric we like, we can hit a + button on the side of a metric to attach it to the experiment. The results of this search section would basically be the metrics directory view.

This issue was updated in part because of [experimenter feedback](https://betterexperiments.wordpress.com/2020/12/01/removing-references-to-legacy-frontend-a-b-tests-calypso-abtest/#comment-1669)!

jessie-ross:
**This PR:**
- Highlights that there are tooltips for metric descriptions
- Switches the metric box for an autocomplete, allowing searching
- Adds metric descriptions to the metrics in the autocomplete

Video to be posted to slack

Resolves #378, I went in the direction of improving the autocomplete instead of adding the table in.
It might be worth adding the metric table in later but we would need to improve the metric table quite a bit first, possibly changing it to use AG-Grid. I think we also need to create one line "Short descriptions" for metrics for an effective table view.

**Future PR:**
- Refactor this code and include it in the "Add Metric" dialog.

(BTW, seems like we had the right data already)

## How has this been tested?

### DIFF
--- a/src/components/experiments/wizard/BasicInfo.tsx
+++ b/src/components/experiments/wizard/BasicInfo.tsx
@@ -148,6 +148,7 @@ const BasicInfo = ({
           fullWidth
           options={userCompletionDataSource.data ?? []}
           loading={userCompletionDataSource.isLoading}
+          noOptionsText='No users found'
           renderInput={(params: AutocompleteRenderInputParams) => (
             <MuiTextField
               {...params}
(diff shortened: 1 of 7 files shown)


aaronyan (approved):
This is amazing. I'm a big fan of the descriptions and formatting of the metric results. LGTM!

PR #540, opened 2021-05-27
Rearrange panels

### LINKED ISSUE #475: Experiment Overview UI Overflow
It seems like long metrics can cause the grid to overflow into the Audience area, we should truncate them or fix otherwise.

jessie-ross:
**This PR rearranges the Experiment overview panels:**
- Makes conclusion more prominent
- Gives more space for metrics

Resolves #475 

See demo in slack thread.

## How has this been tested?


jessie-ross:
I didn't want to go down the truncation with title solution as I want the hover to be available to show metric descriptions.

aaronyan (approved):
This looks great! Thanks for helping to solve this so quickly @jessie-ross.

We can consider having both the metric name and description in the tooltip, which would allow us to truncate if needed. Having conclusions so high up feels a little unnatural to me, but I think we should ship this for now and we can iterate on that design later. :)

PR #542, opened 2021-05-31
Make analysis strategy switch generally available

jessie-ross:
**This PR makes the analysis strategy switch generally available:**
- Limits the options to what is available
- Specifies which option is recommended
- Changes the debug mode key, hiding it again from experimenters.

Can't find a reference in slack, I think me and Yanir talked about opening it up in our 1:1?
- Experimenters already know about it.
- It is educational being able to explore it.

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/ActualExperimentResults.tsx
+++ b/src/components/experiments/single-view/results/ActualExperimentResults.tsx
@@ -119,7 +119,7 @@ const useStyles = makeStyles((theme: Theme) =>
     tableTitle: {
       margin: theme.spacing(4, 2, 2),
     },
-    accordion: {
+    accordions: {
       margin: theme.spacing(2, 0),
     },
     accordionDetails: {
@@ -436,21 +436,6 @@ export default function ActualExperimentResults({
 
   return (
     <div className={classes.root}>
-      <Paper className={classes.advancedControls}>
-        <FormControl>
-          <InputLabel htmlFor='strategy-selector' id='strategy-selector-label'>
-            Analysis Strategy:
-          </InputLabel>
-          <Select id='strategy-selector' labelId='strategy-selector-label' value={strategy} onChange={onStrategyChange}>
-            {availableAnalysisStrategies.map((strat) => (
-              <MenuItem key={strat} value={strat}>
-                {Analyses.AnalysisStrategyToHuman[strat]}
-                {strat === Experiments.getDefaultAnalysisStrategy(experiment) && ' (recommended)'}
-              </MenuItem>
-            ))}
-          </Select>
-        </FormControl>
-      </Paper>
       {hasAnalyses ? (
         <>
           <div className={classes.summary}>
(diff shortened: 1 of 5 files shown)


src/utils/general.ts
```diff
@@ -17,7 +17,7 @@ export function createUnresolvingPromise<T>(): Promise<T> {
   return new Promise<T>(() => null)
 }
 
-const debugModeLocalStorageKey = 'abacus-debug-mode'
+const debugModeLocalStorageKey = `Rob, this totally isn't the debug mode.`
```
yanirs, line 20: :tada: :joy:

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -434,23 +436,21 @@ export default function ActualExperimentResults({
...
-          </Paper>
-        )
-      }
+      <Paper className={classes.advancedControls}>
+        <FormControl>
+          <InputLabel htmlFor='strategy-selector' id='strategy-selector-label'>
+            Analysis Strategy:
+          </InputLabel>
+          <Select id='strategy-selector' labelId='strategy-selector-label' value={strategy} onChange={onStrategyChange}>
+            {availableAnalysisStrategies.map((strat) => (
```
yanirs, line 445: Not critical at all, but `strat` looks like misspelled `start` at a glance. Spelling it out as `strategy` would be nicer. :slightly_smiling_face:

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -157,13 +157,15 @@ export default function ActualExperimentResults({
   const classes = useStyles()
   const theme = useTheme()
 
+  const availableAnalysisStrategies = [
+    AnalysisStrategy.IttPure,
+    AnalysisStrategy.MittNoCrossovers,
+    AnalysisStrategy.MittNoSpammers,
+    AnalysisStrategy.MittNoSpammersNoCrossovers,
+  ]
+  experiment.exposureEvents && availableAnalysisStrategies.push(AnalysisStrategy.PpNaive)
```
yanirs, line 166: I think this would be a bit more readable.

```suggestion
  if (experiment.exposureEvents) {
    availableAnalysisStrategies.push(AnalysisStrategy.PpNaive)
  }
```

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -434,23 +436,21 @@ export default function ActualExperimentResults({
...
-                  </MenuItem>
-                ))}
-              </Select>
-            </FormControl>
-          </Paper>
-        )
-      }
+      <Paper className={classes.advancedControls}>
+        <FormControl>
+          <InputLabel htmlFor='strategy-selector' id='strategy-selector-label'>
```
yanirs, line 441: Since this is more of an advanced setting, what do you think of moving it to the bottom of the page (just below or just above the Early Monitoring panel)? It'd also be nice to make the panel full width, add help text, and make the heading consistent with the Early Monitoring panel.

Something like:

> ##### Advanced: Choose an Analysis Strategy
>
> Choosing a different analysis strategy is useful for checking the effect of different modelling decisions on the results. These are the available strategies:
> * _All participants_: Pure intention-to-treat – all the participants are analysed based on their initial variation assignment.
> * _Without crossovers_:  Modified intention-to-treat – same as _all participants_, but excluding participants that were assigned to multiple experiment variations before or on the analysis date (aka crossovers).
> * _Without spammers_: Modified intention-to-treat – same as _all participants_, but excluding spammers that were flagged on the analysis date.
> * _Without crossovers and spammers_: Modified intention-to-treat – same as _all participants_, but excluding both spammers and crossovers.
> * _Exposed without crossovers and spammers_: Naive per-protocol – only participants that triggered one of the experiment's exposure events, excluding both spammers and crossovers. This analysis strategy is only available if the experiment has exposure events, while the other four strategies are used for every experiment.
>
> `<Select ...>`

Like the Early Monitoring panel, we can collapse this panel by default, but I'm not too fussed.

jessie-ross: Great idea!

jessie-ross: This is how it is looking:

Screen Shot 2021-06-01 at 10 09 17 pm
Screen Shot 2021-06-01 at 10 09 10 pm

yanirs (approved):
LGTM. :tada: 

Small UX note: Now that the selector is at the bottom, it may be worth adding a visual indication that something has changed when the selected strategy changes. For example, scroll slowly to the top of the page. Otherwise, the user doesn't get a strong indication that anything changed, and they have to scroll manually to the top to see the changes.

jessie-ross:
I am trying out an "Advanced" show-hide panel higher up, see the video in Slack:
- Makes the controls more visible
- Shows the controls are advanced
- Extensible to other areas

I am not such a fan of the double accordion, maybe we could make it more a set of controls:
- Taking the strategy select out of the accordion, adding an info tooltip with all the info in it.
- Making the "Live event flow" a modal behind a button.

But I think this is a good for-now step.

jessie-ross:
> Small UX note: Now that the selector is at the bottom, it may be worth adding a visual indication that something has changed when the selected strategy changes. For example, scroll slowly to the top of the page. Otherwise, the user doesn't get a strong indication that anything changed, and they have to scroll manually to the top to see the changes.

I was looking into flashing the page but it seems like to much work for this, I ended up adding some helper text under the select:

Screen Shot 2021-06-11 at 1 29 27 am

Alternatively I could use a snackbar notification, but that seems overkill. As for the scrolling, I think that might be annoying - the real fix there would be to have it at the top or at the side.

yanirs:
Looks good. I think that having it at the bottom and the note is fine for now.

PR #543, opened 2021-06-03
Fix missing analysis logic

jessie-ross:
**This PR corrects our missing analysis check.**

Even though analyses exist they can have null metric estimates, we use the pre-existing aggregateRecommendationDecision to check.

Resolves issue in p1620396357160000-slack-C7HH3V5AS

A better fix for #529

## How has this been tested?


yanirs (approved):
:ship:

PR #547, opened 2021-06-07
Update bookmarklet to pass in anon ID

yanirs:
As part of the change to backend storage systems, update the bookmarklet code to pass in the anon ID from the `tk_ai` cookie (if available) and handle the updated return values from D62396-code.

## How has this been tested?

Applied D62396-code on the backend, sandboxed the API, and tested three scenarios on a production A/A experiment:
* Authenticated user in Abacus
* A page with logged out IDs and a correct `tk_ai` cookie (e.g., logged out WP.com)
* No IDs passed in (e.g., a non-WP.com site)

Checked that the backend tables get written as expected.

### DIFF
--- a/src/components/experiments/single-view/overview/VariationsTable.tsx
+++ b/src/components/experiments/single-view/overview/VariationsTable.tsx
@@ -47,6 +47,7 @@ function assignmentHref(variationName: string, experimentName: string) {
   nameSchema.validateSync(variationName)
   return `javascript:(async () => {
     const token = JSON.parse(localStorage.getItem('experiments_auth_info'));
+    const anonId = document.cookie.match('(^|;)\\\\s*tk_ai\\\\s*=\\\\s*([^;]+)')?.pop() || '';
     const headers = {'Content-Type': 'application/json'};
     if (token && token.accessToken) {
        headers.Authorization = 'Bearer ' + token['accessToken'];
@@ -57,7 +58,7 @@ function assignmentHref(variationName: string, experimentName: string) {
         credentials: 'include', 
         method: 'PATCH', 
         headers, 
-        body: JSON.stringify({variations: {${experimentName}: '${variationName}'}})
+        body: JSON.stringify({variations: {${experimentName}: '${variationName}'}, anon_id: anonId})
       }
     );
     const responseBody = await response.json();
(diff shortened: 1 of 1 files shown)


src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -71,30 +72,21 @@ function assignmentHref(variationName: string, experimentName: string) {
...
       default:
         if (!responseBody.variations) {
           alert('An unknown error occurred: ' + responseBody.message);
           break;
         }
-        const duration = responseBody.duration === 'unlimited' ?
-          responseBody.duration : Math.ceil(responseBody.duration / 60 / 60);
-
-        if (responseBody.storage_method === 'cookie') {
-          window.localStorage.setItem(
```
yanirs, line 83: No need for the local storage setting, as the assignment should persist on the backend as long as the same anon ID gets passed in.

aaronyan (approved):
LGTM! Cool to see custom messages per anon vs logged-in users.

src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -71,30 +72,21 @@ function assignmentHref(variationName: string, experimentName: string) {
...
-                'ttl': responseBody.duration === 'unlimited' ? Infinity : responseBody.duration,
-              }
-            )
-          );
-
-          alert('ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\nExperiment: ${experimentName}\\nVariation: ${variationName}\\n\\nMethod: Logged-out assignment, expires in ' + duration + ' hours\\nClient-side: Applies to the current domain (LocalStorage).\\nServer-side: Applies to current session (Cookie).');
+        const baseMessage = 'ExPlat: Successful Assignment\\n–––––––––––––––––––––––––––––\\n\\n' +
+          'Experiment: ${experimentName}\\nVariation: ${variationName}\\n\\n';
+        if (responseBody.storage_method === 'anon_sqooped_out_table') {
+          alert(baseMessage + 'Method: Logged-out assignment\\nApplies to the current anon user.');
```
jessie-ross, line 87:
```suggestion
          alert(baseMessage + 'Method: Logged-out assignment\\nApplies to the current anon user (tk_ai).');
```
I think mentioning the mechanism is useful as otherwise it can seem a bit black boxy.

yanirs: Good point :slightly_smiling_face:

src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -48,12 +48,14 @@ function assignmentHref(variationName: string, experimentName: string) {
   nameSchema.validateSync(variationName)
   return `javascript:(async () => {
     const token = JSON.parse(localStorage.getItem('experiments_auth_info'));
+    const anonId = document.cookie.match('(^|;)\\\\s*tk_ai\\\\s*=\\\\s*([^;]+)')?.pop() || '';
```
jessie-ross, line 51: I recommend keeping all information necessary in the `body` if we are sending a body rather than mixing between query params and body. One way to do that is to ensure the correct encoding of anon_id in this code:

```suggestion
    const anonId = decodeURIComponent(document.cookie.match('(^|;)\\\\s*tk_ai\\\\s*=\\\\s*([^;]+)')?.pop() || '');
```

That way we are making the encoding explicit instead of relying on PHPs automatic GET/cookie decoding.

yanirs: Sounds good. That was my first idea, but I figured that passing the `anon_id` as a query parameter matches the assignment GET.

Issue #524, opened 2021-05-06
Apply metric-level result redesign


aaronyan:
After the design and analysis adjustments are complete, we can then implement the experiment results summary table in Abacus!

**Edited by Yanir**: The discussion led to the design in pbmo2S-LK-p2#comment-1911. @jessie-ross can decide how to best split the work to PRs, but here are the main elements of the redesign:

- [x] Card-based design for the expanded metric assignment view
- [x] Summary card at the top, which includes an English explanation of the recommendation, the metric description, and last analysis date (see 641-gh-Automattic/experimentation-platform for the recommendation wording) 
- [x] Analysis card below the Summary card
- [x] Time series plots following the Analysis card
- [x] Collapsed observed data table as a panel at the bottom

jessie-ross:
This is looking great @aaronyan, thanks for pushing this idea forward!

Love the small details like putting attribution window into parentheses and it is easy to see how this extends to A/B/n - I think we just need a selector with the diff graph to show the difference between different variations.

It's all straightforward to build :-)

PR #551, opened 2021-06-10
Implement metric assignment result redesign (part 1)

jessie-ross:
**This PR implements part 1 of the redesign from https://github.com/Automattic/abacus/issues/524:**
- New cards layout with added summary card
- Moved and changed small print

Screenshot in slack.

**Future PRs:**
- Observations card
- Decision explanation
- Altering the outside table to include attr window

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
+++ b/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
@@ -272,7 +272,11 @@ export default function MetricAssignmentResults({
           <TableHead>
             <TableRow>
               <TableCell>Variant</TableCell>
-              <TableCell align='right'>Conv. rate interval</TableCell>
+              <TableCell align='right'>
+                {metric.parameterType === MetricParameterType.Revenue
+                  ? 'Average revenue per user (ARPU) interval'
+                  : 'Conversion rate interval'}
+              </TableCell>
               <TableCell align='right'>Absolute change</TableCell>
               <TableCell align='right'>Relative change (lift)</TableCell>
             </TableRow>
--- a/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
+++ b/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
@@ -9,7 +9,6 @@ import {
   TableHead,
   TableRow,
   Theme,
-  Tooltip,
   Typography,
 } from '@material-ui/core'
 import clsx from 'clsx'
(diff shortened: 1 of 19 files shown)


src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -227,41 +245,94 @@ export default function MetricAssignmentResults({
...
+                          />
+                          .
+                        </>
+                      }
                     >
-                      Difference
-                    </TableCell>
-                    <TableCell className={classes.monospace}>
+                      <span className={classes.tooltipped}>
+                        <MetricValue
```
yanirs, line 314: Per the design, the display of ranges in this table should match that of the top-level display, i.e., show the unit once. It is a bit odd to see something like `-0.21 to +0.2 USD` in the top-level display and `USD -0.21 to USD 0.2` in the details.

To aid with consistency, how about we add `MetricValueInterval` as a component?

jessie-ross:
@yanirs Hmm, actually the 'conversion' interval won't work for revenue metrics, what about "Metric value interval" or changing it to "ARPU interval" just for revenue 🤔

yanirs:
> @yanirs @aaronyan Hmm, actually the 'conversion' interval won't work for revenue metrics, what about "Metric value interval" or changing it to "ARPU interval" just for revenue

Good point. @aaronyan said in the original post:

> * Conversion metrics will have “conversions”, “conversion rate”, and “conversion rate intervals”.
> * Revenue metrics will have “revenue”, “ARPU”, and “ARPU intervals”.

So "ARPU interval" would work. Since we have the space, we can spell it out as "Average revenue per user interval" or "Revenue per user interval". Alternatively, it can be explained in a tooltip.

aaronyan (approved):
LGTM!

I'm not the biggest fan of the text between the analysis stats card and the graphs

> 95% Credible Intervals (CIs). Experimenter-set minimum practical difference: x.xx pp. Participants: n (n_1 in control; n_2 in treatment).

Since the summary card explanations (min diff amount), observational data card (participants), and graph legends (min diff in context) should have this information, I think we can remove this line in future diffs. :)

If we like having a header to describe the graphs, maybe we have something like?

> Historical 95% credible intervals

I don't have a strong opinion on the naming though. I think my preference is for there to not have any header, since we already have titles for the graphs.

jessie-ross:
> Since the summary card explanations (min diff amount), observational data card (participants), and graph legends (min diff in context) should have this information, I think we can remove this line in future diffs. :)

Yep thats what I was thinking too :)

> I don't have a strong opinion on the naming though. I think my preference is for there to not have any header, since we already have titles for the graphs.

I don't think we need a title for the graphs but I think we will need to mention somewhere that the analyses are 95% intervals.

PR #556, opened 2021-06-22
Metric assignment: observed data

jessie-ross:
**This PR adds a collapsible observed data table:**
- Uses the estimate data as we don't have actual observed data here - do we need actual observed data here? I am not sure what this is solving?
- Adds "sampled" to the title so Experimenters don't think this is the actual observed data.
- Ended up adding a very basic show/hide as there was nothing pre-existing in MUI

Screen Shot 2021-06-23 at 12 05 08 am

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
+++ b/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
@@ -29,7 +29,6 @@ import {
   MetricBare,
   MetricParameterType,
 } from 'src/lib/schemas'
-import * as Variations from 'src/lib/variations'
 import * as Visualizations from 'src/lib/visualizations'
 
 import MetricValueInterval from '../../../general/MetricValueInterval'
@@ -339,14 +338,7 @@ export default function MetricAssignmentResults({
           metricParameterType={metric.parameterType}
           isDifference={true}
         />
-        . <strong>Participants:</strong> {latestAnalysis.participantStats.total} (
-        {_.join(
-          Variations.sort(experiment.variations).map(
-            ({ variationId, name }) => `${latestAnalysis.participantStats[`variation_${variationId}`]} in ${name}`,
-          ),
-          '; ',
-        )}
-        ).
+        .
       </Typography>
       {dates.length > 1 ? (
         <div className={classes.metricEstimatePlots}>
--- a/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
+++ b/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
@@ -11,10 +11,11 @@ import {
   Theme,
   Typography,
 } from '@material-ui/core'
+import { ChevronRight, ExpandMore } from '@material-ui/icons'
 import clsx from 'clsx'
 import _, { identity } from 'lodash'
 import { PlotData } from 'plotly.js'
-import React from 'react'
+import React, { useState } from 'react'
 import Plot from 'react-plotly.js'
 
 import DatetimeText from 'src/components/general/DatetimeText'
@@ -112,6 +113,13 @@ const useStyles = makeStyles((theme: Theme) =>
         borderRightColor: theme.palette.grey[300],
       },
     },
+    clickable: {
+      cursor: 'pointer',
+      userSelect: 'none',
+    },
+    expandCollapseIcon: {
+      verticalAlign: 'middle',
+    },
   }),
 )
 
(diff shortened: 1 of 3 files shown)


src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -376,6 +381,67 @@ export default function MetricAssignmentResults({
...
+        className={clsx(classes.dataTableHeader, classes.clickable)}
+        onClick={toggleIsShowObservedData}
+        role='button'
+      >
+        Observed data (sampled){' '}
+        {isShowObservedData ? (
+          <ExpandMore className={classes.expandCollapseIcon} />
+        ) : (
+          <ChevronRight className={classes.expandCollapseIcon} />
+        )}{' '}
```
aaronyan, line 394: > Ended up adding a very basic show/hide as there was nothing pre-existing in MUI

Cool, looks and feels good to me. I think I prefer the chevron on the left, since all the other hide/expand are on the left. It looks like a pretty tight fit though. I like that you've allowed the whole row to be clickable, so even if we shrink it to the left, it could work well.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -376,6 +381,67 @@ export default function MetricAssignmentResults({
...
+                    >
+                      <span className={classes.monospace}>{variation.name}</span>
+                    </TableCell>
+                    <TableCell className={classes.monospace} align='right'>
+                      {latestAnalysis.participantStats[`variation_${variation.variationId}`].toLocaleString()}
+                    </TableCell>
+                    <TableCell className={classes.monospace} align='right'>
+                      {(
+                        latestAnalysis.participantStats[`variation_${variation.variationId}`] *
+                        latestEstimates[`variation_${variation.variationId}`].estimate
```
aaronyan, line 432: > I am not sure what this is solving?

In my eyes, adding this observational data helps experimenters understand how to interpret their results. These are metrics that are very easy to understand and are gateways to understanding more complex measures like the credible intervals. These are also measures I think most experimenters will look for and having them available is a familiar and comfortable experience.

I preferred having the observational data at the top to help tell the story of easy to more complex analysis (like here are the ingredients we used to calculate credible intervals). But we came to a compromise to show the data but have it hidden as to not emphasize it.

> Uses the estimate data as we don't have actual observed data here - do we need actual observed data here? 

I believe this is considered observed experiment data. The `estimate` is calculated by taking the `conversions / # of users`. I think we return only the `# of users` and `estimate`, but the way you've calculated `conversions` looks right to me.

What did you have in mind when you say actual observed data?

> Adds "sampled" to the title so Experimenters don't think this is the actual observed data.

I'm not the biggest fan since "sampled" doesn't add any information for me; I think this is all considered experiment data and I'm having trouble wrapping my head around sampled vs non-sampled data in this context. I prefer to remove it, but let's first discuss here on where you're coming from.

aaronyan (approved):
LGTM! Thanks for adding this so fast. I added some discussion comments, no blockers. :)

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: > In my eyes, adding this observational data helps experimenters understand how to interpret their results. These are metrics that are very easy to understand and are gateways to understanding more complex measures like the credible intervals. These are also measures I think most experimenters will look for and having them available is a familiar and comfortable experience.
>
> I preferred having the observational data at the top to help tell the story of easy to more complex analysis (like here are the ingredients we used to calculate credible intervals). But we came to a compromise to show the data but have it hidden as to not emphasize it.

Ah yeah the telling a story part makes a lot of sense.

> I believe this is considered observed experiment data. The estimate is calculated by taking the conversions / # of users. I think we return only the # of users and estimate, but the way you've calculated conversions looks right to me.

Oh, looking at the analyses code again you seem to be right, I had thought we were using the mean from a calculated posterior distribution for the estimate.

I will remove the "sampled" label.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -376,6 +381,67 @@ export default function MetricAssignmentResults({
...
+      </Typography>
+      {isShowObservedData && (
+        <TableContainer component={Paper}>
+          <Table className={classes.coolTable}>
+            <TableHead>
+              <TableRow>
+                <TableCell>Variant</TableCell>
+                <TableCell align='right'>Users</TableCell>
+                <TableCell align='right'>
+                  {metric.parameterType === MetricParameterType.Revenue ? 'Revenue (USD)' : 'Conversions'}
```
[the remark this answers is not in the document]
jessie-ross: For the moment I think they are all different enough to keep seperate, but I'll keep in mind for the future.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: We actually have them both to the left and the right in different situations, e.g. the advanced accordion below, but I am happy to move it to the left.

Issue #554, opened 2021-06-14
Show partial results without a primary metric analysis


yanirs:
When looking into a recent A/A experiment, I noticed that the result page shows the "No Results" message even though it has thousands of participants. This appears to be due to the change in #543: The primary metric has no metric estimates, but participant counts and secondary metric estimates do exist. I think we should still show the results we have in such cases.

Marking this as a medium priority bug because it'd be experienced as a bug by experimenters: They get a Slack message saying that there are thousands of participants, but when they click the link to see the results it says "No Results". I think it's medium priority because it's unlikely to happen much in practice &ndash; in A/A experiments it may happen because we randomly assign metrics that have nothing to do with the experiment and the backend still doesn't handle the all-zero case well.

As part of the fix, it'd be good to pull out some of the logic from `ActualExperimentResults` and test it, as we seem to keep hitting bugs around the display of missing analyses. I think it also makes sense to show the analysis strategy selector if only some strategies have participants.

jessie-ross:
Oh I had assumed that if an experiment didn't have primary MA estimates the secondaries wouldn't have either.

Can you link the A/A in question?

yanirs:
> Oh I had assumed that if an experiment didn't have primary MA estimates the secondaries wouldn't have either.

It's a fair assumption. It technically shouldn't happen with a properly-configured experiment, and should be better once 574-gh-Automattic/experimentation-platform is resolved.

> Can you link the A/A in question?

Yep, it's 20202.

jessie-ross:
I think we can fix this simply by adding a "Also change default strategy" switch under the Strategy selector

yanirs:
What do you mean? Add a different default strategy setting to the experiment?

PR #557, opened 2021-06-23
Add significance data to AggregateRecommendation

jessie-ross:
**This PR adds practical and statistical significance data to AggregateRecommendation:**
- Needed for #524 [using my version of the spreadsheet](https://docs.google.com/spreadsheets/d/1CiumVCKJD0gl4tuTiOvljYHFX12M7yiIwBKitIkyAO0/edit?usp=sharing)  (see also @aaronyan comment to @yanirs in Aarons spreadsheet), see the extra part of the table at the end and how the two columns uniquely identify a message.
- Wanted central and testable location to put this data.
- Since most of the work here is added testing and the same testing is needed for moving the recommender code here I thought I would move it here anyway.

Future PRs:
- Add the UI for the recommender text
- https://github.com/Automattic/abacus/issues/559
- Remove Recommender based code 646-gh-Automattic/experimentation-platform

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/__snapshots__/ExperimentDebug.test.tsx.snap
+++ b/src/components/experiments/single-view/results/__snapshots__/ExperimentDebug.test.tsx.snap
@@ -459,8 +459,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
@@ -503,7 +502,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Inconclusive
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
@@ -575,8 +574,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
@@ -619,8 +617,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
@@ -978,8 +975,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
@@ -1022,8 +1018,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
@@ -1066,8 +1061,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
@@ -1110,8 +1104,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
@@ -1154,8 +1147,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
                     </td>
                     <td
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
--- a/src/lib/analyses.ts
+++ b/src/lib/analyses.ts
@@ -667,6 +667,9 @@ export function getExperimentHealthIndicators(experiment: ExperimentFull): Healt
   }))
 }
 
-export function ratioToPercentDifference(ratio: number): number {
-  return (ratio - 1) * 100
+/**
+ * Takes an B/A ratio and returns the difference ratio: (B-A)/A
+ */
+export function ratioToDifferenceRatio(ratio: number): number {
+  return ratio - 1
 }
--- a/src/lib/analyses.ts
+++ b/src/lib/analyses.ts
@@ -31,7 +31,7 @@ export const RecommendationWarningToHuman = {
   [RecommendationWarning.WideCi]: 'The CI is too wide in comparison to the ROPE. Collect more data to be safer.',
 }
 
-export enum PracticalStatisticalDifferenceStatus {
+export enum PracticalSignificanceStatus {
   /**
    * The CI is entirely outside of ROPE (the region of practical equivalence, set by experimenters with minDifference):
    * If the CI is entirely outside of ROPE then it is above the threshold minumum practical difference set by the experimenter (minDifference)
@@ -54,8 +54,8 @@ export enum PracticalStatisticalDifferenceStatus {
 interface DiffCredibleIntervalStats {
   /**
    * A difference is a difference only if it makes a difference:
-   * - A diff CI is practically statistically different if it exists entirely above (or below) an experimenter set "Minimum difference".
-   * - A diff CI has "uncertian" practical statistical difference if it might still be practically statistically different - part of it exists above or below the set "Minimum difference".
+   * - A diff CI is practically significant if it exists entirely above (or below) an experimenter set "Minimum difference".
+   * - A diff CI has "uncertian" practical significance if it might still be practically significant - part of it exists above or below the set "Minimum difference".
    *
    * Practical significance is part of the approach described in
    * https://yanirseroussi.com/2016/06/19/making-bayesian-ab-testing-more-accessible/, which is based on
@@ -70,13 +70,13 @@ interface DiffCredibleIntervalStats {
    * * Bayesian estimation supersedes the t test: http://psy-ed.wdfiles.com/local--files/start/Kruschke2012.pdf
    * * Rejecting or accepting parameter values in Bayesian estimation: https://osf.io/s5vdy/download/?format=pdf
    */
-  practicalStatisticalDifference: PracticalStatisticalDifferenceStatus
+  practicallySignificant: PracticalSignificanceStatus
   /**
    * Old fashioned statistical significance: if a CI doesn't contain zero.
    *
    * Mostly used for adding more subtlety to recommendations and debugging.
    */
-  statisticalDifference: boolean
+  statisticallySignificant: boolean
   /**
    * Whether or not a CI is entirely positive. This doesn't necessarily mean better or worse at this point.
    */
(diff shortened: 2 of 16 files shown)


src/lib/analyses.ts
```diff
@@ -23,6 +23,15 @@ export const RecommendationWarningToHuman = {
   [RecommendationWarning.WideCi]: 'The CI is too wide in comparison to the ROPE. Collect more data to be safer.',
 }
 
+export enum PracticalSignificanceStatus {
```
yanirs, line 26: Can you please explain the need for this? The recommendation reason already maps nicely to the explanations we discussed.

src/lib/analyses.ts
```diff
@@ -57,28 +71,67 @@ export function getAggregateRecommendation(
...
+      statisticallySignificant,
+      practicallySignificant,
     }
   }
 
+  // Got so close before I realised higherIsBetter is missing:
+  // const defaultVariation = experiment.variations.find(variation => variation.isDefault) as Variation
+  // const nonDefaultVariation = experiment.variations.find(variation => !variation.isDefault) as Variation
+  // (Doesn't matter which end of the interval we choose as practical signficance implies statistical significance.)
+  // const chosenVariation = analysis.metricEstimates.diff.bottom > 0 === metric.higherIsBetter ? nonDefaultVariation : defaultVariation
```
yanirs, line 122: Why would this be needed? For the purpose of the explanations we discussed, we only care about whether the value is different from zero. Statistical significance shouldn't be used for the recommendation's bottom line.

jessie-ross: It isn't, practical significance is used here, we are just seeing which side of 0 we are on.

yanirs: Then why do we need `higherIsBetter`?

src/lib/analyses.ts
jessie-ross: Practical and Statistical significance are what we care about when generating the recommendation/explanation, it is also the main thing we are communicating with these added explanations. It will also be used in other areas of the UI such as in highlighting significant values.

See the table at the end of https://docs.google.com/spreadsheets/d/1CiumVCKJD0gl4tuTiOvljYHFX12M7yiIwBKitIkyAO0/edit#gid=0

- Statistical maps to explanation 1
- Practical maps to explanation 2
- We can easily generate the overview sentence.

Being able to branch the code on practical and statistical significance levels makes the control flow simpler and the code much easier to understand - e.g. a result is inconclusive because a value might be practically significant vs a result is inconclusive because CI overlaps with ROPE. We want to move from the implementation detail of CI relation to ROPE and 0 to the higher level concept of statistical and practical significance.

The alternative is having a massive switch statement on the 'reason' codes, without the ability to factor together similar explanations, at least not without adding more complexity. I am not a fan of the reason codes, by conflating the two dimensions of significance together we are losing clarity and cleanness of our code.

Further, for this slight reframe we get the ability to remove `recommender.py` for free - 400 lines of code complicating the ETL.

yanirs: Moving the recommendation logic to Abacus makes sense &ndash; I'm on board with that.

I think that the problem is with the idea of "maybe" being a valid answer to "is this practically significant?" It's not really saying much, and it means that the `practicallySignificant?` field has four possible values (including undefined). It'd be better to keep it as a boolean with:

* `true`: The difference is practically significant with high certainty.
* `false`: The difference isn't practically significant with high certainty.
* `undefined`: Uncertainty is too high to say anything about practical significance.

This would also simplify the code, as the focus will be on setting the decision based on the analysis rather than repeating the full analysis structure creation.

src/lib/analyses.ts
jessie-ross: We need `higherIsBetter` to know which variation is better - to get the `chosenVariationId`.

We need to know whether the diff is positive or negative _and then_ whether positive/negative is a good thing.

src/lib/analyses.ts
jessie-ross: I am somewhat against the boolean/undefined values - it makes it very easy to make errors as we would be using them in a rather unorthodox way, especially with both false and undefined being falsey. 

Would using 'Uncertain' instead of 'Maybe' sweeten the deal?

src/lib/analyses.ts
jessie-ross: In case it isn't clear, these extra 3 lines of code lets us replace `recommender.py`.

src/lib/analyses.ts
jessie-ross: > it means that the practicallySignificant? field has four possible values (including undefined, which is sort of like "maybe").

It is only undefined for missingAnalysisData, for which we show nothing.

src/lib/analyses.ts
yanirs: Yeah, that's the bit of context I missed &ndash; I thought this PR was only about supporting the explanations.

src/lib/analyses.ts
yanirs: > Would using 'Uncertain' instead of 'Maybe' sweeten the deal?

Yeah, that's much better. 

> It is only undefined for missingAnalysisData, for which we show nothing.

Also for "manual analysis required". :slightly_smiling_face: 

> Further, for this slight reframe we get the ability to remove `recommender.py` for free - 400 lines of code complicating the ETL.

I just had a look and `recommender.py` is actually 160 lines, with about half of them being documentation. While I'm on board with moving the logic here, we shouldn't lose the documentation, as this stuff isn't trivial. I know that this is still a WIP, but can you please use the opportunity to fully document the recommendation code?

jessie-ross: > 400 lines of code 

I'm including the test code.

src/lib/schemas.ts
```diff
@@ -108,14 +108,14 @@ export const metricBareSchema = yup
     name: nameSchema.defined(),
     description: yup.string().defined(),
     parameterType: yup.string().oneOf(Object.values(MetricParameterType)).defined(),
+    higherIsBetter: yup.boolean().defined(),
```
yanirs, line 111: Not sure if that's already the plan since this is still WIP, but it'd be good to have a separate PR with the retirement of `MetricBare`.

jessie-ross: Yep that's the plan, just doing the minimal here so I can focus on this change.

src/components/experiments/single-view/results/__snapshots__/ExperimentDebug.test.tsx.snap
```diff
@@ -459,8 +459,7 @@ exports[`renders the full tables with some analyses in debug mode 2`] = `
                       class="MuiTableCell-root MuiTableCell-body MuiTableCell-alignLeft"
                       style="box-sizing: border-box;"
                     >
-                      Deploy 
-                      test
+                      Deploy either variation
```
yanirs, line 462: Why have these snapshot changed?

yanirs (changes requested):
I haven't fully looked through the tests, but here are my top-level comments so far:
- This is a pretty big PR. I'm guessing there may be an issue with the dummy data, but changing it to not affect existing snapshot tests would make reviewing easier.
- I think that `getAggregateRecommendation()` should be broken up and focused on a single recommendation to make it easier to work with and maintain. It may have made sense when we had less logic in Abacus, but doing so much in the function opens the door to many subtle bugs and plenty of confusion.
- I think it'd be better to revert the name changes from https://github.com/Automattic/abacus/commit/ab92d6586e7f613b51b0755b6b5df36383645b54 
- Documentation of exported functionality should be detailed enough for people who don't come with much context.

src/components/experiments/single-view/results/__snapshots__/ExperimentDebug.test.tsx.snap
jessie-ross: Yeah we used the recommendation object before and now we use metricEstimates - they aren't consistent in the fixtures.

src/lib/analyses.ts
```diff
@@ -23,6 +31,117 @@ export const RecommendationWarningToHuman = {
...
+}
+
+/**
+ * Takes DiffCredibleIntervalStats from each strategy and returns whether there is a conflict between them -
+ * when there are multiple practically significant CIs pointing in different directions
+ */
+export function isDiffCredibleIntervalConflict(allDiffCredibleIntervalStats: (DiffCredibleIntervalStats | null)[]) {
+  return (
+    [
+      ...new Set(
```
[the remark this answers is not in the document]
jessie-ross: Good point, this comes from the original implementation.

src/lib/analyses.ts
```diff
@@ -43,42 +164,70 @@ export interface AggregateRecommendation {
  * @param defaultStrategy Default strategy in the context of an aggregateRecommendation..
  */
 export function getAggregateRecommendation(
```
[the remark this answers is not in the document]
jessie-ross: Sure I can split it further, but the functionality here hasn't inflated at all, it doesn't do anything it didn't do before.

But I would want to split it into: 
- `getMetricAssignmentRecommendation(experiment, metric, analysis)`
- `getAggregateMetricAssignmentRecommendation(metricAssignmentRecommendations)`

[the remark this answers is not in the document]
yanirs: Yeah, this split would work too. :slightly_smiling_face:

src/components/experiments/single-view/results/__snapshots__/ExperimentDebug.test.tsx.snap
yanirs: Yeah, I figured this is what happened. It'd be nice to make them consistent if it's not too much effort.

src/lib/analyses.ts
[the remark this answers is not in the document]
jessie-ross: This has ended up being quite a large change, especially with testing changes, should I split it into another PR on top?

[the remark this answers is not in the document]
yanirs: Yeah, if you think it'd help. I don't mind looking at it as a separate commit in this PR if it's too much effort to split.

src/lib/analyses.ts
```diff
@@ -32,8 +132,17 @@ export enum AggregateRecommendationDecision {
 }
 
 export interface AggregateRecommendation {
```
yanirs, line 134: Should we just call them `Recommendation` and `RecommendationDecision` now? They're not really aggregates.

jessie-ross: Yeah, I would like to do so, but I would like to do so in another PR as this one is ballooning.

Just working on the documentation now.

src/lib/analyses.ts
```diff
@@ -23,6 +31,117 @@ export const RecommendationWarningToHuman = {
   [RecommendationWarning.WideCi]: 'The CI is too wide in comparison to the ROPE. Collect more data to be safer.',
 }
 
+export enum PracticalStatisticalDifferenceStatus {
```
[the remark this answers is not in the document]
jessie-ross: There are probably still typos but I have added a top-level comment 😄

src/components/experiments/single-view/results/__snapshots__/ExperimentDebug.test.tsx.snap
jessie-ross: I have marked `recommendation` as deprecated, so I don't think there is a need to make them consistent now.
Eventually we can actually remove all references to recommendation while still letting the API add recommendation in.

jessie-ross: It is currently only used in the Debug view

yanirs: Yeah, we can probably retire the Debug tab now that we have the strategy selector and observed data in the main view, especially once #554 is fixed.

src/lib/analyses.ts
```diff
@@ -32,54 +150,89 @@ export enum AggregateRecommendationDecision {
...
-    .map((analysis) => analysis.recommendation)
-    .map((recommendation) => recommendation?.chosenVariationId)
-    .filter((x) => x)
-  const recommendationConflict = [...new Set(recommendationChosenVariationIds)].length > 1
-  if (recommendationConflict) {
+  const metricAssignment = experiment.metricAssignments.find(
+    (metricAssignment) => metricAssignment.metricAssignmentId === analysis.metricAssignmentId,
+  )
+  const diffCredibleIntervalStats =
+    analysis && metricAssignment && getDiffCredibleIntervalStats(analysis, metricAssignment)
```
yanirs, line 180: There shouldn't be a need to check the `analysis` again, and we should be able to assume that the metric assignment exists.

Also, `getDiffCredibleIntervalStats()` can handle a null `analysis` and line 177 assumes a non-null/undefined `analysis`.

```suggestion
  const metricAssignment = experiment.metricAssignments.find(
    (metricAssignment) => metricAssignment.metricAssignmentId === analysis.metricAssignmentId,
  ) as MetricAssignment
  const diffCredibleIntervalStats = getDiffCredibleIntervalStats(analysis, metricAssignment)
```

src/lib/analyses.ts
```diff
@@ -32,54 +150,89 @@ export enum AggregateRecommendationDecision {
...
-      decision: AggregateRecommendationDecision.MissingAnalysis,
-    }
+  const { practicallySignificant, statisticallySignificant, isPositive } = diffCredibleIntervalStats
+  const decision = PracticalSignificanceStatusToDecision[practicallySignificant]
+  const defaultVariation = experiment.variations.find((variation) => variation.isDefault) as Variation
+  const nonDefaultVariation = experiment.variations.find((variation) => !variation.isDefault) as Variation
+  let chosenVariationId = undefined
+  if (decision === AggregateRecommendationDecision.DeployChosenVariation) {
+    chosenVariationId =
+      isPositive === metric.higherIsBetter ? nonDefaultVariation.variationId : defaultVariation.variationId
```
yanirs, line 196: Just a couple of other ways. Not fussed about this. :slightly_smiling_face: 

No need to go through the variations if we don't use the result:
```suggestion
  let chosenVariationId = undefined
  if (decision === AggregateRecommendationDecision.DeployChosenVariation) {
    const defaultVariation = experiment.variations.find((variation) => variation.isDefault) as Variation
    const nonDefaultVariation = experiment.variations.find((variation) => !variation.isDefault) as Variation
    chosenVariationId =
      isPositive === metric.higherIsBetter ? nonDefaultVariation.variationId : defaultVariation.variationId
```

And we don't need to iterate twice:
```suggestion
  let chosenVariationId = undefined
  if (decision === AggregateRecommendationDecision.DeployChosenVariation) {
    const chooseDefaultVariation = isPositive !== metric.higherIsBetter
    const chosenVariation = experiment.variations.find((variation) => variation.isDefault === chooseDefaultVariation) as Variation
    chosenVariationId = chosenVariation.variationId
```

src/lib/analyses.ts
```diff
@@ -566,6 +719,7 @@ export function getExperimentAnalysesHealthIndicators(
   const diffCiWidth = Math.abs(analysis.metricEstimates.diff.top - analysis.metricEstimates.diff.bottom)
   const ropeWidth = metricAssignment.minDifference * 2
   const indicatorDefinitions = [
+    // See Kruschke's Precision as a goal for data collection: https://www.youtube.com/playlist?list=PL_mlm7M63Y7j641Y7QJG3TfSxeZMGOsQ4
```
yanirs, line 722: Can drop this now that it's in the top-level comment.

```suggestion
```

src/lib/analyses.ts
```diff
@@ -23,6 +31,116 @@ export const RecommendationWarningToHuman = {
...
+ * for recommendations but is included for debugging and better understanding
+ * results.
+ *
+ * ## Practical Significance
+ *
+ * A difference CI is practically significant if it doesn't contain any part of
+ * the ROPE. Practical significance is what we use for recommendations.
+ *
+ * Benefits of practical significance:
+ * - More accurate burden of proof - tailored to the Experimenters specific decision they are making.
```
yanirs, line 65:
```suggestion
 * - More accurate burden of proof - tailored to the experimenters' decisions around specific metrics.
```

yanirs (approved):
:ship: with just a tiny bit of inline nitpicking :smile: 

The overall documentation can probably be improved a bit, but it's not critical and I may play with it after the stack is merged.

src/lib/analyses.ts
jessie-ross: It isn't top level of this file once the recommendation code is moved out.

jessie-ross:
👍 I am going to merge this one and address these with a PR on top of the stack

src/lib/analyses.ts
jessie-ross: I will remove the analysis, but there I think it is worth keeping the metricAssignment check - there have been backend/etl bugs before causing the front-end to crash.

src/lib/analyses.ts
jessie-ross: I would actually like to rename `chosenVariationId` to be something like `preferredVariationId` or `betterVariationId` and include it when statistically significant, this would allow us to have better health monitoring. We might also want to trigger manual analysis required for cases where there are conflicts of statistical significance, this would be a step in-between what we currently have of conflicting practical significance and what we had before - conflicting practical significance including not practical significant.

PR #558, opened 2021-06-24
Fix unit consistency

jessie-ross:
**This PR fixes some unit consistency issues mentioned in p1624506676009300-slack-G01BQ5WMR8Q:**
- Use correct units for non-different values
- Use MetricValue in Observations table
- Don't display signs for values that aren't signed
- Improve currency display

I am not too happy with MetricValue as an abstraction, might have a think about it.

## How has this been tested?

### DIFF
--- a/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
+++ b/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
@@ -325,6 +325,7 @@ export default function MetricAssignmentResults({
                       <MetricValueInterval
                         intervalName={'the absolute change between variations'}
                         metricParameterType={metric.parameterType}
+                        isDifference={true}
                         bottomValue={latestEstimates.diff.bottom}
                         topValue={latestEstimates.diff.top}
                       />
--- a/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
+++ b/src/components/experiments/single-view/results/MetricAssignmentResults.tsx
@@ -316,6 +316,7 @@ export default function MetricAssignmentResults({
                       metricParameterType={metric.parameterType}
                       bottomValue={latestEstimates[`variation_${variation.variationId}`].bottom}
                       topValue={latestEstimates[`variation_${variation.variationId}`].top}
+                      displayPositiveSign={false}
                     />
                   </TableCell>
                   <TableCell className={classes.monospace} align='right'>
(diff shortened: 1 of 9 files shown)


src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -316,6 +316,7 @@ export default function MetricAssignmentResults({
                       metricParameterType={metric.parameterType}
                       bottomValue={latestEstimates[`variation_${variation.variationId}`].bottom}
                       topValue={latestEstimates[`variation_${variation.variationId}`].top}
+                      displayPositiveSign={false}
```
yanirs, line 319: Why remove the plus sign here? Is it to differentiate it from the difference and lift intervals?

src/components/general/MetricValue.tsx
```diff
@@ -25,18 +25,35 @@ function DashedTooltip(props: Parameters<typeof Tooltip>[0]) {
...
-  string,
-  { unit: React.ReactNode; prefix: React.ReactNode; postfix: React.ReactNode; transform: (v: number) => number }
-> = {
+export const metricValueFormatData: Record<string, MetricValueFormat> = {
+  count: {
+    unit: '',
+    prefix: '',
+    postfix: '',
+    transform: identity,
+    formatter: (n) => n.toLocaleString(undefined),
```
yanirs, line 49: When running locally, ESLint is unhappy about the lack of type declarations here.

```suggestion
    formatter: (n: number): string => n.toLocaleString(undefined),
```

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
yanirs: Looking around, this seems to be the only case where `displayPositiveSign` is `false` for `MetricValueInterval`. Maybe it's worth always making it `true` to simplify things?

yanirs: Ah, this may be redundant since both revenue and conversion rate can't be negative. I don't mind keeping the `false` here. :slightly_smiling_face:

yanirs (approved):
Should fix the lint issue, but other than that LGTM. :ship:

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: Yep they are always positive so the plus signs are just noise - and it does help differentiate them from diff values.

src/components/general/MetricValue.tsx
jessie-ross: interesting

yanirs: I took the liberty of committing it since I tested it locally and it's only outstanding comment. :slightly_smiling_face:

jessie-ross: Thanks!

PR #565, opened 2021-07-05
Fix client validation error messages

jessie-ross:
**This PR makes validation errors say "This" instead of the qualified name of the input.** 
- Fixed by moving the messages to the start of schemas so they can
be applied.

**Before:**
Screen Shot 2021-07-05 at 6 17 28 pm

**After:**
Screen Shot 2021-07-05 at 6 15 49 pm

## How has this been tested?


yanirs (approved):
:ship:

jessie-ross:
Part of #316

PR #567, opened 2021-07-05
Fix missing validation errors

### LINKED ISSUE #423: Improve date field input and error messages based on feedback from Don
In pbmo2S-j2-p2#comment-1377, @donlair said:

> * The test date text field crashed hard every time I tried to enter a date, even if properly formatted. I eventually typed the date in a separate text editor and pasted it in.
> * I had to look at an existing test to see what date format to use.
> * Some input error explanations would be helpful. For example, it took me a minute to realize I needed to manually type a value in the Minimum Difference field, even though there appears to be a default value.

We should investigate the crash with Don and improve the way the wizard works to make it easier to use.

jessie-ross:
**This PR fixes missing validation errors:**
- User Autocomplete:
  
   Screen Shot 2021-07-05 at 8 00 03 pm

- Platform

Screen Shot 2021-07-05 at 9 27 23 pm

- No metric assignments
Screen Shot 2021-07-05 at 7 45 37 pm

- Attribution Window
Screen Shot 2021-07-05 at 9 29 12 pm

Resolves #423 

## How has this been tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-
-#   Abacus: Automattic's Experimentation Platform UI
+# Abacus: Automattic's Experimentation Platform UI
 
 [![CircleCI](https://circleci.com/gh/Automattic/abacus.svg?style=svg)](https://circleci.com/gh/Automattic/abacus)
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-# Abacus: Automattic's Experimentation Platform UI
+
+#   Abacus: Automattic's Experimentation Platform UI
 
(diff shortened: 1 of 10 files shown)


jessie-ross:
@yanirs, do I have to get full coverage here?

To me the lines without coverage are incredibly low priority.

yanirs:
> @yanirs, do I have to get full coverage here?
> 
> To me the lines without coverage are incredibly low priority.

Yeah, it looks like it's pretty much all the new changes. It'd be good to have them covered rather than rely on manual testing as we change things in the future.

yanirs:
> Attribution Window

As noted on #568, I think it'd be good to have the error text in this case.

yanirs (approved):
:ship:

PR #568, opened 2021-07-05
Fix up the wizard metric assignments

jessie-ross:
**This PR makes misc life improvements for metric assignment:**
- Makes `changeExpected` true by default.
- Removes helper text from `minDifference` errors that were throwing the design out.

   Screen Shot 2021-07-05 at 10 21 51 pm
- Makes minDiff wide enough to see plenty decimal places.

   Screen Shot 2021-07-05 at 10 23 17 pm
- Also improves general error messages

## How has this been tested?

### DIFF
--- a/src/components/experiments/wizard/Metrics.tsx
+++ b/src/components/experiments/wizard/Metrics.tsx
@@ -126,7 +126,7 @@ const createMetricAssignment = (metric: Metric) => {
   return {
     metricId: metric.metricId,
     attributionWindowSeconds: '',
-    isPrimary: false,
+    isPrimary: true,
     changeExpected: false,
     minDifference: '',
   }
--- a/src/components/experiments/wizard/Metrics.tsx
+++ b/src/components/experiments/wizard/Metrics.tsx
@@ -49,6 +49,11 @@ const useStyles = makeStyles((theme: Theme) =>
     monospaced: {
       fontFamily: theme.custom.fonts.monospace,
     },
+    metricNameCell: {
+      maxWidth: '24rem',
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+    },
     metricName: {
       fontFamily: theme.custom.fonts.monospace,
       fontWeight: theme.custom.fontWeights.monospaceBold,
@@ -337,7 +342,7 @@ const Metrics = ({
                       <TableCell>Metric</TableCell>
                       <TableCell>Attribution Window</TableCell>
                       <TableCell>Change Expected?</TableCell>
-                      <TableCell>Minimum Difference</TableCell>
+                      <TableCell>Minimum Practical Difference</TableCell>
                       <TableCell />
                     </TableRow>
                   </TableHead>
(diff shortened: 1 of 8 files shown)


src/components/experiments/wizard/Metrics.tsx
```diff
@@ -126,7 +126,7 @@ const createMetricAssignment = (metric: Metric) => {
   return {
     metricId: metric.metricId,
     attributionWindowSeconds: '',
-    isPrimary: false,
+    isPrimary: true,
     changeExpected: false,
```
yanirs, line 130: Judging by the PR description ("Makes `changeExpected` true by default"), did you mean to make this change instead?

```suggestion
    isPrimary: false,
    changeExpected: true,
```

src/components/general/MetricDifferenceField.tsx
```diff
@@ -14,11 +15,18 @@ const ConversionMetricTextField = formikFieldTransformer(
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
+    root: {
```
yanirs, line 18: While you're here, can you also remove the number spinners on Firefox? It should be a matter of adding `-moz-appearance: textfield` to the styling, unless Formik has an option for it.

image

src/components/general/MetricDifferenceField.tsx
```diff
@@ -76,6 +85,7 @@ export default function MetricDifferenceField(props: {
         InputProps={{
           startAdornment: <InputAdornment position='start'>USD</InputAdornment>,
```
yanirs, line 86: How about we change this to an `endAdornment`, which would match the way metric assignments are now displayed in most places?

src/components/general/MetricDifferenceField.tsx
```diff
@@ -14,11 +15,18 @@ const ConversionMetricTextField = formikFieldTransformer(
...
+      minWidth: '8rem',
+    },
     tooltipped: {
       borderBottomWidth: 1,
       borderBottomStyle: 'dashed',
       borderBottomColor: theme.palette.grey[500],
     },
+    // We hide the helper text shown in errors as the error is obvious and there is limited space as is.
+    helperText: {
+      display: 'none',
```
yanirs, line 28: The error isn't that obvious. For example, not changing anything from the default and moving forward in the wizard, we now just get a red rectangle around the input with no indication that zero is an illegal value. I think it'd be better to keep the helper text.

On a related note, even if the minimum difference is set to a legal value, the assignment may be invalid due to a missing attribution window. There is no visual indicator or a useful error message in this case.

src/components/general/MetricDifferenceField.tsx
yanirs: Also, on Firefox with a long metric name, there is now a horizontal scrollbar that doesn't in production. It'd be good to avoid that by wrapping long names.

src/components/experiments/wizard/Metrics.tsx
jessie-ross: 🤦

jessie-ross:
I've fixed your suggests and also renamed "Minimum Difference" to "Minimum Practical Difference"

yanirs (approved):
:tada:

PR #569, opened 2021-07-08
Show partial results

### LINKED ISSUE #554: Show partial results without a primary metric analysis

jessie-ross:
**This PR alters the analysis logic to show results when there are any analyses at all.**

Also an updated missing analyses message:

Screen Shot 2021-07-08 at 4 42 26 pm

Resolves #554 

## How has this been tested?


src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -242,7 +242,9 @@ export default function ActualExperimentResults({
       ),
     strategy,
   )
-  const hasAnalyses = primaryMetricRecommendation.decision !== Recommendations.Decision.MissingAnalysis
+  const hasAnalyses = allMetricAssignmentAnalysesData.some(
```
yanirs, line 245: An inline comment would be useful here as it's not that intuitive why we're doing things this way.

yanirs (approved):

PR #575, opened 2021-07-26
Make experiment wizard more responsive

### LINKED ISSUE #176: Verify that all the pages are mobile friendly
As discussed in https://github.com/Automattic/abacus/pull/169#issuecomment-651170615, we're going for a "desktop first, mobile second" approach, i.e., a great experience on desktop/tablet and passable usability on mobile. This means that the pages should render nicely on mobile, but not necessarily be optimised for narrow screens. Right now, not all pages are very usable, e.g., some tables require horizontal scrolling. We should do a pass through the main views and tweak things as needed.

jessie-ross:
**This PR makes the experiment wizard more responsive.**
- Makes the navigation horizontal.
- Tweaks the width settings on form-parts.

Best experienced via Vercel

Resolves #176 

## How has this been tested?


aaronyan (commented):
This is looking great! On narrow devices, the horizontal navigation protrudes out the side. Maybe after a certain width, we only show the numbers?

Screen Shot 2021-07-26 at 3 21 10 PM

And then on the basic info step, the card protrudes to the right. Maybe stack the start and end date?

Screen Shot 2021-07-26 at 3 20 47 PM

jessie-ross:
> This is looking great! Would it be possible to include these two adjustments?

I've included these changes, they are for mobile devices however so I haven't been focusing on that size.

aaronyan (approved):
Awesome, thank you for making those changes. This looks and feels great!

PR #577, opened 2021-07-27
Fix form enter behaviour

jessie-ross:
**This PR improves current form enter behaviour.**

Initially the form submitted on enter, then we changed pressing enter to go to the next stage of the form.

In order to improve accessibility this PR disables submission-on-enter and enter-to-go-to-the-next-stage, allowing enter to be pressed on auto-select and tab-focused elements.

## How has this been tested?


aaronyan (approved):
LGTM! Pretty cool accessibility improvement!

Tried out this PR and enter while hovering over "Next" will now only move on to the next step. Previously, hitting enter anywhere will progress to the next section.

PR #579, opened 2021-07-27
Note that exposure events must be validated

### LINKED ISSUE #462: Add text that only validated events are accepted exposure events
Shall we consider having a line of text (or clickable info icon) describing that **only validated events are autocompleted** for exposure events? Having this information available in the experiment creation wizard could help answer some of experimenters questions when they don't find what they're looking for.

I recently forgot this rule and re-learned it by diving into our code. We also recently had an experimenter ask us why their exposure event was not autocompleting in p1612774650033300-slack-C7HH3V5AS.

jessie-ross:
**Note that exposure events must be validated:**

Screen Shot 2021-07-27 at 3 53 32 pm

Resolves  #462

## How has this been tested?

### DIFF
--- a/src/components/experiments/wizard/Metrics.tsx
+++ b/src/components/experiments/wizard/Metrics.tsx
@@ -573,6 +573,8 @@ const Metrics = ({
         >
           What is an Exposure Event? And when do I need it?
         </Link>
+        <br />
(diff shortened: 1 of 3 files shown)


yanirs (approved):
The snapshots need fixing but LGTM otherwise.

PR #582, opened 2021-07-30
Fix name overflows

### LINKED ISSUE #578: Fix long name UX
> Sort of related: There are several places with view overflows even on wide displays. I tested by adding an experiment and metric named `explat_test_very_long_name_yes_this_name_is_very_long_and_it_still_fits_within_our_limits_will_it_look_good_though_who_knows`. With this:
> 
> * Experiment List looks good
> * Experiment Overview has several overflows
> * Metric List overflows
> 
> Can you please look into it? We can also split this into a separate issue since it's a general overflow problem.
> 

_Originally posted by @yanirs in https://github.com/Automattic/abacus/issues/176#issuecomment-887114230_

jessie-ross:
This PR fixes some name overflows seen in the wild:
- Exposure event names in overview
- Metric names in overview

Resolves #578 

## How has this been tested?


yanirs:
> Resolves #578

This is a good start, but it's not enough to resolve the issue. I think we should also fix the experiment name in the heading of the overview page, as it overflows even with moderately long names on narrow displays (e.g., an iPad).

It'd also be good to fix `/metrics`, but I don't mind leaving it out for now since long names are less likely there and it's less useful to most experimenters.

jessie-ross:
>  This is a good start, but it's not enough to resolve the issue. I think we should also fix the experiment name in the heading of the overview page, as it overflows even with moderately long names on narrow displays (e.g., an iPad).

I've fixed the experiment name now.

> It'd also be good to fix /metrics, but I don't mind leaving it out for now since long names are less likely there and it's less useful to most experimenters.

This one is a bit annoying due to material table, might leave it for now and switch it over to ag-grid later.

yanirs (approved):
:tada:

PR #584, opened 2021-07-30
Experiments table: Store last search string in memory

### LINKED ISSUE #545: Retain experiment search string in browser history
Moving 570-gh-Automattic/experimentation-platform here. :slightly_smiling_face:

jessie-ross:
**This PR stores the last search string in memory.**

Resolves #545 

## How has this been tested?


yanirs:
My reading of the issue is different: "retain experiment search string in browser history" means changing the URL to me, which would allow experimenters to share and bookmark specific searches, e.g., https://abacus-git-add-remember-last-search-a8c.vercel.app/experiments?search=yanirseroussi would show my experiments.

Wouldn't this be fairly straightforward to do with a bit of history and query string manipulation?

PR #586, opened 2021-07-30
Fix metric difference field alignment

### LINKED ISSUE #452: Align minimum difference fields in experiment wizard
When helping experimenters edit their experiment, I stumbled on a few potentially easy, quality of life, upgrades to the experiment wizard metric assignment:
- Make input fields same width for both conversion and revenue metrics
- Move units to right of input fields
- Right align input fields (the numbers)
- Left align the units "pp" and "USD"

Here's a before and after (expected):

## Before

Screen Shot 2021-01-19 at 9 59 56 AM

## After

Screen Shot 2021-01-19 at 9 59 56 AM copy

## Action items
- [ ] Make the above changes or discuss/ abandon

jessie-ross:
**This PR right aligns metric difference input and makes the adornments fixed width so they align on the wizard metrics form-part:**

Screen Shot 2021-07-30 at 6 20 04 pm

Resolves #452 

## How has this been tested?


aaronyan (approved):
Yay, this looks great! LGTM.

I think the snapshots need to be updated? And then we're good to go!

PR #589, opened 2021-08-04
Add bookmarklet support for the test user parameter

yanirs:
As part of the work on 649-gh-Automattic/experimentation-platform, prompt for and pass in the `test_username` parameter in bookmarklet calls and change the messaging accordingly.

## How has this been tested?

- Additional manual testing instructions (please specify): See backend diff (D65112-code)


src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -49,33 +49,50 @@ function assignmentHref(variationName: string, experimentName: string) {
   return `javascript:(async () => {
     const token = JSON.parse(localStorage.getItem('experiments_auth_info'));
     const anonId = decodeURIComponent(document.cookie.match('(^|;)\\\\s*tk_ai\\\\s*=\\\\s*([^;]+)')?.pop() || '');
+    let testUsername = '';
     const headers = {'Content-Type': 'application/json'};
     if (token && token.accessToken) {
        headers.Authorization = 'Bearer ' + token['accessToken'];
+       testUsername = prompt(
```
jessie-ross, line 56: Can we rename testUsername to something else?

src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -49,33 +49,50 @@ function assignmentHref(variationName: string, experimentName: string) {
...
+         'This will assign a logged-in user to the ${variationName} variation of ${experimentName}. ' +
+         'You have two options:\\n' +
+         '(1) Enter a test username to assign (only works for staging experiments).\\n' +
```
jessie-ross, line 59:
```suggestion
         '(1) Enter a username to assign (only works for staging experiments).\\n' +
```

src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -49,33 +49,50 @@ function assignmentHref(variationName: string, experimentName: string) {
...
         break;
       case 'user_not_assignable':
-        alert('You must be proxied or sandboxed to use this bookmark');
+        alert('You must be proxied or sandboxed to use this bookmark.');
         break;
       case 'invalid_ids':
-        alert('To assign yourself, you must run this from Abacus.\\n\\nTo assign the current anonymous user, verify ' +
-              'that Tracks is not blocked and run in an environment where the tk_ai cookie is set.');
+        alert(
+          'To assign yourself or a test user, you must run this from Abacus and specify an existing user.\\n\\n' +
```
jessie-ross, line 89:
```suggestion
          'To assign yourself or a specified user, you must run this from Abacus and specify an existing user.\\n\\n' +
```

jessie-ross (changes requested):
Left some minor code/clarity comments.

Looking at it now, manual assignment is feeling a bit overloaded and magical.

I think what might help is manual assignment is expanding the tooltip or [opening up a modal instead](https://material-ui.com/components/modal/) which explains the new manual assignment, and providing separate links for different types of assignment - and within those different links adding more detailed error checking.

At the very least I think we need to expand on the instructions in the tooltip.

src/components/experiments/single-view/overview/VariationsTable.tsx
yanirs: I'm not too fussed. Do you have any suggestions?

src/components/experiments/single-view/overview/VariationsTable.tsx
yanirs: I was trying to emphasise that it shouldn't be a real user, but I don't mind going with this suggestion.

yanirs:
> Left some minor code/clarity comments.
> 
> Looking at it now, manual assignment is feeling a bit overloaded and magical.
> 
> I think what might help is manual assignment is expanding the tooltip or [opening up a modal instead](https://material-ui.com/components/modal/) which explains the new manual assignment, and providing separate links for different types of assignment - and within those different links adding more detailed error checking.
> 
> At the very least I think we need to expand on the instructions in the tooltip.

Thanks @jessie-ross! I expanded the tooltip instructions in https://github.com/Automattic/abacus/pull/589/commits/1389afce333e62cacab58d0e748d61ec54d389b9.

I agree that it is a bit too magical. I believe that the original thinking was to make it easy to switch variations while testing a specific page, but things are different now that we rely a lot on Abacus authentication. It may be better to completely remove the bookmarklet in favour of a dialogue where users can specify a named user or an anonymous user ID. We'd just have to make it a bit easier to grab the anon ID. What do you think? I can open an issue and we can address it separately.

src/components/experiments/single-view/overview/VariationsTable.tsx
jessie-ross: `username_override`?

jessie-ross:
> It may be better to completely remove the bookmarklet in favour of a dialogue where users can specify a named user or an anonymous user ID. We'd just have to make it a bit easier to grab the anon ID. What do you think? I can open an issue and we can address it separately.

Yeah! Sounds like a good idea that we can look into at a later time.

src/components/experiments/single-view/overview/VariationsTable.tsx
```diff
@@ -154,9 +172,13 @@ function VariationsTable({
...
-                          <li> Click the bookmark when you are on the page you want to manually assign. </li>
-                          <li> You will receive an alert on success. </li>
+                          <li>Drag the above link to your bookmarks or bookmark bar.</li>
+                          <li>
+                            To assign yourself or another user, click the bookmark when you are on any Abacus page. To
+                            assign an anonymous user, click the bookmark when you are logged out on any page with Tracks
+                            enabled (the tk_ai cookie should be set).
+                          </li>
+                          <li>You will receive an alert on success.</li>
                         </ol>
```
jessie-ross, line 182: Hmm, how about splitting this over two separate headers: 'Logged-in Assignment Instructions' and 'Logged-out Assignment Instructions'?

- I think we should also specify the need to be proxied/authorised.
- Also we could add "check that tracks is enabled and running" (a lot of us run ad-blockers).

yanirs: Good ideas!

yanirs:
> Yeah! Sounds like a good idea that we can look into at a later time.

Opened #590 and addressed the other comments in 1f992b35a4928d8091a01b6d839114193449421b.

jessie-ross (approved):
Screen Shot 2021-08-05 at 4 28 18 pm

Looking good!

PR #593, opened 2021-08-16
Implement A/B/n in wizard

jessie-ross:
**This PR implements and tests A/B/n in the experiment wizard.**

## How has this been tested?

See test experiment: `/experiments/20267-explat-test-scjr-2021-08-16-10-35/overview`


.gitignore
```diff
@@ -21,6 +21,7 @@ pids
 
 # Dependency directories
 /node_modules/
+/node_modules
```
jessie-ross, line 24: I am trying out a symlink for node_modules, hope you don't mind

yanirs: Sure. It'd be good to add a comment above this line to say that this covers the symlink scenario, as it's not that obvious.

src/components/experiments/wizard/Audience.tsx
```diff
@@ -253,38 +273,111 @@ const Audience = ({
...
-                })}
+                <FieldArray
+                  name={`experiment.variations`}
+                  render={(arrayHelpers) => {
+                    const onAddVariation = () => {
+                      arrayHelpers.push({
+                        name: `treatment_${variations.length}`,
+                        isDefault: false,
+                        allocatedPercentage: String(Math.floor(100 / (variations.length + 1))),
+                      })
```
yanirs, line 295: This gets pretty weird when variations are removed and when the sum of percentages is higher than 100. How about we leave the `name` and `allocatedPercentage` empty?

Peek 2021-08-16 15-58

src/components/experiments/wizard/Audience.tsx
```diff
@@ -253,38 +273,111 @@ const Audience = ({
...
+                        {variations.map((variation, index) => {
+                          return (
+                            // The key here needs to be changed for variable variations
+                            <TableRow key={index}>
+                              <TableCell>
+                                {variation.isDefault ? (
+                                  variation.name
+                                ) : (
+                                  <Field
+                                    // className={classes.variationAllocatedPercentage}
```
yanirs, line 313: Was this left in on purpose?

src/components/experiments/wizard/Audience.tsx
```diff
@@ -253,38 +273,111 @@ const Audience = ({
...
+                              </TableCell>
+                            </TableRow>
+                          )
+                        })}
+                        <TableRow>
+                          <TableCell colSpan={3}>
+                            <Alert severity='warning' className={classes.abnWarning}>
+                              <strong> Manual analysis only A/B/n </strong>
+                              <br />
+                              <p>Experiments with more than a single treatment variation is in an early alpha stage.</p>
```
yanirs, line 355:
```suggestion
                              <p>Experiments with more than a single treatment variation are in an early alpha stage.</p>
```

src/lib/schemas.ts
```diff
@@ -434,7 +434,26 @@ export const experimentFullNewSchema = experimentFullSchema.shape({
...
+    )
+    .test(
+      'max-total',
+      'The sum of allocated percentages must be less than or equal to 100',
+      (variations: VariationNew[]) =>
+        variations && variations.reduce((acc, variation) => acc + Number(variation.allocatedPercentage), 0) <= 100,
+    )
+    .test(
+      'unique-names',
+      'Variation names must be unique.',
```
yanirs, line 454: Consistency :slightly_smiling_face: 

```suggestion
      'Variation names must be unique',
```

src/components/experiments/wizard/Audience.tsx
```diff
@@ -253,38 +273,111 @@ const Audience = ({
             Define the percentages to include in the experiment. <br /> Use &ldquo;control&rdquo; for the default
```
yanirs, line 273: "Define the percentages to include in the experiment" is a bit awkward.

Maybe something like this? "Set the percentage of traffic allocated to each variation. Percentages may sum to less than 100 to avoid allocating the entire userbase."

src/components/experiments/wizard/__snapshots__/Audience.test.tsx.snap
```diff
@@ -1049,7 +1165,7 @@ exports[`renders as expected 2`] = `
                 <span
                   class="MuiChip-label"
                 >
-                  Locale: en-AU
+                  Country: au
```
yanirs, line 1168: Was this supposed to change?

yanirs (commented):
Looks pretty good! My main issue is with the pre-populated names and percentages. I think that leaving them empty would be cleaner than adding default values.

src/components/experiments/wizard/__snapshots__/Audience.test.tsx.snap
jessie-ross: Looks like a race condition, I have fixed it using some `act`s.

yanirs (approved):
Looks like the snapshots need adjusting, but other than that LGTM :ship:

Issue #404, opened 2020-11-20
Add "Participant Allocation" for Experiment in Single-View and Wizard


jessie-ross:
As discussed in pbmo2S-zV-p2, we want to add a "Participant Allocation" section to help simplify the UI.

This section will contain Variations, Exclusion Groups, and Exposure Events.

PR #595, opened 2021-08-23
Remove experiment overview grid

yanirs:
As discussed on Slack, this PR replaces the Experiment Overview tab grid with full-width panels. This addresses exposure event overflows as seen on experiment 20252 and looks a bit better (in my opinion). As the page is pretty text-heavy, I don't think there's much value in having the Audience panel to the right of the General panel &ndash; with the full-width display, viewers can focus on one thing at a time. :slightly_smiling_face: 

## How has this been tested?

- Manual testing with production data: Specifically looked at experiment 20252 and at a staging experiment


jessie-ross (approved):
LGTM

Might be a good idea to show to at least one experimenter to get their opinion?

yanirs:
> Might be a good idea to show to at least one experimenter to get their opinion?

Given that it's a fairly minor change and we both agree on it, I think it's fine.

We might want to get more feedback on the bigger reshuffle in #404 (or even involve a designer at some stage).

PR #596, opened 2021-08-24
Try to reduce CircleCI flakiness by separating integration and e2e tests

yanirs:
I noticed that the `integration-e2e` job sometimes fails on the `integration` step and sometimes on the `e2e` step. By separating the steps, we can rerun only the failed step and increase the overall probability of success.

## How has this been tested?

CircleCI tests on this PR:
image

If we go with this change, I'll adjust the settings to make the `integration` and `e2e` checks required and remove `integration-e2e` as a required check.


jessie-ross (approved):
🚢

PR #599, opened 2021-09-27
Align experiment name height on Experiment details page.

### LINKED ISSUE #591: Align experiment name height on experiment details
While clicking through Abacus, I found the experiment name slug is vertically misaligned with the "Experiment:" header. See screenshot for more details.

Screen Shot 2021-08-11 at 3 42 30 PM

is4ac:
Changed display property of `.titleName` from `'inline-block'` to `'inline'`.

The experiment title was misaligned previously. Changing the display property for the `titleName` class in the `ExperimentPageView` component fixed this issue.

Closes #591 

## How has this been tested?

- Ran all automated tests
- Manual testing with local development environment (checked on all major browsers)


jessie-ross (changes requested):
Before:
Screen Shot 2021-09-27 at 11 00 34 am

After:
Screen Shot 2021-09-27 at 10 58 54 am

The inline-block was actually necessary here to allow the text to shorten and use ellipsis when it gets too long.

I think it would be better to move some of these styles up a level from `titleName` to the `h2`, to make the whole `h2` ellipse properly on one line.

is4ac:
Great point, didn't catch that before.

Here is the new preview:

Screen Shot 2021-09-28 at 11 06 38 AM

I think this is what you meant by ellipsing on 1 line, correct?

jessie-ross (approved):
LGTM 🚢

PR #600, opened 2021-10-03
Change font styles for metric assignment results based on significance.

### LINKED ISSUE #555: Emphasize and suppress top-level results by significance
We received some feedback (pbmo2S-MY-p2#comment-1931) to distinguish statistically significant results. We show a ton of numbers and ranges, and it may be beneficial (and an educational opportunity) to emphasize/ suppress top-level metrics by their results statistical and/or practical significance.

We could use **font styles** to emphasize important results:
- Color results light grey (`rgb(204, 202, 202)`) if not statistically significant
- Color results current grey (`rgb(117, 117, 117)`) if statistically significant but not practically significant
- Color results `rgba(0, 0, 0, 0.87)` with font weight of `600` if statistically significant and practically significant (same style as metric name)

We could then **add an info icon** somewhere in the header educating experimenters about these three styles. What do y'all think? 

If we like this, we can consider combining this issue with https://github.com/Automattic/abacus/issues/524. I created this new issue, so we can discuss without creating noise in that issue.

Here's a mockup of what these style would look like:

Screen Shot 2021-06-16 at 11 35 51 AM

is4ac:
This PR emphasizes significant results and de-emphasizes insignificant ones in the Metric Assignment Results table.

- Bold font == practically and statistically significant
- Light grey font == practically insignificant and/or statistically insignificant
- The light grey is the same as the Health Report table grey
- Also added an info icon with a tooltip explaining the meaning of the style differences

ActualExperimentResults.tsx was edited to add the varying font styles based on if the results were practically significant and statistically significant.

In issue #555, it was mentioned that:

> We could then add an info icon somewhere in the header educating experimenters about these three styles.

Is the info icon with tooltip a good implementation of this, or is there a better way to design this?

Before
Screen Shot 2021-10-03 at 4 58 37 PM

After
Screen Shot 2021-10-03 at 4 38 50 PM

Closes #555 

## How has this been tested?

- Updated snapshot for testing this change

### DIFF
--- a/src/components/experiments/single-view/results/ActualExperimentResults.tsx
+++ b/src/components/experiments/single-view/results/ActualExperimentResults.tsx
@@ -16,7 +16,7 @@ import {
   Typography,
   useTheme,
 } from '@material-ui/core'
-import { ExpandMore as ExpandMoreIcon, InfoOutlined as InfoOutlinedIcon } from '@material-ui/icons'
+import { ExpandMore as ExpandMoreIcon } from '@material-ui/icons'
 import clsx from 'clsx'
 import _ from 'lodash'
 import MaterialTable from 'material-table'
@@ -49,17 +49,12 @@ import MetricAssignmentResults from './MetricAssignmentResults'
 import RecommendationDisplay from './RecommendationDisplay'
 
 const indicationSeverityClassSymbol = (severity: Analyses.HealthIndicationSeverity) => `indicationSeverity${severity}`
-const significanceStatusClassSymbol = (recommendation: Recommendations.Recommendation) => {
-  if (
+
+const isPracticallySignificant = (recommendation: Recommendations.Recommendation) => {
+  return (
     typeof recommendation.practicallySignificant !== 'undefined' &&
-    typeof recommendation.statisticallySignificant !== 'undefined' &&
-    recommendation.practicallySignificant === Recommendations.PracticalSignificanceStatus.Yes &&
-    recommendation.statisticallySignificant
-  ) {
-    return `significanceStatus${Recommendations.PracticalSignificanceStatus.Yes}`
-  } else {
-    return `significanceStatus${Recommendations.PracticalSignificanceStatus.No}`
-  }
+    recommendation.practicallySignificant === Recommendations.PracticalSignificanceStatus.Yes
+  )
 }
 
 const useStyles = makeStyles((theme: Theme) =>
@@ -132,8 +127,6 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     tableTitle: {
       margin: theme.spacing(4, 2, 2),
-      display: 'flex',
-      alignItems: 'center',
     },
     accordions: {
       margin: theme.spacing(2, 0),
@@ -157,12 +150,9 @@ const useStyles = makeStyles((theme: Theme) =>
     metricAssignmentNameLine: {
       whiteSpace: 'nowrap',
     },
-    [`significanceStatus${Recommendations.PracticalSignificanceStatus.Yes}`]: {
+    isSignificant: {
       color: 'rgba(0, 0, 0, 0.87)',
       fontWeight: 600,
-    },
-    [`significanceStatus${Recommendations.PracticalSignificanceStatus.No}`]: {
-      color: theme.palette.grey[600],
     },
     infoIcon: {
       color: theme.palette.info.main,
@@ -343,17 +333,18 @@ export default function ActualExperimentResults({
           return null
         }
 
+        const className = isPracticallySignificant(recommendation) ? classes.isSignificant : ''
+
         return (
-          <div className={classes[significanceStatusClassSymbol(recommendation)]}>
-            <MetricValueInterval
-              intervalName={'the absolute change between variations'}
-              isDifference={true}
-              metricParameterType={metric.parameterType}
-              bottomValue={latestEstimates.diff.bottom}
-              topValue={latestEstimates.diff.top}
-              displayTooltipHint={false}
-            />
-          </div>
+          <MetricValueInterval
+            className={className}
+            intervalName={'the absolute change between variations'}
+            isDifference={true}
+            metricParameterType={metric.parameterType}
+            bottomValue={latestEstimates.diff.bottom}
+            topValue={latestEstimates.diff.top}
+            displayTooltipHint={false}
+          />
         )
       },
       cellStyle: {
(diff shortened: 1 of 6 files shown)


jessie-ross (changes requested):
Hmm, yeah this is a bit tricky.

I think it would help to go back to the purpose of this change, which is to make a long list of metric assignments a bit more readable. Now that I am looking at an example design, I can see some potential issues with the overall idea:
- "Deploy either variation" is an important and significant result, which we don't want to get buried.
- Inconclusive can be important too - it could hide a potentially large uncertainty to hide these.

I am a bit on the fence about adding this feature but, maybe a toned down version where we just bold the recommendation when it is a deploy recommendation would work. What do you think?

(Also leaving some inline comments for learning.)

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -483,7 +515,14 @@ export default function ActualExperimentResults({
...
           <Typography variant='h3' className={classes.tableTitle}>
-            Metric Assignment Results
+            <span>Metric Assignment Results</span>
+            &nbsp;
+            <Tooltip
+              title='Results that are both practically and statistically significant are indicated in bold. 
+            Insignificant results are indicated by a lighter grey.'
+            >
+              <InfoOutlinedIcon className={classes.infoIcon} />
+            </Tooltip>
```
jessie-ross, line 525: I think if we need to explain the design we are going in the wrong direction.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -49,6 +49,18 @@ import MetricAssignmentResults from './MetricAssignmentResults'
 import RecommendationDisplay from './RecommendationDisplay'
 
 const indicationSeverityClassSymbol = (severity: Analyses.HealthIndicationSeverity) => `indicationSeverity${severity}`
+const significanceStatusClassSymbol = (recommendation: Recommendations.Recommendation) => {
+  if (
+    typeof recommendation.practicallySignificant !== 'undefined' &&
+    typeof recommendation.statisticallySignificant !== 'undefined' &&
```
jessie-ross, line 55: I think these should never be undefined, if you are using VSCode / another editor with a TS language server, it should let you know.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -49,6 +49,18 @@ import MetricAssignmentResults from './MetricAssignmentResults'
...
+    recommendation.practicallySignificant === Recommendations.PracticalSignificanceStatus.Yes &&
+    recommendation.statisticallySignificant
```
jessie-ross, line 57: Practically significant implies statistical significance 😉

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -143,6 +157,16 @@ const useStyles = makeStyles((theme: Theme) =>
     metricAssignmentNameLine: {
       whiteSpace: 'nowrap',
     },
+    [`significanceStatus${Recommendations.PracticalSignificanceStatus.Yes}`]: {
```
jessie-ross, line 160: If we were going round this path we could use a single class "isSignificant", with no need for dynamic class names.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -357,13 +383,15 @@ export default function ActualExperimentResults({
...
         return (
-          <MetricValueInterval
-            intervalName={'the relative change between variations'}
-            metricParameterType={MetricParameterType.Conversion}
-            bottomValue={Analyses.ratioToDifferenceRatio(latestEstimates.ratio.bottom)}
-            topValue={Analyses.ratioToDifferenceRatio(latestEstimates.ratio.top)}
-            displayTooltipHint={false}
-          />
+          <div className={classes[significanceStatusClassSymbol(recommendation)]}>
+            <MetricValueInterval
```
jessie-ross, line 387: With custom components like this and `RecommendationDisplay`, we should add `className` properties, instead of wrapping in a div.

is4ac:
> Hmm, yeah this is a bit tricky.
> 
> I think it would help to go back to the purpose of this change, which is to make a long list of metric assignments a bit more readable. Now that I am looking at an example design, I can see some potential issues with the overall idea:
> 
> * "Deploy either variation" is an important and significant result, which we don't want to get buried.
> * Inconclusive can be important too - it could hide a potentially large uncertainty to hide these.
> 
> I am a bit on the fence about adding this feature but, maybe a toned down version where we just bold the recommendation when it is a deploy recommendation would work. What do you think?

I agree that it might be a bit much to de-emphasize "insignificant" results, especially if they might contain valuable lessons for the experimenter. Bolding could work, but that still makes it hard to distinguish between different kinds of significant outcomes. 

What if we had color coded icons next to each type of recommendation? That way, it doesn't necessarily imply to the user that they should ignore certain results, but it would still allow for quick visual processing. For example, one idea could be:

- A green check mark for "Deploy treatment" or "Deploy either variation"
- A yellow sign (warning symbol?) for "Deploy control" or "Inconclusive"
- A red X for "Missing analysis" or "Manual analysis required"

I'm not exactly sure what the best way to color code the various results would be.. Thoughts?

src/components/experiments/single-view/results/ActualExperimentResults.tsx
is4ac: I don't think that's always true of designs in general (especially if it's a totally new interface), but in this case, I agree that it should be something that intuitively makes sense to people.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
is4ac: Gotcha, in the interface definition it has those two variables with a ? at the end, so I assumed it was possible that they could be undefined.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
is4ac: Ooh, right, good point! 😁 Still learning! 😊

src/components/experiments/single-view/results/ActualExperimentResults.tsx
jessie-ross: Oh, I think I was looking at the wrong place.

jessie-ross:
> What if we had color coded icons next to each type of recommendation? That way, it doesn't necessarily imply to the user that they should ignore certain results, but it would still allow for quick visual processing. For example, one idea could be:

Hmm I feel these could clutter the page a bit, and I don't think we should be associating some of these states as good/bad/warning as they may not mean that.

> Or we could make it match the Health Report table and change the background color of the Recommendation cells in the Metric Assignment Results table?

Yeah this is similar design wise as the current plan, just a different way to emphasise and de-emphasise, but linking it more to the categorical good/bad/warning.

---

I think we need to look at from a high level and decide:
- What states do we want to emphasise (why do we want to emphasise some states).
- What states do we want to de-emphasise (why do we want to de-emphasise some states).
- If we need to, is there a way to emphasise some rows without de-emphasising other rows.

aaronyan:
Where did y'all land in your 1:1 discussion around this?

is4ac:
> Where did y'all land in your 1:1 discussion around this?

@jessie-ross can correct me if I'm wrong, but I think we settled on tabling the issue until we could have a big picture discussion on what we want to actually emphasize or de-emphasize in the table. We thought it was low priority enough at the time that we didn't continue the discussion in favor of focusing on other things.

is4ac:
Closing in favor of (690-gh-Automattic/experimentation-platform)

Issue #604, opened 2021-10-20
Improve Metric UX


jessie-ross:
A lot of users are having issues with metrics, there are some issues around:
- Choosing metrics (Explorability)
- Setting min-diff (Wizard documentation + Process improvements)
- Understanding how metrics work
- Opening up adding/modifying metrics.

jessie-ross:
In particular - ensuring Experimenters know the min-diff is absolute rather than relative

aaronyan:
Nice job with the bite-sized stacked PRs @is4ac! The metric assignment table PRs look good to me, but it's probably best to let @jessie-ross review these when they're back next week.

PR #608, opened 2021-11-17
Metrics AG Grid: Convert MetricsTable on Metrics page to use AG Grid with details

is4ac:
## Improving Metric UX

**This PR is Part 4 of 4 of related PRs. Part 1 is #622, Part 2 is #623, and Part 3 is #644**

This PR is part 1 of a larger issue of trying to improve the UX surrounding metrics #604. This is the first attempt to improve the **explorability** of metrics. The current UX for viewing metrics can be a bit overwhelming, and will likely become a bigger problem in the future as more metrics are added. 

This PR uses the AG Grid with master/detail component from #623 and the GridContainer component from #644 to create a Metrics AG Grid. This adds minor improvements in terms of filtering and sorting compared to the previous table. In addition, the structural change will allow for easily adding this component to the Experiment Wizard for metric selection, which should provide a major boost to the explorability.

In the future, we could include tags in the table or other categorization features in order to make exploring metrics even friendlier.

## Next steps

The next steps are to add automated tests and to add a searchable/selectable metrics grid to the Experiment Wizard.

## How has this been tested?

- Automated unit testing


is4ac:
Tests have been added and PR is ready for review.

src/lib/schemas.ts
```diff
@@ -214,6 +214,10 @@ export const metricNewOutboundSchema = metricNewSchema.snakeCase().transform(
       : undefined,
   }),
 )
+export const metricDetail = metricSchema.shape({
```
jessie-ross, line 217: This shouldn't be a schema change - schemas are only for core business data types and shouldn't contain any UI related code.

You can extend MetricDetail in `src/components/metrics/MetricsTableAgGrid.tsx` and keep it local there.

jessie-ross (changes requested):
Hi Isaac! Thanks for the PR!

Some overall thoughts:
- Can we generalise the master/detail feature into it's own component? It might simplify things a bit - extricating the new table functionality from its application. I can also see it as something we might want to use for metric results. It could also be a good opportunity for splitting this PR up further.
- There is a bug with the master/detail feature where when you first open an item it covers the items below it.
- Can we make the metric config searchable? A hidden field or something like that?

src/lib/schemas.ts
is4ac: Got it! That makes sense!

is4ac:
> * Can we generalise the master/detail feature into it's own component? It might simplify things a bit - extricating the new table functionality from its application. I can also see it as something we might want to use for metric results. It could also be a good opportunity for splitting this PR up further.

Certainly! That's a good idea, and I was wondering if I should have done that.

> * There is a bug with the master/detail feature where when you first open an item it covers the items below it.

Hmm, do you mean the issue where only part of the detail loads at first before rendering the full height of the detail? I don't think it covers the items below it, unless I'm missing something (maybe a browser issue?). If it's the same glitch that I'm seeing, however, then I think I know how we can work around it, but it would be much more complicated than the current solution. 

The problem currently is that the auto row height feature of AG Grid doesn't work for full width rows, so I had to hack together a solution. I think there could be a work around where we pre-render the detail row (but hide it) and calculate the height before rendering it to the grid.

> * Can we make the metric config searchable? A hidden field or something like that?

What do you mean by metric config? Is that referring to the parameters JSON? Then yeah, I think that'd be simple enough.

jessie-ross:
> Hmm, do you mean the issue where only part of the detail loads at first before rendering the full height of the detail?

See the below video - it's when I open the first or second metric the first time:
https://d.pr/v/ijPanK/9Oox6HEPEF

> If it's the same glitch that I'm seeing, however, then I think I know how we can work around it, but it would be much more complicated than the current solution.

Since it seems to work after the first open, a fix might be to trigger some kind of redraw after it opens.

> Is that referring to the parameters JSON?

Yep, that's what I mean :)

is4ac:
> See the below video - it's when I open the first or second metric the first time:
https://d.pr/v/ijPanK/9Oox6HEPEF

Oh strange, I don't see that bug. Can you tell me what browser/OS you're on, in case that's making the difference?

jessie-ross:
> Oh strange, I don't see that bug. Can you tell me what browser/OS you're on, in case that's making the difference?

Ah sorry I didn't include it: 

```
Firefox 94.0.2
Darwin 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64
```

is4ac:
> > Oh strange, I don't see that bug. Can you tell me what browser/OS you're on, in case that's making the difference?
> 
> Ah sorry I didn't include it:
> 
> ```
> Firefox 94.0.2
> MacOS 11.6
> ```

Hmm, so I just tried testing it with those exact settings and I don't see the bug still.. (well I'm using MacOS 11.6.1 technically, but probably not the reason?). 

Is there potentially a browser extension that might be impacting it? Not sure what could be causing the discrepancy... Any ideas of how to go about debugging this?

is4ac:
Hey @jessie-ross! I've reworked this PR and split it up a bit. I hope that helps with the readability.

I'm very curious to see if this version contains the same visual bug as before though. I have yet to be able to recreate your bug on my machine, unfortunately, so I couldn't test it. But I'll leave a comment at the line I think that could solve it if the bug is still there.

src/components/general/AgGridWithDetails.utils.tsx
```diff
@@ -61,28 +46,33 @@ export const setHeightOfFullWidthRow = (
...
     const rowChild = found.firstElementChild
-    if (rowChild !== null) {
-      const rowHeight = rowChild.clientHeight
-      maxRowHeightMap.set(detailIdFromDataId(getDataId(data)), rowHeight)
-    }
+    const rowHeight = rowChild.clientHeight
+    maxRowHeightMap.set(detailIdFromDataId(getDataId(data)), rowHeight)
 
     gridApiRef?.resetRowHeights()
+    //gridApiRef?.redrawRows()
```
is4ac, line 75: This line could potentially fix the visual bug you were seeing before in FireFox. I'd like to leave it out if it isn't needed though, because it also adds an additional "flash" whenever you click a row.

is4ac: Actually, with the code changes, I ended up seeing this bug myself. So I decided to add in the `redrawRows` to make sure the bug doesn't occur. There is a flash the first time a detail row is opened for the first time, but it shouldn't be too distracting.

is4ac:
> I'm very curious to see if this version contains the same visual bug as before though. I have yet to be able to recreate your bug on my machine, unfortunately, so I couldn't test it. But I'll leave a comment at the line I think that could solve it if the bug is still there.

I left a comment on the code, but I ended up seeing the bug and fixed it using the `redrawRows` function after all.

PR #617, opened 2021-12-13
Add code splitting and bump react-scripts and node versions to reduce memory usage

is4ac:
## Changes in this PR

This PR adds code splitting and updates the `react-scripts` and `node` versions used in CircleCI in order to reduce memory usage during the production build process. This allows our e2e tests on CircleCI to pass again.

While researching, I found that one recommendation to reduce memory usage while building was to [use code splitting](https://create-react-app.dev/docs/code-splitting/). After analyzing the bundle size, I discovered that the `plotly.js` module took up a majority of the bundle size. So, I [imported a partial bundle of plotly](https://github.com/plotly/react-plotly.js/issues/98)) and added lazy loading to reduce the bundle size. In the future, it might even be worth investigating creating our own custom plotly bundle that decreases the size even further. 

The AG Grid library was the second biggest culprit, so I added lazy loading for that component as well.

**Note**: The number of lines changed looks gigantic, but most of it is changed snapshot files. If you'd like me to still split this PR into chunks, I can try, but it may be difficult to get something through that passes the e2e tests.

## How has this been tested?

- Automated unit tests
- Updated snapshots


src/components/general/LazyPlot.test.tsx
```diff
@@ -0,0 +1,84 @@
+import { prettyDOM, screen } from '@testing-library/react'
+import React from 'react'
+
+import * as Visualizations from 'src/lib/visualizations'
+import { render } from 'src/test-helpers/test-utils'
+
+import LazyPlot from './LazyPlot'
+
+// To remove non-deterministic elements from Plotly code
+// https://stackoverflow.com/a/69925334
```
is4ac, line 10: This is pretty hacky and could probably be improved, but I couldn't think of a better way at the time. If anyone has any suggestions on how to handle this.. I saw how the previous tests in `ExperimentResults` handled it, but unfortunately with lazy loading of the plotly library, it doesn't seem to be as straightforward to mock.

src/components/general/LazyPlot.tsx
```diff
@@ -0,0 +1,54 @@
+import { LinearProgress } from '@material-ui/core'
+import { Alert } from '@material-ui/lab'
+import React, { useEffect, useRef, useState } from 'react'
+import { PlotParams } from 'react-plotly.js'
+import createPlotlyComponent from 'react-plotly.js/factory'
+
+interface LazyPlotParams extends PlotParams {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
is4ac, line 8: Not sure how to avoid this one, given that this is the type definition for `import`. I'm not an expert on TypeScript, any ideas, or is this okay?

package.json
```diff
@@ -36,13 +37,13 @@
     "mathjs": "^9.3.2",
     "normalizr": "^3.6.0",
     "notistack": "^1.0.0",
-    "plotly.js": "^1.55.2",
+    "plotly.js-cartesian-dist": "^2.8.0",
```
jessie-ross, line 40: It seems like this is a major update of Plotly, can you ensure there have been no changes affecting our graphs?

package.json
```diff
@@ -94,6 +95,7 @@
     "jest-puppeteer": "^4.4.0",
     "lint-staged": "^10.1.7",
     "mockdate": "^3.0.2",
+    "plotly.js": "^2.8.0",
```
jessie-ross, line 98: I am curious to why this is needed as a dev dependency?

jessie-ross (commented):
Hmm, lazy loading elements seems like a lot of complexity just to get E2E tests working 🤔

package.json
is4ac: Ah yes, I can double check that. Manual testing seems to show that nothing was affected, but I can take a closer look. Worst case scenario, I can install the 1.55.2 version of `plotly.js-cartesian-dist`.

package.json
is4ac: It was for TypeScript types. Having `@types/plotly.js` as a dependency without having `plotly.js` caused compile errors, so I added it as a dev dependency so it wouldn't increase our bundle size.

is4ac:
> Hmm, lazy loading elements seems like a lot of complexity just to get E2E tests working 🤔

Yeah, it does seem complex, but I tried various other fixes suggested on GitHub issues and other forums, and none of them seemed to work for our situation. 

The root problem for us does seem to be the fact that our bundle size is too large. Before this change, our build command would display a warning about how our bundle size was much larger than recommended, so I think code splitting is something we needed to do anyway. I think we do gain other minor performance benefits from doing code splitting, so it wouldn't necessarily be wasted effort that only affects our e2e tests. 

I did just realize that I can make the `LazyPlot` component a bit simpler and have it mirror the `LazyAgGrid` component. I don't know if that will address your concerns, but I will go ahead and make that change so you can see what I mean.

Of course, I am open to simpler solutions for reducing memory usage if you can find them!

package.json
is4ac: On closer look, it seems that version update broke the default hover mode, so I will switch it to 1.55.2, which seems to fix it.

is4ac:
Discussed on Slack and over Zoom and decided to pursue other solutions. Possibly using a TeamCity server in the future.

PR #622, opened 2021-12-22
Metrics AG Grid: Add a ChevronToggleButton component for use in AG Grid master/detail component

is4ac:
## Changes in this PR

**This is Part 1 of 4 of the larger PR #608. Part 2 is #623, Part 3 is #644, and Part 4 is #608.**

This PR adds a `ChevronToggleButton` component for use in the AG Grid master/detail component that will be added in a later PR. This is part of addressing #604 in converting the Metrics table to use AG Grid.

**Note:** This PR does not actually use the `ChevronToggleButton` component anywhere yet. It just includes the component code and tests. A future PR for the AG Grid master/detail component will make use of this component.

## How has this been tested?


src/components/general/RowToggleButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  icon?: JSX.Element
+  onClick?: () => void
+}): JSX.Element => {
+  const classes = useStyles()
+  const [cls, setCls] = useState<string>(toggled ? classes.rotated : classes.notRotated)
+  const [isToggled, setIsToggled] = useState<boolean>(toggled)
+
+  useEffect(() => {
+    setCls(isToggled ? classes.rotated : classes.notRotated)
+  }, [isToggled, classes.rotated, classes.notRotated])
```
jessie-ross, line 35: There is no need to have cls as state, it should be a value directly derived from toggled.

src/components/general/RowToggleButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  useEffect(() => {
+    setIsToggled(toggled)
+  }, [toggled])
```
jessie-ross, line 39: It is an anti-pattern to be caching state within a component, [we should be lifting this state up and out of here](https://reactjs.org/docs/lifting-state-up.html#lessons-learned).

This component shouldn't have any state and shouldn't need any effects, it should simply have params: `isOpen`, and `onClick`.

src/components/general/RowToggleButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  },
+  rotated: {
+    transform: 'rotate(90deg)',
+  },
+  notRotated: {
+    transform: 'none',
+  },
+})
+
+const RowToggleButton = ({
```
jessie-ross, line 18: Let's decouple the naming from the use-case, how about `ChevronToggleButton`

src/components/general/RowToggleButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  useEffect(() => {
+    setCls(isToggled ? classes.rotated : classes.notRotated)
+  }, [isToggled, classes.rotated, classes.notRotated])
+
+  useEffect(() => {
+    setIsToggled(toggled)
+  }, [toggled])
+
+  const handleClick = () => {
+    setIsToggled(!isToggled)
```
jessie-ross, line 42: Case in point: What if the outer component wants to modify the logic that occurs on click, maybe have loading state, maybe disable state, etc.

jessie-ross (changes requested):

src/components/general/RowToggleButton.tsx
is4ac: Thanks, that makes sense. I think I ended up with this design because there were some complications with how AG Grid renders (or doesn't re-render) cell renderers. I'll do some more digging to fix that pattern!

src/components/general/RowToggleButton.tsx
is4ac: I like that idea!

is4ac: Actually, I just realized I wanted to make it possible to customize the icon. What about `IconToggleButton`?

jessie-ross: Rotating sounds good, but I am not sure when this would be used? Do you have a use case?

is4ac: > I am not sure when this would be used? Do you have a use case?

Do you mean besides where it's being used in AgGridWithDetails? I guess it could be used in other custom components that need a button to indicate expansion. If you don't think this needs to be a separate file, it could easily be added to the AgGridWithDetails.utils file.

jessie-ross: As in where we would use a custom icon for this component?

is4ac: > As in where we would use a custom icon for this component?

Oh, I see! I guess I was just thinking if a designer ever wanted to create custom icons in the future, it would be easy to swap out, for not much engineering effort. If you don't think it's necessary I could definitely take out that part.

jessie-ross: Yeah, I think this is a case of YAGNI and it might be a good design constraint to ensure all open/close buttons are consistent.

is4ac:
Renamed to `ChevronToggleButton` and got rid of the custom icon option. Let me know how it looks now!

src/components/general/ChevronToggleButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+}: {
+  isOpen: boolean
+  className?: string
+  icon?: JSX.Element
+  onClick?: () => void
+}): JSX.Element => {
+  const classes = useStyles()
+  const toggleClass = isOpen ? classes.rotated : classes.notRotated
+
+  const handleClick = () => {
```
jessie-ross, line 31: Let's also make `onClick` required so we don't need this function.

src/components/general/ChevronToggleButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+})
+
+const ChevronToggleButton = ({
+  isOpen,
+  className,
+  onClick,
+}: {
+  isOpen: boolean
+  className?: string
+  icon?: JSX.Element
```
jessie-ross, line 25: I think this was left in by accident.

src/components/general/ChevronToggleButton.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+test('calls callback function when clicked', async () => {
+  const mockCallback = jest.fn()
+  render(<ChevronToggleButton isOpen={false} onClick={mockCallback} />)
+  const button = await screen.findByLabelText('Toggle Button')
+  fireEvent.click(button)
+
+  expect(mockCallback.mock.calls.length).toBe(1)
+})
+
+test('button rotates depending on isOpen prop', async () => {
```
jessie-ross, line 22: Let's go ahead with this PR, but a note for the future: Since ChevronToggleButton is so small and simple I would include it in the next PR, and not write any unit tests specifically for it - it will be covered by where it is used.

jessie-ross (approved):
Just a couple of last changes, but other than that it is good to go.

PR #634, opened 2022-01-05
Retain search history: Add @testing-library/user-event for testing ExperimentsTableAgGrid

is4ac:
**Stacked PR 1/2 for retaining search history**: next #635 

This PR adds [the `@testing-library/user-event` module](https://testing-library.com/docs/ecosystem-user-event/) which allows for more advanced simulation of browser interactions than `fireEvent`. This was necessary (or at least made more convenient) the writing of tests for the new `ExperimentsTableAgGrid` code that allows for retaining search history.

Related to #545  

## How has this been tested?

- Doesn't affect any other parts of the codebase

PR #635, opened 2022-01-05
Retain grid history: Save last search, sort, and filter state as URL params

### LINKED ISSUE #545: Retain experiment search string in browser history

is4ac:
**Stacked PR 2/2 for retaining search history:** based on #634.

Saves the search string as a URL param, e.g. `/experiments?search=explat_test` so that the last search filter is retained. Allows the back button to work properly on Abacus.

This series of stacked PRs closes #545.

## How has this been tested?


aaronyan:
Tested the save last search string, and it works well for me!

jessie-ross (changes requested):
- I think we should be looking at url state on the page level from the `/pages` component, passing page parameters up and down. I think we should be able to avoid the state caching in the middle.
- We need to cleanly abstract the parameters and url code from the rest of the code - lets have `pageParams`, `setPageParams` and maybe `replacePageParams` (to indicate a history replacement) abstracting it away.
- We need a way to "type" pageParams depending on the page.
- The other side of this is we want to abstract the certain values out of the grid, we want to be able to pass in `gridState` and `setGridState`.
- We also want this to be decoupled from any particular grid so we can apply it to Metrics table as well.

is4ac:
Those are all great ideas! Thanks for the direction.

> * We need a way to "type" pageParams depending on the page.

I was having difficulty with this one. I was originally trying to [use template string pattern index signatures](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#symbol-and-template-string-pattern-index-signatures) to do this properly, but the project's version of TypeScript needs to be updated, and I wasn't sure how to update it and get it to work with CRA. Any ideas or help on this one would be appreciated.

is4ac:
Still a work in progress, but I've tried abstracting the parts you mentioned into 2 custom hooks: `useGridState` to manage the grid search, sort, and filter states, and a `usePageParams` that manages the URL params and history manipulation. The functionality works well so far, and I think it's made the code a lot cleaner and more reusable, but I'm still working on commenting and polishing it up. 

@jessie-ross what do you think of the direction so far?

Also: I might just end up merging the remaining 3 PRs into one PR since they are all related features. This version has the full functionality of saving the search, sort, and filter states.

src/utils/url-params.ts
```diff
@@ -0,0 +1,77 @@
...
+export const getParamsObjFromSearchString = (search: string): UrlParams => {
+  return Object.fromEntries(new URLSearchParams(search).entries()) as UrlParams
+}
+
+/**
+ * TODO comments
+ *
+ * @param initialParams TODO
+ */
+export const usePageParams = (initialParams?: UrlParams): UsePageParamsProps => {
```
jessie-ross, line 44: I would use a typescript generic for `UrlParams` and perhaps we should pass in a schema to validate the params.

jessie-ross: Naming: UrlParams vs ParamsObj vs PageParams - we should stick to one

src/utils/url-params.ts
```diff
@@ -0,0 +1,77 @@
...
+  setPageParams: (params: UrlParams) => void
+  replacePageParams: (newParams: UrlParams) => void
+}
+
+/**
+ * TODO comments
+ *
+ * @param paramsObj TODO
+ */
+export const getParamsStringFromObj = (paramsObj: UrlParams): string => {
```
jessie-ross, line 19: `FromObj` is a bit lazy, lets write full names

src/utils/url-params.ts
```diff
@@ -0,0 +1,77 @@
...
+ * TODO comments
+ *
+ * @param initialParams TODO
+ */
+export const usePageParams = (initialParams?: UrlParams): UsePageParamsProps => {
+  const history = useHistory()
+  const location = useLocation()
+
+  const params = initialParams || getParamsObjFromSearchString(location.search)
+  const [pageParams, setPageParams] = useState<UrlParams>(params)
```
jessie-ross, line 49: I think we should be able to avoid cached state in here by using and updating the URL state directly along with default state if anything isn't set.

src/utils/url-params.ts
jessie-ross, line 44: `initialState` should instead be `defaultState`.

src/utils/url-params.ts
```diff
@@ -0,0 +1,77 @@
...
+        pathname: location.pathname,
+        search: `?${getParamsStringFromObj(newParams)}`,
+      })
+    },
+    [history, location.pathname],
+  )
+
+  return {
+    pageParams,
+    setPageParams,
```
jessie-ross, line 74: You should draw a diagram to understand the way state updates, ideally everything should flow down.

If we give out export state we need to update state upwards and have to keep state in two places (here and in the url) in sync.

src/utils/ag-grid.test.tsx
```diff
@@ -0,0 +1,289 @@
+import { render } from '@testing-library/react'
```
jessie-ross, line 1: In general I would avoid writing tests until you know what the solution is going to look like.

src/utils/ag-grid.ts
```diff
@@ -0,0 +1,324 @@
...
+  }
+}
+
+/**
+ * TODO comments
+ *
+ * @param initialGridState TODO comments
+ */
+export const useGridState = (initialGridState?: GridState): UseGridStateProps => {
+  const initialState = initialGridState || {
```
jessie-ross, line 272: We should have grid state just be an object that describes grid state, we should be able to remove reference management from it.

src/pages/experiments/ExperimentsAgGrid.tsx
```diff
@@ -1,24 +1,98 @@
...
+}
+
 const Experiments = function (): JSX.Element {
   debug('ExperimentsPage#render')
 
   const { isLoading, data: experiments, error } = useDataSource(() => ExperimentsApi.findAll(), [])
 
   useDataLoadingError(error, 'Experiment')
 
+  const { gridState, updateGridState } = useGridState(defaultGridState)
```
jessie-ross, line 47: We shouldn't need grid state stored in this component, we should have a `pageParams -> gridState` function and `gridState -> pageParams` to use in `onGridStateChange((gridState) => pushPageParams(gridStateToPageParam(gridState)))`

src/utils/url-params.ts
is4ac: This was great advice! It helped me solve some frustrating issues with the previous method with infinite loops/race conditions.

src/utils/url-params.ts
is4ac: How about `UrlSearchParams` to differentiate it from react-router-dom `useParams`? I think UrlParams or PageParams isn't descriptive enough, and ["search" params matches the standard query string terminology](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

src/pages/experiments/ExperimentsAgGrid.tsx
is4ac: Currently, I can't get the Reset button to work without using a ref because resetting multiple parts of the grid state at once (search text, sort, filter) causes `onGridStateChange` to trigger multiple times and causes a race condition.

So does having a `gridStateRef` in this component acceptable?

is4ac:
Addressed everything except for the typing and schema validation because I wasn't very confident on how to do that yet. I hope how it's set up currently makes it not too difficult to add.

PR #636, opened 2022-01-05
Retain search history: Save last sort options

is4ac:
**Stacked PR 3/4 for retaining search history:** based on #635. Next #637 

Saves sorting options on the experiments AG grid to the URL params.

Related to #545 

## How has this been tested?


aaronyan:
Tested the save last sort options, and they work well for me!

PR #638, opened 2022-01-05
Expand width of "Name" column on main experiments table

### LINKED ISSUE #612: Increase width of Name column on experiments overview page
An unintended side effect of fixing #610 was that now the experiments overview page looks like this:

Screen Shot 2021-12-02 at 7 39 51 PM

Now all of the columns are equal width, but that hides a lot of the longer names of experiments. We should set default column widths so that the proportions of the grid look better.

Should be a simple fix, I could even tackle it after finishing my current task.

is4ac:
Adds a default width to the "Name" column on the experiments table so that it more conveniently shows longer experiment names.

### Preview

Screen Shot 2022-01-05 at 1 52 20 PM

Previously, the "Name" column had the same width as all the other columns, which made it difficult to read experiment names which are typically longer.

Closes #612 

## How has this been tested?


aaronyan (approved):
LGTM!

Issue #647, opened 2022-01-17
Update the early monitoring query to fit WYeast


jessie-ross:
The early monitoring query needs to be updated to fit WYeast's Hue.

See p1642416300311600-slack-C03DKP1JP

PR #659, opened 2022-02-25
Update early monitoring query to use Trino syntax

aaronyan:
Resolves https://github.com/Automattic/abacus/issues/647.

Adds the new syntax for the early monitoring query. To test, go to the new Hue on Wyeast (p4qSXL-4gR-p2) and test out the new query. You can use something like `experiment_id: 20520`.

## How has this been tested?


jessie-ross (approved):
Thanks for this!

Issue #660, opened 2022-03-04
Add new recommendations


aaronyan:
Let's implement the new recommendations based on these designs (pbmo2S-14p-p2). Here's a summary:

## Sub-scopes
- [x] Change 'Recommendation' to 'Analysis' (🟩 🟩 🟩 🟩 🟩)
- [x] Add more types of analysis (🟩 🟩 🟩 🟩 🟩) #672
- [x] Add deployment recommendations (🟩 🟩 🟩 🟩 🟩) #674
- [ ] ~Add simplified square diagrams to metric overview header (🟩 🟩 🟩 🟨 ⬛) [WIP]~
- [x] Add credible intervals for relative change (🟩 🟩 🟩 🟩 🟨) #675

### Before

<img width="961" alt="group-58-2" src="https://user-images.githubusercontent.com/4505888/156828385-2825079a-88b8-4de0-8bf9-a9a97f991548.png">

### After

<img width="961" alt="group-58_3-1" src="https://user-images.githubusercontent.com/4505888/156828390-e2294f6d-bf57-4e53-9a44-181f8f545b7c.png">

## Change 'Recommendation' to 'Analysis'

We are moving away from the metric overview header showing "Recommendation" and replacing it with "Analysis". This may involve changing variable names in Abacus. 

## Add more types of analysis

Abacus currently has three recommendation types: “deploy {variant}”, “deploy any variation”, and “inconclusive”. We want to add more types of analysis, so you can better understand your experiment results even when there is not a clear winner.

- _{Variant} wins_ – There is high certainty that the variant is statistically and practically significant.
- _{Variant} ahead_ – Variant is ahead and is statistically different, but there is not enough certainty to say the change is practically significant.
- _{Variant} barely ahead_ – Variant is barely ahead and is statistically different, but there is high certainty that the change isn’t practically significant.
- _Inconclusive_ – There is not enough certainty to draw a conclusion at this time.
- _No difference_ – There is high certainty that difference in performance is not practically significant. Learn more.

<img width="1480" alt="Screen Shot 2022-04-14 at 11 20 48 AM" src="https://user-images.githubusercontent.com/4505888/163454099-ee03b071-512b-4585-bf26-686e01624248.png">

## Add deployment recommendations

Within the **overview** (see image of Google sheet), we should provide deployment recommendations only if certain conditions are met. Here is that flowchart:

[...]

PR #672, opened 2022-04-01
Introduce new types of analysis

razvanpapadopol:
Add more types of analysis as described in #660 
- [x] Update logic for new (and existing) types
- [x] Update analysis titles
- [x] Update analysis overview and explanation lines 1 & 2
- [x] (separate PR https://github.com/Automattic/abacus/pull/674) Update deployment recommendations
- [ ] (separate PR) Add horizontal bar with yellow/green indicators 

## How has this been tested?


src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -156,6 +160,8 @@ const explanationLine2: Record<Recommendations.PracticalSignificanceStatus, stri
     .Uncertain]: `Uncertainty is too high because the absolute change overlaps with the specified minimum practical difference between `,
   [Recommendations.PracticalSignificanceStatus
     .No]: `With high certainty, there is no practical difference between the variations because the absolute change is inside the specified minimum difference between `,
+  [Recommendations.PracticalSignificanceStatus
+    .IncludesRope]: `With high certainty, there is no practical difference between the variations because the absolute change includes the specified minimum difference between `,
```
jessie-ross, line 164: I mentioned this below as well: [the logic](https://github.com/Automattic/abacus/pull/672/files#diff-c9dda79fde675d0b1554966dd1a22c4beca1e2f2723a538046f07299aad75018R100-R101) indicates that the Diff CI includes the ROPE, in that case there uncertainty rather than high certainty - there is the possibility of practical difference still.

jessie-ross (commented):
General shape of the change is good, see comment about the specifics around using "practicalDifference" and avoiding ROPE.

src/components/experiments/single-view/results/AnalysisDisplay.tsx
```diff
@@ -15,6 +15,16 @@ const useStyles = makeStyles((theme: Theme) =>
   }),
 )
 
+function getChosenVariationName(analysis: Recommendation, experiment: ExperimentFull): string {
```
jessie-ross, line 18: I think it would be more reusable to get the whole variation rather than just the name: `getChosenVariation`

The basic principle here is to pass around the collections - i.e. interfaces we have defined rather than individual items.

src/lib/recommendations.ts
```diff
@@ -113,9 +121,12 @@ export function getDiffCredibleIntervalStats(
 export enum Decision {
   ManualAnalysisRequired = 'ManualAnalysisRequired',
   MissingAnalysis = 'MissingAnalysis',
+  AbsoluteChangeIncludesRope = 'AbsoluteChangeIncludesRope',
```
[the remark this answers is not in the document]
razvanpapadopol: The second description line is different but I see that we're not directly using the `Decision` for it so I think it's ok to remove this and use `Inconclusive` instead. Thanks!

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
razvanpapadopol: I've updated the description now. It was one of the WIP items (unchecked ToDos from PR descriptions) 🙈

src/lib/recommendations.ts
```diff
@@ -86,13 +87,20 @@ export function getDiffCredibleIntervalStats(
...
     metricAssignment.minDifference <= analysis.metricEstimates.diff.bottom ||
     analysis.metricEstimates.diff.top <= -metricAssignment.minDifference
   ) {
     practicallySignificant = PracticalSignificanceStatus.Yes
+  } else if (
+    // CI includes the experimenter set minDifference:
+    metricAssignment.minDifference < analysis.metricEstimates.diff.top &&
+    analysis.metricEstimates.diff.bottom < -metricAssignment.minDifference
+  ) {
+    practicallySignificant = PracticalSignificanceStatus.IncludesRope
```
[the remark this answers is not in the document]
razvanpapadopol: > but [the text here matches](https://github.com/Automattic/abacus/pull/672/files#diff-371c70a25deef9b4efcdb7ca1f7ac64bae51c82513c32e5cd392054f986b6c69R164) matches the latter.

[Updated the text](https://github.com/Automattic/abacus/pull/672#discussion_r840830477) and sorry if it was a source for confusion. 

> we could possible add a new dimension to DiffCredibleIntervalStats by adding practicalSignificanceDirection

Really interesting idea. I'll definitely consider it as it may help also when drawing the visualization of the analysis.

src/lib/recommendations.ts
```diff
@@ -126,10 +137,20 @@ export interface Recommendation {
   practicallySignificant?: PracticalSignificanceStatus
 }
 
-const PracticalSignificanceStatusToDecision: Record<PracticalSignificanceStatus, Decision> = {
-  [PracticalSignificanceStatus.No]: Decision.DeployAnyVariation,
-  [PracticalSignificanceStatus.Uncertain]: Decision.Inconclusive,
-  [PracticalSignificanceStatus.Yes]: Decision.DeployChosenVariation,
+function getDecision(practicallySignificant: PracticalSignificanceStatus, statisticallySignificant: boolean): Decision {
```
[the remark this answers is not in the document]
aaronyan: +1 I'm liking the logic for getting the `Decision`. 

The logic looks good to me! If we synchronize the explanation2 for inconclusive states, then I think we can get rid of `IncludesRope` and `AbsoluteChangeIncludesRope`.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: Ah, if this is a WIP PR please prefix it with `WIP:` or use the draft PR setting

jessie-ross:
Ah I've deleted the screenshots and the edits as they may contain sensitive info - this is a public repo.

src/lib/recommendations.ts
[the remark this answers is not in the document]
jessie-ross: > What new information is gained by creating a new indicator `practicalSignificanceDirection`?

This would allow us to isolate the state mentioned above - but if we are using the same explanation2 it seems we won't need it.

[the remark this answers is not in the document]
razvanpapadopol: Even if we were to use the same explanation2, `practicalSignificanceDirection` can be still useful for the simplified square diagrams:
- `positive`
Screenshot 2022-04-05 at 16 21 42

- `uncertain`
Screenshot 2022-04-05 at 16 21 48

However, if it's not needed elsewhere, we can contain the new dimension within the scope of the diagrams.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
razvanpapadopol: I was looking for a Label but prefixing also makes sense. I'll keep `WIP:` until all the ToDos are checked. Thanks!

src/lib/recommendations.ts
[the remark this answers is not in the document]
razvanpapadopol: Updated the function to take in DiffCredibleIntervalStats in https://github.com/Automattic/abacus/pull/672/commits/781a56c0e141c5f050a468f9ecc79261228bedc7

Cleaned up `IncludesRope` and `AbsoluteChangeIncludesRope` in https://github.com/Automattic/abacus/pull/672/commits/bf8dcaa3f25d2bd0b7eba8fb44eaf7672043d32e

src/components/experiments/single-view/results/AnalysisDisplay.tsx
razvanpapadopol: Good point! Done in https://github.com/Automattic/abacus/pull/672/commits/e62ad91b0649583230bf28a0211e4d12c251bd2e

src/lib/recommendations.ts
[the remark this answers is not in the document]
aaronyan: I don't see a strong use for `practicalSignificanceDirection`, but +1 to leaving it up to you!

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -136,19 +136,31 @@ const differenceOverviewMessages: Record<
   Record<StringifiedStatisticalDifference, string>
 > = {
   [Recommendations.PracticalSignificanceStatus.Yes]: {
-    true: `There is high certainty that the change is practically significant.`,
+    true: `Deploy %s with confidence. There is high certainty that the variant is statistically and practically significant.`,
```
aaronyan, line 139: Ah sorry @razvanpapadopol. In the spreadsheet, I should have put "variant" in curly braces to indicate it's a variable. Would it be possible to adjust so it to read "{variant} wins and there is high certainty the change is statistically and practically significant" where {variant} is either `treatment` or `control`? I've adjusted the sheet accordingly.

Screen Shot 2022-04-14 at 11 20 48 AM

aaronyan: I'd be cool with adjusting this later if it is easier to push the stacked PRs first! Especially since the stacked PR introduces `getChosenVariation()`.

aaronyan (approved):
LGTM! Thanks for making these adjustments @razvanpapadopol!

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -136,32 +136,41 @@ const differenceOverviewMessages: Record<
...
 
+function getOverviewMessage(experiment: ExperimentFull, recommendation: Recommendations.Recommendation) {
+  const message =
+    differenceOverviewMessages[recommendation.practicallySignificant as Recommendations.PracticalSignificanceStatus][
+      String(recommendation.statisticallySignificant) as StringifiedStatisticalDifference
+    ]
+  const variationName = recommendation.chosenVariationId
+    ? getChosenVariation(experiment, recommendation)?.name
+    : 'variant'
+  const mapReplaceObject: Record<string, string> = {
```
jessie-ross, line 160: [Lodash includes a nice little string templating system](https://lodash.com/docs/#template), I think we should use it here as we'd get extensibility and readability for free.

jessie-ross (approved):
LGTM too!

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
aaronyan: Thanks so much, Razvan! Looks great!

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
razvanpapadopol: Hey, @jessie-ross ! What do you think about the changes in [473c2d4](https://github.com/Automattic/abacus/pull/672/commits/473c2d411d23065a222813c9b241df3f7cacbb1c)?

I've tried initially to follow the 5th example in https://lodash.com/docs/4.17.15#template but the linter complains about using `${}` in a regular string. The mustache way should also help with readability as you immediately understand this is a template and not a regular ES6 interpolation. However, I'm not completely sold on the extensibility part 🤔

razvanpapadopol: You're right. This approach enables us to do more than `String.replace` if we need to and it's also reasonably readable. Let's keep it 👍🏼

PR #675, opened 2022-04-14
Add CI diagrams for relative change

razvanpapadopol:
part of https://github.com/Automattic/abacus/issues/660

## Summary of changes
* Use `CredibleIntervalVisualization` to render a diagram based on the 95% CI and the user-defined `minDifference`
* Set the center of the diagram always at Zero
* Set the edges of the diagram for each metric depending on the most extreme values (positive or negative) and render the rest proportionally (see Demo)

## To do
- [x] Try improving the scaling of the diagram. Right now it uses absolute top/bottom diff values and the user-defined “Minimum Difference”
- [x] Check if centering 0 to the diagram is a requirement
- [x] Check if possible (needed) to use dynamic width for the diagram
- [x] Check if we still have to remove the `Absolute Change` column or we'll limit the width of the metric name

## Demo
** It is using a subset of metrics from experiment `20508`
Screenshot 2022-06-28 at 00 25 28

<details>
  <summary>Before latest design focused changes</summary>
Screenshot 2022-05-24 at 23 23 09
</details>

<details>
  <summary>Original version</summary>
Screenshot 2022-04-15 at 02 58 17
</details>

## How has this been tested?


aaronyan (commented):
Yay, thanks for trying to tackle this @razvanpapadopol! Cool to see it coming together.

> Try improving the scaling of the diagram. Right now it uses absolute top/bottom diff values and the user-defined “Minimum Difference”

Scaling would be helpful in making the diagrams a bit cleaner. It would also be awesome if we could have all of the diagrams use the same scale.

> Check if centering 0 to the diagram is a requirement

Centering 0 is not required imo. But we should have 0 across all diagrams be in the same vertical position.

> Check if possible (needed) to use dynamic width for the diagram

No need for the width to be dynamic imo. Our results page is not really responsive, so we can consider making more of the page responsive later.

src/components/experiments/single-view/results/ActualExperimentResults.tsx
```diff
@@ -356,12 +332,12 @@ export default function ActualExperimentResults({
...
-          <MetricValueInterval
-            intervalName={'the relative change between variations'}
-            metricParameterType={MetricParameterType.Conversion}
+          <CredibleIntervalVisualization
+            metricEstimateDiff={latestEstimates.diff}
+            metricAssignment={metricAssignment}
             bottomValue={Analyses.ratioToDifferenceRatio(latestEstimates.ratio.bottom)}
             topValue={Analyses.ratioToDifferenceRatio(latestEstimates.ratio.top)}
-            displayTooltipHint={false}
+            domain={rangeOfDiff}
```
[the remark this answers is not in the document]
razvanpapadopol: Thanks for this suggestion, @jessie-ross 🙏🏼 
Feel free to check out the new logic even if I'm still working on stories and unit tests.

I know it contradicts what Aaron mentioned earlier about all of the diagrams using the same scale. Still, I've eventually realized that having a 'domain' per metric was the only (sane 😅) way to render a set of meaningful diagrams regardless of the experiment setup and results.

razvanpapadopol:
Hey, @aaronyan, could you have another look at the (way nicer) CI diagrams here?

Basically, I've followed Jessie's [suggestion](https://github.com/Automattic/abacus/pull/675#discussion_r854267843) and removed the logic that was determining the extreme values for all the metrics in order to render the diagrams and the Zero point accordingly. For the little we were gaining, it resulted in a messy scale for 95% of the actual experiments.

So we now have every metric showing a meaningful diagram, especially if the `minDifference` has been set close to the actual results. But at least if a metric is completely off, it doesn't affect the rest of them.

The only possible downside of this approach is that 0 is always centered even if all the rendered intervals for all the metrics are positive or negative. But I'd say it's quite a rare scenario and a minor loss of space.

I've updated the description with the latest state of things.

aaronyan:
Woow, nice work @razvanpapadopol! This looks so good. A picture is worth a thousand words. 🙌 

Tested this locally, and it works really well! The summaries at a glance are soooo much more helpful now. I can't wait for experimenters to get this in their hands (and for us to write a P2 post or data.blog post about this)!

I only have one piece of feedback to consider:

- What do you think about center aligning the header text "Relative change (lift)"?

> Basically, I've followed Jessie's https://github.com/Automattic/abacus/pull/675#discussion_r854267843 and removed the logic that was determining the extreme values for all the metrics in order to render the diagrams and the Zero point accordingly. For the little we were gaining, it resulted in a messy scale for 95% of the actual experiments.
> 
> So we now have every metric showing a meaningful diagram, especially if the minDifference has been set close to the actual results. And most importantly, if a metric is completely off, it doesn't affect the rest of them.
> 
> The only possible downside of this approach is that 0 is always centered even if all the rendered intervals for all the metrics are on the positive or the negative side. But I'd say it's quite a rare scenario and just a tiny loss of space.

That's cool! The diagrams look really good right now, and not having the same scale might actually be a feature. It might be misleading to compare metrics against each other. A 5% increase in revenue might be more meaningful than 30% more random button clicks.

> Check if we still have to remove the Absolute Change column or we'll limit the width of the metric name
> I propose we just leave it for now while adding a limit to the metric name column. We'll probably get back to it again soon in https://github.com/Automattic/abacus/issues/710

That's a good point. Yeah, I'm cool with leaving it in for now. I like the idea of replacing `Absolute change` with something like `Impact` and use the absolute units like "users" and "$" estimates.

src/components/experiments/single-view/results/CredibleIntervalVisualization.tsx
```diff
@@ -0,0 +1,125 @@
...
+          y2={DIAGRAM_HEIGHT - 12}
+          stroke='#4e4e4e'
+          strokeWidth={2}
+          strokeDasharray='4 4'
+          strokeDashoffset={1.5}
+        />
+
+        {/* middle black line */}
+        <line x1={scale(0)} x2={scale(0)} y2={DIAGRAM_HEIGHT} stroke='#4e4e4e' strokeWidth={2} />
+      </svg>
```
jessie-ross, line 113: Ah hmm, just plain svg now 🤔 I guess that should be ok.

src/components/experiments/single-view/results/CredibleIntervalVisualization.tsx
```diff
@@ -0,0 +1,125 @@
...
+}): JSX.Element {
+  const classes = useStyles()
+
+  const metric = {
+    top: metricEstimates.diffs[variationDiffKey].top_95,
+    bottom: metricEstimates.diffs[variationDiffKey].bottom_95,
+    minDifference: metricAssignment.minDifference,
+  }
+
+  const valueDomain = Math.max(metric.minDifference, metric.top, -metric.bottom)
```
jessie-ross, line 49: Simple and effective :)

jessie-ross:
Code looks good, I think it needs some design improvements, here are my thoughts:
- Need a slight background grouping the lower bound, graph, and upper bound - so it becomes one distinct object.
- I think the interval should have transparency and be in a layer above both the 0 and min-diff lines.
- Maybe 1.5x more margin between the graph and the bounds?
- I think we can improve visualisation of some edge cases:
    - If the min-diff is really close to the 0 line, place the min diff right next to but not covering the 0 line.
    - If a min-diff line is too far out, maybe add a little arrow next to the min-diff to indicate that: ->|

> The only possible downside of this approach is that 0 is always centered even if all the rendered intervals for all the metrics are on the positive or the negative side.

I think this is a feature :) It's nice to have a constant reference in all the charts.

razvanpapadopol:
Hi @jessie-ross! Thanks for the tips on design for sharing the TS error (it was some GIT duplication coming from a merge). I think everything is covered now so could you have a look at the UI while I push more non-visible changes (unit tests and code cleanup)? 🙏🏼 

> * Need a slight background grouping the lower bound, graph, and upper bound - so it becomes one distinct object.

I've done this using the lightest grey currently found in the application. I've tried also with a ligher shade (see below) but I think most screens won't even display it different than white.
Screenshot 2022-06-22 at 23 05 36

> * I think the interval should have transparency and be in a layer above both the 0 and min-diff lines.

Yes and yes. Done ✅ 

> * Maybe 1.5x more margin between the graph and the bounds?

I think it hasn't any margin previously. I've added now some tiny offset to prevent clipping + `spacing(1)` with the background grouping mentioned above

> * I think we can improve visualisation of some edge cases:
>   
>   * If the min-diff is really close to the 0 line, place the min diff right next to but not covering the 0 line. Although I am not too sure it is worth the effort 🤔
>   * If a min-diff line is too far out, maybe add a little arrow next to the min-diff to indicate that: ->| , - if we decide in some cases to not have the min-diff within the graph domain.

The min-diff or any edge of the interval really close to the 0 line is now fixed (we have also two examples in Storybook). If the min-diff is too far our I think it's actually good that we display it at the edges. Also confirmed with @aaronyan that any new element (eg: `->|`) will need additional explainers.

@aaronyan, the Relative Change header text is also centered now ✅

aaronyan:
Looking amazing @razvanpapadopol! Thanks for making those changes. The experiments `20647` and `20653` are great examples of how these interval diagrams greatly improves communicating the experiment results. Great work! I'm very excited for experimenters to receive this update.

> Need a slight background grouping the lower bound, graph, and upper bound - so it becomes one distinct object.

It's not my favorite, but I'm cool with keeping this for now. :)

razvanpapadopol:
Thanks for trying out the new changes, @aaronyan! I'm also looking forward to seeing this live and can't wait to get more feedback from experimenters!

> > Need a slight background grouping the lower bound, graph, and upper bound - so it becomes one distinct object.
> 
> It's not my favorite, but I'm cool with keeping this for now. :)

Same here. Maybe we can go with just an outline that would achieve the grouping but preserve the much needed contrast.

Screenshot 2022-06-23 at 12 24 12
@jessie-ross , WDYT?

jessie-ross:
> Need a slight background grouping the lower bound, graph, and upper bound - so it becomes one distinct object.

Ah by lower bound and upper bound I meant the numbers that are given before and after the graph to show that they are part of the same thing. E.g. in the first line including the `+0.24%` and `+5.16%`.

razvanpapadopol:
> > Need a slight background grouping the lower bound, graph, and upper bound - so it becomes one distinct object.
> 
> Ah by lower bound and upper bound I meant the numbers that are given before and after the graph to show that they are part of the same thing. E.g. in the first line including the `+0.24%` and `+5.16%`.

Oh, it makes sense. Thanks for the clarification 👍🏼 
Here is the new version. Looks good to me!

Screenshot 2022-06-23 at 18 52 26

aaronyan:
> I meant the numbers that are given before and after the graph to show that they are part of the same thing

Ohhh I see too. I like the background/ border now 😄. It definitely helps with indicating the % values are associated with the diagram. I'd vote the border option, but happy with both!

aaronyan:
> Pushed this (without the analysis square diagrams) so y'all can test it as well.

Thanks for testing out this idea, @razvanpapadopol. Looks great!

> I really like the idea! This would achieve the grouping effect without any need for border/background thus removing any potential distraction 🎉

That's a good point, and I'm cool with the interval numbers below the graph. It also gives more room for the graph to be wider too. :)

For the future analysis square diagram PR, what do y'all think about moving the analysis text below the analysis square diagrams? That way we'll have consistent design of diagram on top and text on bottom.

razvanpapadopol:
> For the future analysis square diagram PR, what do y'all think about moving the analysis text below the analysis square diagrams? That way we'll have consistent design of diagram on top and text on bottom.

I don't have a strong opinion here. Displaying the analysis text first makes sense so you first see the conclusion and then the "Why". Also, my non-designer eye (but with some photography experience) tells me it would be more visually balanced this way 😁 

Then we can also explore reordering the columns, as you proposed in https://github.com/Automattic/abacus/issues/660#issuecomment-1136543333

jessie-ross:
This looks good to me!!! I can't wait for experimenters to see it!

We can further discuss things in the square diagram :)

jessie-ross:
> >   For the future analysis square diagram PR, what do y'all think about moving the analysis text below the analysis square diagrams? That way we'll have consistent design of diagram on top and text on bottom.
>
> I don't have a strong opinion here. Displaying the analysis text first makes sense so you first see the conclusion and then the "Why". Also, my non-designer eye (but with some photography experience) tells me it would be more visually balanced this way 😁


Perhaps text could be on top for both?

jessie-ross:
Oops! Sorry I was excited and clicked the update button, I'll leave the PR to you

razvanpapadopol:
> Perhaps text could be on top for both?

It's worth trying out. I guess we can play a bit with the text size, color, and spacing to get to a good balance 👌🏼 

> Oops! Sorry I was excited and clicked the update button, I'll leave the PR to you

Haha, no worries! I'm happy to see all checks green 💚  
And so excited to see the thing live!! 🕺🏼

PR #684, opened 2022-05-05
Rework analysis transitioning

### LINKED ISSUE #683: Fix variation ordering for diffs/ratios in MetricEstimatesNext
During the deploy of the new backend we found that diffs and ratios were flipped and this was to do with the assumed ordering of variationIds.

We tried a hotfix in https://github.com/Automattic/abacus/pull/682, but then found out that variations aren't consistently ordered 😭.

We need some way of determining which variation is default or not [in the part of the code that transforms from MetricEstimatesNext to MetricEstimatesPrevious](https://github.com/Automattic/abacus/blob/7f9806b7c39383a04ed560926e7d2ebadef353e3/src/lib/schemas.ts#L601).

jessie-ross:
**This PR reworks analysis transitioning between previous and next to avoid needing to assume the variation ordering:**
- `AnalysisAPI.getByExperimentId` now returns `(AnalysisNext | AnalysisPrevious)[]`, this works a bit better as we are working less with raw data.
- The ensure functions now work on typed data and require specifying default and other variationIds.
- We ensure `AnalysisPrevious` in ExperimentPageView.

Resolves #683

## How has this been tested?


src/api/AnalysesApi.ts
```diff
@@ -9,17 +9,21 @@ import { fetchApi } from './utils'
...
+  const analyses = analysesResponse.analyses.map((maybeAnalysis) =>
+    // istanbul ignore next; Transitional
+    Object.hasOwnProperty.call(maybeAnalysis, 'metric_estimates') &&
+    !!maybeAnalysis?.metric_estimates &&
+    Schemas.isRawMetricEstimatesNext(maybeAnalysis?.metric_estimates)
+      ? Schemas.analysisNextSchema.validateSync(maybeAnalysis)
+      : Schemas.analysisPreviousSchema.validateSync(maybeAnalysis),
+  )
 
-  const { analyses } = await Schemas.analysisResponseSchema.validate(analysesResponse)
```
razvanpapadopol, line 22: I guess we may also remove the `analysisResponseSchema` and `AnalysisResponse` interface defined in `src/lib/schemas.ts`.

src/components/experiments/single-view/ExperimentPageView.tsx
```diff
@@ -120,14 +121,28 @@ export default function ExperimentPageView({
   const { isLoading: tagsIsLoading, data: tags, error: tagsError } = useDataSource(() => TagsApi.findAll(), [])
   useDataLoadingError(tagsError, 'Tags')
 
-  const { isLoading: analysesIsLoading, data: analyses, error: analysesError } = useDataSource(async () => {
+  const {
+    isLoading: analysesIsLoading,
+    data: analysesNextPreviousMixed,
```
razvanpapadopol, line 126: I think `analysesMixed` is probably enough to suggest what it's this about. Not a strong opinion though.

razvanpapadopol (approved):
The rework looks good! I guess the less we work with row data the better 🚀

src/api/AnalysesApi.ts
jessie-ross: It's going to come back soon, this is very transitory.

src/components/experiments/single-view/ExperimentPageView.tsx
razvanpapadopol: That's cool 🔥

Issue #670, opened 2022-03-24
Change name of UI from 'Abacus' to 'ExPlat'


aaronyan:
What do y'all think about changing the name of our UI to be "ExPlat"? I think the name "Abacus" is a point of confusion for some experimenters, and if we name the UI "ExPlat", it could help remove that friction. Since most experimenters will only interface with ExPlat using Abacus, it may simplify the experience if experimenters only need to "use ExPlat to run and manage your experiments".

This scenario reminds me of WordPress.com's [Calypso](https://github.com/Automattic/wp-calypso)! A8c uses "Calypso" internally to mean the front-end for WordPress.com, but the users only know it as "WordPress.com". The ExPlat team knows "Abacus" to mean the front-end for ExPlat, but the users (our experimenters) could know it as "ExPlat". What do you think?

This change would include:
- [ ] Change the name of the tool to be ExPlat (leave the repo name as Abacus)
- [ ] Change the documentation (FG)
- [ ] Publish a post sharing the renamed UI

jessie-ross:
Yep! I think this would reduce a lot of friction.

Thoughts:
- We should keep Abacus in the name to avoid confusion for those that know it as Abacus.
- The name should separate it from other part of ExPlat.
- We could add an additional domain name matching what we choose.

Some more options here:
- "ExPlat Control Center: Abacus"
- "ExPlat Manager: Abacus"
- "ExPlat Frontend: Abacus"
- "ExPlat Interface: Abacus"

aaronyan:
Closing this for now since we've made the simple change of going with "ExPlat: Abacus" as the title. We may revisit this branding issue sometime later. :)

PR #685, opened 2022-05-05
Include ExPlat in title

jessie-ross:
**As per some long lost discussions, this PR changes the title from "Abacus" to "ExPlat: Abacus".**

Before:

Screen Shot 2022-05-05 at 11 42 53 pm

After:

Screen Shot 2022-05-05 at 11 42 44 pm

## How has this been tested?


razvanpapadopol (approved):
Looks good even as an interim solution.
I believe this is the original discussion: https://github.com/Automattic/abacus/issues/670

aaronyan (approved):
LGTM to me too!

I'm not too sure what is the overall strategy for us with ExPlat vs Abacus, but I think adding both names here is a good first step.

Issue #687, opened 2022-05-06
Package update fun time


jessie-ross:
**It's time to update some packages!**

The big ones:
- [x] Node v16, similar to [Calypso](https://github.com/Automattic/wp-calypso/pull/60120) (in `nvm.rc` and `.circleci/config.yml`) https://github.com/Automattic/abacus/pull/806
- [x] Create react App
- [x] Typescript
- [ ] MUI 5
- [ ] React 18
- [x] Jest + Puppeteer
- [x] Testing library
- [x] Storybook https://github.com/Automattic/abacus/pull/716

Package maintenance is very important in JS land.

razvanpapadopol:
Have we explored configuring a bot for this, like the [GH Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates)? I guess it makes some noise with many PRs opened but at least it's a good automated reminder. 

For example, `wp-calypso` is using [Renovate](https://github.com/renovatebot/renovate) app. See the [testing-library](https://github.com/Automattic/wp-calypso/pull/61048) and [TypeScript](https://github.com/Automattic/wp-calypso/pull/61511) PRs.

jessie-ross:
I'm not against it, but I am a bit worried it will be too much change for our small team - it's quite nice to have everything stable and batch update maybe 1-4 times a year.

razvanpapadopol:
Ok, it makes sense to keep to the current cadence in our context 👍🏼  
And I've also noticed the existing Dependabot alerts for vulnerabilities so we should be safe.

aaronyan:
Do we have automated reminders for these updates? Even if we stick with manual updates, it could be nice to have automated notifications and reminders.

jessie-ross:
Yeah reminders sounds like a good idea, not the best way to do them.

jessie-ross:
Since I've completed most of the updates in this issue, I've created a new issue for updating the leftovers:
- https://github.com/Automattic/abacus/issues/860

PR #688, opened 2022-05-06
Fix warnings in unit tests

jessie-ross:
**This PR fixes warnings in unit tests.**

- There is one warning left that needs an update of `testing-library` #687

## How has this been tested?


src/components/experiments/single-view/ExperimentPageView.tsx
```diff
@@ -156,7 +156,7 @@ export default function ExperimentPageView({
           <Typography className={classes.titleHeader} variant='h2'>
             Experiment:{' '}
             {experiment ? (
-              <Tooltip title={experiment.name}>
+              <Tooltip title={`${experiment.name}`}>
```
razvanpapadopol, line 159: This will render `null` if the `name` is `null`. Let's use the empty string as a fallback.
```suggestion
              <Tooltip title={experiment.name ?? ''}>
```

razvanpapadopol (changes requested):
Thanks for fixing these! Left just one tiny note.

src/components/experiments/single-view/ExperimentPageView.tsx
jessie-ross: Ah good find!

razvanpapadopol (approved):
🚀

PR #690, opened 2022-05-09
Explain the new observed data

jessie-ross:
**This PR renames the observed data section and adds a little disclaimer.**

As part of the metric estimates refactor we no longer show raw observed data, this PR changes the UI to reflect that.

See 45-gh-Automattic/a8cds

## How has this been tested?


razvanpapadopol (approved):
Thanks for adding this clarification! Looks good as it is but for anyone interested to read more, we could also include a link to pbmo2S-1fH-p2.

Then, as a potential follow-up, we could link instead to a new paragraph about posterior means in the FG. Here maybe? PCYsg-FqD-p2 
cc: @aaronyan

aaronyan:
> Then, as a potential follow-up, we could link instead to a new paragraph about posterior means in the FG. Here maybe?

Added a FG page on the 'Analyze results' page in our ExPlat FG (how do you create a shortcode for a link to the FG page btw?). Thanks for the suggestion, @razvanpapadopol!

razvanpapadopol:
There are 2 ways of generating a shortcode for a FG page:
1. When here in GitHub, use Tampermonkey browser extension as described in PCYsg-Fqg-p2 and just copy-paste a link same as you do with Slack threads, P2 posts, etc.
2. Click on the three-dots button on the floating toolbox at the bottom-right corner in FG and there you have "Copy shortlink" which will generate the same shortlink as the first method.

aaronyan:
Woo! Thanks for the instructions @razvanpapadopol! I totally forgot I can just paste links like that with Tampermonkey installed. And TIL about the three dots at the bottom right corner for FG posts. Thanks!

Issue #500, opened 2021-04-14
Allow descriptions to be searchable in experiment directory


aaronyan:
- [x] We had a request (pau2Xa-2WO-p2#comment-10182) to have experiment description content be searchable in the experiment directory. That would be very helpful for experimenters to match more experiments to their search keywords.

- [ ] We can also consider adjusting the UI to include descriptions below the experiment name. Something in grey text that is in a smaller font compared to experiment name.

yanirs:
Something to keep in mind is that this would increase the payload size of GET `/experiments` if we want to stick to client-side filtering. Definitely in the low priority zone, so not something we should worry about much now.

jessie-ross:
The only "big" endpoint we might have is `/experiments` and if we split that into recent (within 3 months, should always be less than 100 experiments) and archived (greater than 3 months, Experimenters would be willing to wait longer for this) it should basically always be fine to use client-side filtering.

Currently we display 150 experiments, this comes to about 5kb.

Payload size is effectively negligible in an app like this:

<img width="1219" alt="Screen Shot 2021-04-14 at 2 34 09 pm" src="https://user-images.githubusercontent.com/971886/114664648-93e13b00-9d2e-11eb-8ab9-176cb07865fd.png">

Even if we increase the payload size 100 fold to 500kb, it should only add around `150ms` on my poor Aussie internet connection, that would be nothing compared to the unnecessarily large server processing time of `2.3s` (The queuing counts as well as it is waiting for the options request to go through).

I think that "ExperimentBare" and "MetricBare" have ended up being micro-optimisation, and I find they regularly impede development.

yanirs:
> I think that "ExperimentBare" and "MetricBare" have ended up being micro-optimisation, and I find they regularly impede development.

I agree that adding more fields at this stage probably won't hurt, but as `ExperimentFull` also includes a bunch of relations, I think we're likely to see a performance hit if we move to returning the full list and remove caching (as we're planning to do). That said, once we have better instrumentation in place (this month? :crossed_fingers:), we'd be able to make informed decisions about optimisations and avoid performance degradation.

aaronyan:
We can bundle the descriptions work (689-gh-Automattic/experimentation-platform) when adding descriptions, since we may need to adjust `ExperimentBare` and `MetricBare` definitions.

aaronyan:
This looks to be done!

PR #692, opened 2022-05-10
Enable searching by description in the experiment directory

razvanpapadopol:
Part of https://github.com/Automattic/abacus/issues/500

## How has this been tested?

  * Open an existing experiment and copy a part of the description
  * Go back to the experiment directory and search for it


aaronyan (approved):
LGTM! Tested using production APIs, and it works like a charm. Love that we can search for similar experiments like "a/b/n" and "engagement". 🎉 Nice work, @razvanpapadopol!

PR #694, opened 2022-05-11
Remove experiment debug view

jessie-ross:
**This PR removes Experiment debug view.**

I no longer use this view and our normal results view now contains all the info it contains and more - particularly with the addition of the strategy switcher.

## How has this been tested?


aaronyan:
Just to confirm, this is only removing debug mode when viewing the experiment details? As long as we are keeping debug mode for the other parts of the app (like metrics directory to add/ edit metrics), I'm cool with removing experiment details debug mode!

Briefly looking through the code, that seems to be the case! But I just wanted to confirm.

jessie-ross:
> Just to confirm, this is only removing debug mode when viewing the experiment details? As long as we are keeping debug mode for the other parts of the app (like metrics directory to add/ edit metrics), I'm cool with removing experiment details debug mode!
>
> Briefly looking through the code, that seems to be the case! But I just wanted to confirm.

Yep! It's the "Debug" tab.

Issue #673, opened 2022-04-08
Improve disclaimer for deployment recommendation


razvanpapadopol:
LE: this issue is being addressed in the following PRs:
1. https://github.com/Automattic/abacus/pull/699 to improve the deployment recommendation explainer
2. https://github.com/Automattic/abacus/pull/708 to highlight the primary metric by opening its details panel automatically

When displaying the deployment recommendation, we should try to clarify a couple of facts:
- what is the primary metric (specifically, for the current experiment)
- other metrics (secondary) were not taken into consideration when computing this recommendation so it's the responsibility of the experimenter to evaluate them when making a decision

<img width="483" alt="Screenshot 2022-04-08 at 10 48 43" src="https://user-images.githubusercontent.com/14192054/162390470-af478230-55bd-4e61-8edb-8d636cf67c7c.png">

aaronyan:
I'd be open to improving this experience! How were you thinking of making it more clear what is the primary metric for the experiment?

razvanpapadopol:
For the 1st point, I've tried to develop a method of programmatically expanding the first row of the Metric Assignment Results table but it's rather ugly code-wise. However, we can do small things like changing the label from "primary" to "primary metric" and maybe do something about the very transparent gray
<img width="672" alt="Screenshot 2022-05-13 at 21 10 45" src="https://user-images.githubusercontent.com/14192054/168342798-dafb46c2-57c5-466f-9fc2-976b9afe0f5b.png">

Secondly, I've been thinking to include a more detailed explanation/nudge as part of the deployment recommendation subtitle without taking so much real estate. Check out https://github.com/Automattic/abacus/pull/699

aaronyan:
> For the 1st point, I've tried to develop a method of programmatically expanding the first row of the Metric Assignment Results table but it's rather ugly code-wise.

Ooo good idea! It may be worth drafting, since I've had some experience with folks not knowing you can expand the metric values. If it's in a draft, then @jessie-ross might be able to give some input.

> However, we can do small things like changing the label from "primary" to "primary metric" and maybe do something about the very transparent grey color that we're using.

I'm down for adding "primary metric". Great idea to make it consistent throughout. What did you have in mind to replace the transparent grey?

> Secondly, I've been thinking to include a more detailed explanation/nudge as part of the deployment recommendation subtitle without taking so much real estate. Check out https://github.com/Automattic/abacus/pull/699 PR with a proposal.

Very cool! Thanks for coding that up and sharing!

[...]

PR #667, opened 2022-03-18
 Change 'Recommendation' to 'Analysis' in metrics assignment results

razvanpapadopol:
part of https://github.com/Automattic/abacus/issues/660

## Summary of changes
This PR is renaming 'Recommendation' to 'Analysis' in metrics assignment results table. This set of changes is focused mostly on UI and the props of the UI component which is now `AnalysisDisplay`. Types may be updated in a follow-up PR.

* Rename `RecommendationDisplay` file to `AnalysisDisplay`
* Update `ActualExperimentResults`
* Update `MetricAssignmentResults`
* Update experiment debug (`FullLatestAnalyses`)
* Update test snapshots

Screenshot 2022-03-19 at 01 34 32

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)
- Manual testing with mock data (locally or on Vercel)
- Additional manual testing instructions (please specify):

PR #699, opened 2022-05-13
Improve the 'primary metric' disclaimer and visibility

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/673 together with https://github.com/Automattic/abacus/pull/708
### Demo:

1. Add a tooltip on the deployment recommedation subtitle / explainer 
  Screenshot x

<details>
  <summary>Previous version with the Info icon</summary>
  Screenshot 2022-05-17 at 22 53 04
</details>

<details>
  <summary>Original version, before feedback</summary>
  Screenshot 2022-05-13 at 20 52 08
</details>

2. Update wording and color for the 'primary' tag on the Metric Assignment Results table to somewhat match with the explainer above
Screenshot 2022-05-20 at 19 18 09

## How has this been tested?


aaronyan (commented):
I like the direction! What do you think of these suggestions?
- Replace info icon with hoverable tooltip (like pp, see image below)

Screen Shot 2022-05-13 at 3 21 01 PM

If we prefer the info icon, then how about these?
- Reduce the size of the info icon
- Color info icon light grey

src/components/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -487,7 +494,16 @@ export default function ExperimentResults({
...
                         </Typography>
                         <Typography variant='subtitle1'>
-                          <strong>primary metric</strong> recommendation
+                          <span className={classes.summaryStatsSubtitle}>
+                            <strong>primary metric</strong> recommendation
+                          </span>
+                          <Tooltip
+                            className={classes.summaryStatsSubtitleInfoIcon}
+                            title='
+                            All the secondary metrics were not considered. Check the table below to analyze them individually.'
```
aaronyan, line 503: What do you think of this wording? "The recommendation is metric specific and does not consider other metrics. Check the table below to analyze each metric and its corresponding recommendation."

razvanpapadopol: I've included the new wording in the latest version (updated also the PR description) without this fragment:

> and its corresponding recommendation

I thought we won't be referring to the (partial) conclusions from the Analysis table as "recommendations" after https://github.com/Automattic/abacus/pull/667. Also, "analyze each metric" should be clear enough in my opinion.

razvanpapadopol:
Thanks for reviewing these changes and for the great suggestions, @aaronyan !

> If we prefer the info icon, then how about these?
>
> Reduce the size of the info icon
> Color info icon light grey

Included both in latest version and updated Demo screenshot ✅ 

> Replace info icon with hoverable tooltip 

Here is a demo with the hoverable tooltip and corresponding styles: 

Screenshot 2022-05-17 at 22 43 46

I'm ok with both, but in this case, I wanted to highlight somehow the additional explainer. Even a warning icon (but grey) could be interesting. @jessie-ross , WDYT?

src/components/experiments/single-view/results/ExperimentResults.tsx
aaronyan: Thanks for iterating on the tooltip text! Ah good call. Yeah, we should focus the text on the analysis rather than the recommendation. The current text looks great! Resolving this for now.

aaronyan:
Thanks for demoing the other hoverable tooltip, @razvanpapadopol! Both ways look great to me. I'm cool with going with the warning icon if we want to additionally emphasize and highlight the explanation.

src/components/experiments/single-view/results/ExperimentResults.tsx
razvanpapadopol: Thanks for the confirmation!

razvanpapadopol:
Thanks for supporting this direction, @aaronyan 🙏🏼 
I've included the warning icon and also two little things mentioned in https://github.com/Automattic/abacus/issues/673#issuecomment-1126552900 :
- updated wording for the 'primary' tag in the metrics table for consistency 
- made the tag a darker shade of grey for improved a11y

aaronyan:
Oh oops, I thought "warning icon" was the same as "info icon". 😅 I think the warning icon is a little bit too alarming; I prefer the info icon. :) Sorry about that confusion!

Thanks for updating all three in the PR description. Cool to see the progression and iteration.

jessie-ross:
Yeah info icon would be ok, but I'd say even no icon would be fine as this will constantly be around and I'd expect Experimenters to hover over it at some point.

razvanpapadopol:
> Yeah info icon would be ok, but I'd say even no icon would be fine as this will constantly be around and I'd expect Experimenters to hover over it at some point.

Yup, let's go with the "no icon" solution! The attention-grabbing change in https://github.com/Automattic/abacus/pull/708 should be enough for experimenters to understand the relation between the deployment recommendation and primary metric.

aaronyan (approved):
LGTM! I'm a fan of the simple and subtle addition!

PR #700, opened 2022-05-16
Refactor AnalysesNext further into ExperimentResults

jessie-ross:
**This PR makes the following changes:**
- Accept only analysisNext from API.
- Send analysesNext to experiment results.
- Rename previous analysis fixtures.
- Add AnalysesNext fixtures.
- Restrict ExperimentResults to AnalysisNext.

## How has this been tested?


razvanpapadopol (approved):
The refactor looks good and everything seems to compile and work as expected. 
Awesome work adding the new fixtures and updating the existing tests! 🙌🏼

src/api/AnalysesApi.ts
```diff
@@ -9,22 +9,14 @@ import { fetchApi } from './utils'
...
-    // istanbul ignore next; Transitional
-    Object.hasOwnProperty.call(maybeAnalysis, 'metric_estimates') &&
-    !!maybeAnalysis?.metric_estimates &&
-    Schemas.isRawMetricEstimatesNext(maybeAnalysis?.metric_estimates)
-      ? Schemas.analysisNextSchema.validateSync(maybeAnalysis)
-      : Schemas.analysisPreviousSchema.validateSync(maybeAnalysis),
-  )
-
-  return analyses
+async function findByExperimentId(experimentId: number): Promise<Schemas.AnalysisNext[]> {
```
razvanpapadopol, line 12: With the API returning only `AnalysisNext[]` we should probably update also in https://github.com/Automattic/abacus/blob/refactor/analyses-next/src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx#L252

PR #706, opened 2022-05-18
Update ExperimentResults stories

razvanpapadopol:
* Update `experiment` value used in existing stories to render the variations correctly
* Add "Treatment Wins" story
* Add "Control Barely Ahead" story
* Update A/b/n story with various `participantStats` values

## How has this been tested?

Run `npm run storybook` and check out the new stories:
- `/?path=/story/experiment-results--treatment-wins`
- `/?path=/story/experiment-results--control-barely-ahead`
- `/?path=/story/experiment-results--abn-experiment-analyses` - mostly for the Health Report


jessie-ross:
I'm curious what you think of Storybook, I haven't used it since I started I think, I almost always use `REACT_APP_PRODUCTION_CONFIG_IN_DEVELOPMENT=true npm run dev` and run with production data (some would call that the A8C way.)

jessie-ross (approved):
LGTM!

PR #713, opened 2022-05-25
Implement A/B/n Analysis Controls

jessie-ross:
**This PR implements A/B/n analysis controls.**
- Adds controls.
- Alters the A/B/n warning banner.
- Alters the MetricAssignment detail view to only show the two selected variations.

Screen Shot 2022-05-25 at 3 53 06 pm

It needs some touching up design-wise. I think we need to do something else to make it a bit clearer that all the stats on the page are relative to the select boxes 🤔 But I think that can happen in a later PR.

Resolves 661-gh-Automattic/experimentation-platform

## How has this been tested?


src/components/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -467,29 +491,70 @@ export default function ExperimentResults({
...
+                          .map((variation) => (
+                            <MenuItem key={variation.variationId} value={variation.variationId}>
+                              {variation.name}
+                            </MenuItem>
+                          ))}
+                      </Select>
+                      <FormHelperText>Change variation to measure against base.</FormHelperText>
+                    </FormControl>
+                  </div>
+                </Paper>
```
razvanpapadopol, line 556: I think the variation control is a good candidate for a separate component that gets a `baseVariationId`, optional `changeVariationId`, `variations` list and a callback. 

I'm thinking we could reuse it for some cases where we want (and it's possible from UI perspective) to display all the variations and let the possibility of changing just the baseline. Eg: in the Analysis tables where everything works fine on `master` after this got fixed: https://github.com/Automattic/abacus/pull/701#discussion_r874781147

src/components/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -187,13 +194,30 @@ export default function ExperimentResults({
...
+  // For A/B/n baseline and change to compare
+  const [baseVariationId, setBaseVariationId] = useState<number | undefined>(
+    () => experiment.variations.find((v) => v.isDefault)?.variationId,
+  )
+  const [changeVariationId, setChangeVariationId] = useState<number | undefined>(
+    () => experiment.variations.find((v) => !v.isDefault)?.variationId,
+  )
+  const onBaseVariationIdChange = (event: React.ChangeEvent<{ value: unknown }>) => {
+    const baseVariationId = event.target.value as number
+    setBaseVariationId(baseVariationId)
```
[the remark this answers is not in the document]
jessie-ross: I don't quite understand, shouldn't the baseVariationId always be set on a change (when a user selects an item with the select control)?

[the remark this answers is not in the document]
razvanpapadopol: Oh, sorry about this. I've misread the function we're calling here 🙈
Marking the comment as resolved.

razvanpapadopol:
Hey, @jessie-ross, I must say the controls are looking and working great even if it's still "just" an Alpha feature ❤️ 

After a lot of testing and comparing with production, I kind of miss having all the variants together in these 3 places:
* Analysis tables (when you open a metric assignment.) - with a "Baseline" (that [we may be able to control](https://github.com/Automattic/abacus/pull/713#discussion_r882796532)) and absolute and relative changes for the rest of the variants.
* Variation value charts
* Observed data
However, depending on what @aaronyan also thinks, we can probably come back to designing a solution for these later.

Then, I noticed a bug regarding the analysis conclusion for each metric. For the experiment  `20587`, if I select the base `treatment_b` and change `treatment_a`, for some metrics `control` is winning. And if I select the change `control`, for some metrics `treatment_a` is winning. Because of this, probably also the deployment recommendation (derived from the primary metric) isn't correct.

aaronyan:
Good catch on the analysis, @razvanpapadopol! I'm seeing the same thing when comparing `treatment_a` to `treatment_b`, and `treatment_b` should be winning.

> I kind of miss having all the variants together in these 3 places

I also prefer seeing all the variants together. I do see the value in doing the 1-1 comparisons though. Would it be possible to make the default view all variants, and then check a box to convert it to 1-1 comparisons?

Since the controls only change the metric assignments, it might be nice to move both the disclaimer and the controls below the participants by variation graph.

src/components/experiments/single-view/results/ExperimentResults.tsx
jessie-ross: That sounds like a good idea, did you think it should be separated out in this PR or just mentioning it for the future?

jessie-ross:
> Then, while doing lot of testing and comparing with production, I kind of miss having all the variants together in these 3 places

Yeah this was a design decision I made - the simplest first iteration was to take what we have and add the switcher.

The big issue with having a switcher like this is making sure experimenters are aware of it and know how it is affecting the data so if everything is switched and we don't have multiple at once I was hoping it might be more consistent and intuitive.

Next steps is to rethink the decision mechanism (maybe run a "tournament" of decisions across combinations of two variations and see what wins?), and then add in the table with all variations and then maybe add a table of graphs for every combo of variations.

> I've noticed only a bug regarding the analysis conclusion for each metric. For the experiment 20587, if I select the Base variation treatment_b and Change variation treatment_a, for some metrics control is winning. And if I select control as the Change variation, for some metrics treatment_a is winning. Because of this, probably also the deployment recommendation (derived from the primary metric) isn't correct.

Oooh thanks for catching that!

jessie-ross:
> I've noticed only a bug regarding the analysis conclusion for each metric. For the experiment 20587, if I select the Base variation treatment_b and Change variation treatment_a, for some metrics control is winning. And if I select control as the Change variation, for some metrics treatment_a is winning. Because of this, probably also the deployment recommendation (derived from the primary metric) isn't correct.

Fixed in [68ef1a6](https://github.com/Automattic/abacus/pull/713/commits/68ef1a64af18c7d8ebad41df0636bc30445efa46) 🙂

aaronyan:
> Next steps is to rethink the decision mechanism (maybe run a "tournament" of decisions across combinations of two variations and see what wins?)

Sounds good to me!

> maybe add a table of graphs for every combo of variations.

This sounds like a lot to me, but I'd be open to some mockups or prototypes!  If we're very interested in 1-1 comparisons, I'd prefer the comparison controls over a table of every combo of analyses and graphs.

razvanpapadopol (approved):
Thanks for addressing all the feedback, @jessie-ross! The code changes look good and so is the functionality. Here's a summary from my side:

### Working good now 🎉 
> The big issue with having a switcher like this is making sure experimenters are aware of it and know how it is affecting the data so if everything is switched and we don't have multiple at once I was hoping it might be more consistent and intuitive.

I'm good with the clarity provided by the Base/Change variations selector so until we get more input from experimenters, let's roll it out like this.

> Since the controls only change the metric assignments, it might be nice to move both the disclaimer and the controls below the participants by variation graph.

I don't think this is needed now because the top-right deployment recommendation for the primary metric works fine according to the two variations selected.

### Follow-up needed
Bring back the full A/B/n results, reordered (first Base, second Change and then the rest). See my proposal in https://github.com/Automattic/abacus/pull/726. 

This will address the notes shared in https://github.com/Automattic/abacus/pull/713#issuecomment-1138823731 and https://github.com/Automattic/abacus/pull/713#issuecomment-1139127154 while keeping the variations selector and relevant analysis conclusions that were introduced in this PR.

### Further enhancements
** we can track these in separate issues
- Extracting VariationControlGroup into a separate component (mentioned in https://github.com/Automattic/abacus/pull/713#discussion_r882796532)
- Variation value charts with multiple (selectable maybe) variations (mentioned in https://github.com/Automattic/abacus/pull/713#issuecomment-1138823731)
- Explore the Decision Tournament idea

src/components/experiments/single-view/results/ExperimentResults.tsx
razvanpapadopol: > possibility of changing just the baseline

After more playing around with this, I'd opt for keeping only 1 variation selector at the top of the page, exactly how you designed it. So I don't think it should be a priority to extract the component for now. 
Mentioned also in https://github.com/Automattic/abacus/pull/713#pullrequestreview-996564614

Issue #723, opened 2022-06-02
Allow search by metric definition


aaronyan:
We should allow search by metric definition in the metric directory. If an experimenter knows the Tracks event, but doesn't know the metric name in Abacus, this will significantly make it easier to find the right metric.

Looking at out [APIs and schemas](https://app.swaggerhub.com/apis/yanir/experiments/0.1.0#/management/get_metrics), it doesn't look like this would require a change to the API!

razvanpapadopol:
Yes, it looks like we have everything we need 🎉 
I've drafted the new functionality in https://github.com/Automattic/abacus/pull/725 using the same pattern as in the experiment directory search by description (https://github.com/Automattic/abacus/pull/692).

PR #725, opened 2022-06-03
Metrics table: allow search by metric params

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/723

* To make the params searchable, we render a hidden column using stringified event/revenue params.

## How has this been tested?

### Manual testing instructions (please specify):
- test by searching in the Metrics table for:
  - a Tracks event; eg: `wpcom_checklist_completed`
  - a prop value for a Tracks event; eg: `is_new_site` 
  - a product slug used in a Revenue metric; eg: `jetpack-search`


src/components/metrics/MetricsTable.tsx
```diff
@@ -49,6 +49,9 @@ const useMetricDetailStyles = makeStyles((theme: Theme) =>
   }),
 )
 
+const stringifyMetricParams = (metric: Metric): string =>
+  JSON.stringify(metric.parameterType === 'conversion' ? metric.eventParams : metric.revenueParams, null, 4)
```
jessie-ross, line 53: Just as an aside: I think having the metric params in separate columns was a bad idea, there should have just been `metric.params`.

src/components/metrics/MetricsTable.tsx
```diff
@@ -107,6 +104,15 @@ const MetricsTable = ({
 }): JSX.Element => {
   debug('MetricsTable#render')
 
+  const processedMetrics = useMemo(
```
jessie-ross, line 107: I think useMemo is a bit of a micro optimisation here. Even if we had 2000 metrics this would still be fast enough.

jessie-ross (commented):
Overall looks good, thanks for the quick PR! I'll approve when this is out of WIP

src/components/metrics/MetricsTable.tsx
razvanpapadopol: I'd leave it in just as good practice but you're definitely right since metrics are manually added, unlike generated daily analyses for example. Also, the current "processing" has linear complexity.

aaronyan (commented):
Thank you for drafting this so quickly, @razvanpapadopol! LGTM so far with one minor tweak. The column spacing looks funny after making a search:

Screen Shot 2022-06-06 at 3 16 12 PM

aaronyan:
Woo, thanks for the quick fix. LGTM!

razvanpapadopol:
Thanks, both, for the initial reviews! 
@jessie-ross, I think this is ready for final approval now so I've removed the `WIP` 🚀

jessie-ross (approved):
LGTM!

PR #726, opened 2022-06-06
Show full A/b/n data and analysis where possible

razvanpapadopol:
Follow up proposal to #713 

### Summary of changes
This PR adds the `variationsNotSelected` concept so metric assignment results for all the rest of the A/b/n variations are appended to the globally selected Base and Change variations in Analysis and Observed Data tables.

**Other**:
* Style the appended rows differently
* Add A/B/n Experiment story to ExperimentResults 

### Demo
The following screenshots are taken using the newly introduced story with `treatment1` selected as Base and `treatment2` as Change variation.

Screenshot 2022-06-06 at 15 05 48

Screenshot 2022-06-06 at 15 05 58

## How has this been tested?


aaronyan (commented):
This is awesome! I am a big fan of showing all the values, and this is a creative way to blend the analysis controls with showing all values.

For a future PR (or this one), would it be possible to keep the original order of the variants and the results? So we always see  the rows in the following order: `control`, `treatment_a`, and `treatment_b`, but the comparison becomes bold and the rest fades into they grey. I'm cool with addressing later if this feels out of scope for this PR!

aaronyan:
That looks perfect, @razvanpapadopol! Thanks for including the static row order and adjusting the color on A/B/n analysis controls change.

I just noticed that the result rows auto-collapse when changing the controls too. Would it be possible to leave opened results rows open on controls change?

jessie-ross (approved):
Nice small and effective change! LGTM!

Looks like the snapshots just need updating.

razvanpapadopol:
> That looks perfect, @razvanpapadopol! Thanks for including the static row order and adjusting the color on A/B/n analysis controls change.
> 
> I just noticed that the result rows auto-collapse when changing the controls too. Would it be possible to leave opened results rows open on controls change? This is not very important so if it takes a lot of work, I say we ship your changes just like this (just a nice quality of life thing)!

Thanks for having a look! Speaking of row auto-collapse, I'm pretty sure it's not related to this change and after some investigation, I've noticed it applies to changing the Analysis Strategy as well. It also doesn't have a quick fix so let's discuss further in https://github.com/Automattic/abacus/issues/731

> Nice small and effective change! LGTM!
> 
> Looks like the snapshots just need updating.

Thanks for reviewing this, @jessie-ross !
It was also a bit of coverage needed so I've fixed that in [a128c3](https://github.com/Automattic/abacus/pull/726/commits/a128c339181301712b00ef5e923ebf7cd52055d7) together with extraction of A/B/n fixtures and updating (fixing maybe?) [the other A/B/n related snapshot](https://github.com/Automattic/abacus/pull/726/commits/a128c339181301712b00ef5e923ebf7cd52055d7#diff-7f5e48288ae0bd806fe9491ab68b264d57079961ec8a9e957f617e07396be041). Since the functionality was left untouched, I'll go ahead and release 🚢

Issue #717, opened 2022-05-26
Adjust all references from wiki to FG page


aaronyan:
We recently finished migrating the wiki to the FG (pbmo2S-1lm-p2)! Let's now adjust [all the references in Abacus](https://github.com/Automattic/abacus/search?q=wiki) from the wiki to the new FG pages.

PR #729, opened 2022-06-08
Replace wiki links with FG links

aaronyan:
Fixes https://github.com/Automattic/abacus/issues/717.

This replaces all the wiki links with FieldGuide links and is the final step in the wiki to FG migration (696-gh-Automattic/experimentation-platform). 

The pages with wiki links were found using this search: https://github.com/Automattic/abacus/search?q=wiki


jessie-ross (approved):
Thanks for updating these! 🚀

src/components/experiments/wizard/Beginning.tsx
```diff
@@ -34,10 +34,23 @@ const Beginning = (): JSX.Element => {
...
+        </Link>{' '}
+        is a great place to start, it will instruct you on{' '}
+        <Link
+          underline='always'
+          href='https://fieldguide.automattic.com/the-experimentation-platform/document-experiment/'
+          target='_blank'
+        >
+          documenting your experiment
+        </Link>{' '}
+        and creating a P2 post.
```
razvanpapadopol, line 53: Love the added extra bits of guidance! ❤️

src/lib/analyses.test.ts
```diff
@@ -472,12 +471,9 @@ describe('getExperimentHealthIndicators', () => {
...
           },
-          "link": "https://github.com/Automattic/experimentation-platform/wiki/Experiment-Health#experiment-run-time",
+          "link": "https://fieldguide.automattic.com/the-experimentation-platform/experiment-health/#experiment-run-time",
           "name": "Experiment run time",
           "unit": "days",
-          "value": ${expectedExperimentRunTimeDays},
-        },
-      ]
-    `)
+          "value": `)
```
razvanpapadopol: This should fix the failing unit test:
```suggestion
          "value": ${experimentRunTimeDays},
        },
      ]
    `)
```

src/lib/analyses.test.ts
```diff
@@ -450,7 +450,6 @@ describe('getExperimentHealthIndicators', () => {
   it('should work for an experiment that ran too long', () => {
     const experimentRunTimeDays = 50
     const startDatetime = subDays(new Date(), experimentRunTimeDays)
-    const expectedExperimentRunTimeDays = differenceInHours(new Date(), startDatetime) / 24
```
razvanpapadopol, line 453: This was indeed not needed since we're explicitly defining the `experimentRunTimeDays` in the test setup.

razvanpapadopol (approved):
🚀

src/lib/analyses.test.ts
aaronyan: Thanks for the suggestion! I ended up adding this back in https://github.com/Automattic/abacus/pull/729/commits/bc3a8a44ddbc89e5f25f409ee55a5f00118631da.

PR #730, opened 2022-06-10
Add A/B/n Health Stats

jessie-ross:
**This PR adds A/B/n health stats:**
- Because we are changing the test to a different approximation, I've turned up the sample size and updated snapshots in the first commit to make it easier to see the changes ([view the snapshot update commit separately](https://github.com/Automattic/abacus/pull/730/commits/45a068a29e3385b1e6e2d46d2f1a02cec82a8345).)
- Added chi-squared-test library and added typescript declarations.
  This is a small stable library that has been around for 8+ years now, wrapping the chi-squared lib which is also small and stable, 8+ years old, and wraps the gamma lib also small stable and old.
- Replaces the binomial test with the chi-squared test. See inline comments.

## How has this been tested?


src/lib/analyses.ts
```diff
@@ -143,37 +141,44 @@ export function getExperimentParticipantStats(
...
-            }),
-            assignedNoSpammersNoCrossoversDistributionMatchingAllocated: binomialProbValue({
-              successfulTrials: variationCountsSet.assignedNoSpammersNoCrossovers,
-              totalTrials: participantCounts.total.assignedNoSpammersNoCrossovers,
-              probabilityOfSuccess: allocatedPercentage / totalAllocatedPercentage,
-            }),
-          },
-        ]
-      }),
+  // We use the Pearson Chi Squared Test, replacing the previous binomial test to allow for more than 2 variations.
```
jessie-ross, line 144: cc @demetd in case you want to take a look at the Pearson's Chi-Squared Test usage.

src/lib/analyses.test.ts
```diff
@@ -136,87 +136,78 @@ describe('getExperimentParticipantStats', () => {
         {
           [AnalysisStrategy.IttPure]: Fixtures.createAnalysis({
             participantStats: {
-              total: 130,
-              variation_1: 40,
-              variation_2: 70,
+              total: 11000,
```
aaronyan, line 139: Do we have low count tests too? It would be good to check to make sure the chi-square test works on low counts too (like less than 100). :)

src/lib/analyses.ts
```diff
@@ -143,37 +141,44 @@ export function getExperimentParticipantStats(
...
+  // The Pearson Chi Squared Test is recommended by Trustworthy Online Controlled Experiments (chapter 21: Sample Ratio Mismatch) and allows us to compare expected frequencies of categories against observed frequencies.
```
aaronyan, line 145: I couldn't find the recommendation to use the Pearson Chi Squared Test in the book. Happen to have more information on where we could find that (maybe page number)? 

A chi-square test seems like the right test to use though. In [this paper](https://exp-platform.com/Documents/2019_KDDFabijanGupchupFuptaOmhoverVermeerDmitriev.pdf) dedicated to SRM, they recommend chi-square tests too in 2.3.1 (where they reference [Pearson's original article from 1900](https://www.tandfonline.com/doi/abs/10.1080/14786440009463897), wow!).

src/utils/math.ts
```diff
@@ -39,3 +40,19 @@ export function binomialProbValue({
   const y = (successfulTrials - mean) / Math.sqrt(2 * variance)
   return 1 - erf(abs(y))
 }
+
+/**
+ * A slightly modified chiSquaredTest, see `chiSquaredTest` for more info:
+ *  - With all 0 observations and expectations, returns 1 (certianty).
```
aaronyan, line 46:
```suggestion
 *  - With all 0 observations and expectations, returns 1 (certainty).
```

src/utils/math.ts
```diff
@@ -39,3 +40,19 @@ export function binomialProbValue({
...
+ */
+export function chiSquaredTestProbValue(
+  observations: number[],
+  expectations: number[],
+  degreesOfFreedomReduction: number,
+): number {
+  if (observations.every((x) => x === 0) && expectations.every((x) => x === 0)) {
```
aaronyan, line 53: Should we also include a check for a [minimum of 5](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3900058/#:~:text=The%20value%20of%20the%20cell%20expecteds%20should%20be%205%20or%20more%20in%20at%20least%2080%25%20of%20the%20cells%2C%20and%20no%20cell%20should%20have%20an%20expected%20of%20less%20than%20one%20(3).) for each variant (both expected and observed)? We can be even more conservative and bump it up to like 100.

src/lib/analyses.test.ts
```diff
@@ -331,7 +323,117 @@ describe('getExperimentParticipantStatHealthIndicators', () => {
...
+            [AnalysisStrategy.IttPure]: Fixtures.createAnalysis({
+              participantStats: {
+                total: 11000,
+                variation_1: 5650,
+                variation_2: 5350,
+              },
+            }),
+            [AnalysisStrategy.MittNoCrossovers]: Fixtures.createAnalysis({
+              participantStats: {
+                total: 0,
```
aaronyan, line 349: Are these values zero because we're not checking SRM for `MittNoCrossovers` and `MittNoSpammers` (only the other three strategies)?

src/lib/analyses.ts
```diff
@@ -143,37 +141,44 @@ export function getExperimentParticipantStats(
...
+  const variationProportionProbabilities = {
+    assignedDistributionMatchingAllocated: chiSquaredTestProbValue(
+      experiment.variations
+        .map((variation) => participantCounts.byVariationId[variation.variationId])
+        .map((variationCounts) => variationCounts.assigned),
+      experiment.variations.map(
+        (variation) => (variation.allocatedPercentage / totalAllocatedPercentage) * participantCounts.total.assigned,
+      ),
+      1,
+    ),
```
aaronyan, line 156: I'm not the biggest fan of this repeating three times, but separate functions for observed and expected probably won't help too much

aaronyan (approved):
LGTM! Nice work, @jessie-ross! No blockers, but a few questions and suggestions.

I'd love to see comparisons between `binomialProbValue` and `chiSquaredTestProbValue` in a notebook with varying assignment ratios and sample sizes. It could make an interesting p2 post too!

src/lib/analyses.ts
jessie-ross: I only had a pdf of TOCE with me and it didn't have page numbers :( It's in chapter 21 though.

src/lib/analyses.test.ts
jessie-ross: Yep that's right 👍

src/lib/analyses.ts
jessie-ross: Yeah I think it would over-complicate things to try refactor this, particularly since the benefits are small (this code won't be used anywhere else, and the repetitions are colocated.) I'm a big fan of allowing duplication where it makes sense.

jessie-ross:
Thanks for resolving the conflicts here @razvanpapadopol!

src/utils/math.ts
jessie-ross: See https://github.com/Automattic/abacus/pull/730#issuecomment-1155278230

src/lib/analyses.test.ts
jessie-ross: See https://github.com/Automattic/abacus/pull/730#issuecomment-1155278230

PR #733, opened 2022-06-10
Manual Assignment: Fix COORS error

### LINKED ISSUE #727: Fix assignment bookmarklet CORS error
Experimenters are seeing a CORS error when they attempt to assign themselves an experiment variation in Abacus using the bookmarklet (p1654179042089569-slack-C7HH3V5AS). It gets resolved when sandboxing the API (p1654517167700969-slack-C7HH3V5AS), but we should fix it so experimenters don't have to jump through hoops to test their experiment. It may be coincidence, but this issue seems to have come up when Barry made a few changes to the API:

- Prevent cache stampede (pMz3w-fi5-p2#comment-98861)
- Rate limiting API (pMz3w-fi5-p2#comment-98875)

jessie-ross:
**This PR fixes the COORS error in manual assignment on Abacus.**

I am not sure what has changed but it looks like there is a redirect from `.../assignments` to `.../assignments/` and there is now a COORS restriction on redirects, perhaps a security related change.

In this PR I've added the trailing slash in so the redirect doesn't occur.

Resolves #727

## How has this been tested?

  Tested with and without the trailing slash on my local environment.


razvanpapadopol (approved):
Good catch, @jessie-ross ! LGTM 🚀

jessie-ross:
OOops it is CORS and not COORS

aaronyan:
Thanks for fixing this, Jessie! And for the quick review, Razvan.

> OOops it is CORS and not COORS

We can fix the COORS issue at the next meetup! 🍻

Issue #731, opened 2022-06-10
Metric Assignment Results: Preserve the rows open state


razvanpapadopol:
I just noticed that the result rows auto-collapse when changing the controls too. Would it be possible to leave opened results rows open on controls change? This is not very important so if it takes a lot of work, I say we ship your changes just like this (just a nice quality of life thing)!

_Originally posted by @aaronyan in https://github.com/Automattic/abacus/issues/726#issuecomment-1149275765_

After some investigation, it looks like this behavior also applies to changing the Analysis Strategy.

Related thread (no straight-forward solution): https://github.com/mbrn/material-table/issues/1021

aaronyan:
Thanks for moving this request to a separate issue, @razvanpapadopol! Too bad there's not a straight-forward fix.

PR #734, opened 2022-06-10
Experiment Results: update row open state handling

razvanpapadopol:
Fixes: https://github.com/Automattic/abacus/issues/731

This PR is addressing two main UX issues:
* Preserve the open state of the last active row when the A/B/n variation or the Analysis Strategy are updated
* Prevent the first metric from [automatically expanding](https://github.com/Automattic/abacus/pull/708) outside of the initial page load (eg: after any of the state updates mentioned above)

## How has this been tested?

  * Check if automatic expanding is prevented
  * Check if the opened state is persisted

### Additional manual testing instructions (please specify):
  * For any experiment, expand one of the rows and then change the Analysis Strategy. The last expanded row should be still open
  * For an A/B/n experiment, see the first row automatically expanded. Change Base/Change variations and see how the data updates while the row stays open
    * Close the first row to see many rows on the screen. Change Base/Change variations and the rows should stay closed


jessie-ross (approved):
LGTM! This seems like a cleaner solution than before.

(I haven't tested the UI of it.)

aaronyan (approved):
Yay, thanks for making this happen, @razvanpapadopol! LGTM. Tested locally too and works well!

PR #698, opened 2022-05-13
Update copy for the new recommendations

razvanpapadopol:
Add a couple of changes as per suggestions in https://github.com/Automattic/abacus/issues/660#issuecomment-1119167861
- Use present continuous when describing the 'winning' variant

Screenshot 2022-05-13 at 19 34 45

- Shorten the "not enough data" deployment recommendation

Screenshot 2022-05-13 at 19 35 24

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)
- Manual testing with mock data (locally or on Vercel)
- Additional manual testing instructions (please specify):

PR #736, opened 2022-06-21
Experiment results: update wording of the analysis conclusion

razvanpapadopol:
As a follow-up of #698, let's use present continuous when describing the 'winning' variant everywhere. 

Before this change, there were cases for which Abacus says that a variant "wins" when analyzing the primary metric but the deployment recommendation was "Not enough certainty" because `isDataStrongEnough` with all the other checks was still negative.

Screenshot 2022-06-21 at 15 05 04

## How has this been tested?


aaronyan (approved):
LGTM! Thanks for adding this quick win and making sure the recommendation matches the analysis. 🙌

Issue #709, opened 2022-05-20
Add 'copy to clipboard' button next to the experiment name


razvanpapadopol:
A feature that might be useful when having to implement code, share an experiment or simply search for an experiment is the possibility to copy its name easily.

<img width="1267" alt="Screenshot 2022-05-20 at 19 56 39" src="https://user-images.githubusercontent.com/14192054/169576998-2a419035-1d78-4518-be8f-4aa9676c3284.png">

Here's an example from wp.com, after publishing a post:
<img width="349" alt="Screenshot 2022-05-20 at 19 47 50" src="https://user-images.githubusercontent.com/14192054/169577014-c5142e23-bf0f-4b15-b535-ecc8ee335ac9.png">

### Acceptance criteria:
- [ ] Implement the button next to the title
- [ ] Make sure the button is displayed correctly also when the title is displayed truncated (eg: `exp_name_...`)
- [ ] Add a confirmation after copying such as modifying the button label to "Copied!"

#### Bonus item:
* Add a possibility to quickly copy the experiment name in the "Create an experiment" wizard (on the last step maybe)

aaronyan:
Yess! I love this idea. Good call on making sure it displays correctly even when truncating the title.

> Add a confirmation after copying such as modifying the button label to "Copied!"

Yes please on a confirmation like this. If we go with a copy icon, then maybe we can do a checkmark icon to confirm it's been copied. Something like these two?

- [CopyContent](https://mui.com/material-ui/material-icons/?query=copy&selected=ContentCopy)
- [AssignmentTurnedIn](https://mui.com/material-ui/material-icons/?query=checkmark&selected=AssignmentTurnedIn)

razvanpapadopol:
Woah, I haven't seen those until now! ✨  
The jiggly feedback is fun but I'd try also with a checkmark, as Aaron suggested, for extra clarity.

PR #740, opened 2022-06-24
Add 'Copy to clipboard' feature

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/709

- [x] 'Copy to clipboard' functionality 
- [x] Integration of the `ClipboardButton` with the experiment name
- [x] The button itself with its states (default and done)
- [x] Snackbar notifications for success/error 

## How has this been tested?


src/components/general/ClipboardButton.tsx
```diff
@@ -0,0 +1,23 @@
+import { Button } from '@material-ui/core'
+import { Assignment, AssignmentTurnedIn } from '@material-ui/icons'
+import Clipboard from 'clipboard'
+import React, { useEffect, useState } from 'react'
+
+export default function ClipboardButton({ text = '' }: { text?: string }): JSX.Element {
+  const [label, setLabel] = useState('Copy')
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
razvanpapadopol: Yes indeed. It works really well. Added in [a5ee138](https://github.com/Automattic/abacus/pull/740/commits/a5ee1383430265145aff3a25a0d994b3ee45ba38)
Thanks for the hint! 🥇

package.json
```diff
@@ -81,6 +81,7 @@
     "@typescript-eslint/parser": "^4.2.0",
     "babel-loader": "^8.1.0",
     "babel-plugin-remove-debug": "0.0.4",
+    "clipboard": "^2.0.11",
```
jessie-ross, line 84: I think we should use native API for this instead of adding a library:
https://www.w3.org/TR/clipboard-apis/#dom-clipboard-write

We can rely on Experimenters having modern browsers :)

razvanpapadopol: We have this library as a dependency also [in wp-calypso](https://github.com/Automattic/wp-calypso/blob/trunk/client/package.json#L101) for a similar [Clipboard component](https://github.com/Automattic/wp-calypso/blob/trunk/client/components/forms/clipboard-button/index.jsx). If migrated to [@automattic/components ](https://github.com/Automattic/wp-calypso/tree/trunk/packages/components) we could've used it directly as a React component 😁 

I was considering using the native API for simplicity but I thought the "Security context" requirement is a problem for local development.
> Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS)
*[See it [mentioned here](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard) at the top of the page]

💡 However, I've just discovered that we could use the flag `--unsafely-treat-insecure-origin-as-secure` to 'whitelist' Abacus local URL. It is a bit of a trade-off IMO but I'm ok with it since it's just when testing this type of functionality. WDYT?

jessie-ross: > We have this library as a dependency also [in wp-calypso](https://github.com/Automattic/wp-calypso/blob/trunk/client/package.json#L101) for a similar [Clipboard component](https://github.com/Automattic/wp-calypso/blob/trunk/client/components/forms/clipboard-button/index.jsx). If migrated to [@automattic/components ](https://github.com/Automattic/wp-calypso/tree/trunk/packages/components)we could've used it directly as a React component 😁

Yeah that makes sense if that's something you'd like to do.

> 💡 However, I've just discovered that we could use the flag --unsafely-treat-insecure-origin-as-secure to 'whitelist' Abacus local URL. It is a bit of a trade-off IMO but I'm ok with it since it's just when testing this type of functionality. WDYT? Or do you know better (safer) solutions?

I think it would be ok not having this feature available when running locally 🤷 

The tradeoff here is not having an extra dep, which I think is very important in terms of keeping Abacus as lean as possible and making upgrades as simple as possible.

razvanpapadopol: Switched to using the native API in [27d6c1d](https://github.com/Automattic/abacus/pull/740/commits/27d6c1d1ffc981142179c9f8b77e46d2ce552350)
Works well, including testing on local, with the [Chrome flag](https://github.com/Automattic/abacus/pull/740#discussion_r908715048)

src/components/general/ClipboardButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const setClipboard = () => {
+    navigator.clipboard
+      .writeText(text)
+      .then(() => {
+        setWasTextCopied(true)
+        enqueueSnackbar(`${targetName} copied!`, { variant: 'success' })
+        return
+      })
+      .catch(() => enqueueSnackbar('Copying failed!', { variant: 'error' }))
```
jessie-ross, line 24: Let's also log the error to console, and maybe include the error message here.

jessie-ross (approved):
This is looking good!

Issue #712, opened 2022-05-24
Prevent experiment creation with over 12 month duration


aaronyan:
Abacus prevents experiments from being displayed if an experiment has a [duration > 12 months](https://github.com/Automattic/abacus/blob/d7b8c5fef81c51df31afa75873f9c005cc7184e0/src/lib/schemas.ts#L413). We should prevent experiments from being created that have a duration over 12 months! :)

Potential action items:

- [x] Prevent experiments with over 12 month duration from being created.
*Possible N/A. See https://github.com/Automattic/abacus/issues/712#issuecomment-1170234174
- [x] Remove the limit when displaying experiments, so experiments can always be displayed (in case an advanced experimenter manually creates an experiment not in Abacus).
*Fixed in https://github.com/Automattic/abacus/pull/741

aaronyan:
Thanks for posting the UI validation, @razvanpapadopol. I'm going to check off the first item for now; we'll keep this in mind if we happen to see an experiment created with over 12 months through Abacus.

aaronyan:
This looks good to close!

PR #741, opened 2022-06-29
Enable displaying longer experiments

razvanpapadopol:
Part of https://github.com/Automattic/abacus/issues/712

>  Remove the limit when displaying experiments, so experiments can always be displayed (in case an advanced experimenter manually creates an experiment not in Abacus).

## How has this been tested?


jessie-ross (approved):
Thanks for fixing this!

In general there are probably other things that we can move from the FullNew schemas to the Full schemas.

PR #746, opened 2022-07-11
Add Min Diff Calculator

jessie-ross:
**This PR adds a min practical difference calculator to our metric creation wizard.**

Part of 423-gh-Automattic/experimentation-platform


jessie-ross:
We need to specify what kind of revenue we are asking for, should we ask for revenue per month or per year:

Screen Shot 2022-07-21 at 3 43 24 pm

In the case of per month we will have to annualise the figure across subscribers to get ARPU.

I'm guessing per month is what we want to ask for here

jessie-ross:
Ah, I thought ARPU was per year but it actually doesn't specify a unit of time, do we use it per month?

jessie-ross:
**Update:**
- Added tests.
- Disabled the sample size inverse norm as it wasn't importing correctly for some reason.
- Cleaned up the language a bit.
- Turned the NaNs to zeros and added rounding.

This is ready for review.

aaronyan:
> We need to specify what kind of revenue we are asking for, should we ask for revenue per month or per year

I like asking for revenue per month, since it matches with # of participants we're asking.

> Ah, I thought ARPU was per year but it actually doesn't specify a unit of time, do we use it per month?

Yeah, there's no unit of time for ARPU and has units $ per user. ARPU = (revenue per n time) / (number of participants per n time). The n time cancels out; we just need to make sure that # of participants and the revenue ask for the same time unit.

jessie-ross:
Yeah absolute revenue values are really weird as we run everything on a subscriber basis, I think we should really be using the ARPUPM (Average revenue per user per month) everywhere. Otherwise it is really ARPU over the experimentation period.

Anyway I have changed some language referring to revenue, let me know what you think 🙂

aaronyan:
The language for both conversion and revenue metrics looks really good! I'm a fan of the simple yet descriptive language.

src/components/experiments/wizard/MinDiffCalculator.tsx
```diff
@@ -0,0 +1,378 @@
...
+    },
+  }),
+)
+
+function nanToZero(x: number): number {
+  return isNaN(x) ? 0 : x
+}
+
+function roundToNumPlaces(x: number, nPlaces: number): number {
+  return Math.round(x * Math.pow(10, nPlaces)) * Math.pow(10, -nPlaces)
```
razvanpapadopol, line 56:
```suggestion
  return Math.round(x * Math.pow(10, nPlaces)) / Math.pow(10, nPlaces)
```

Because of JS precision issues with small numbers. 
Eg: `Math.pow(10, -4)` returns 0.0000999...

src/components/experiments/wizard/MinDiffCalculator.tsx
```diff
@@ -0,0 +1,378 @@
...
+                  id={`extra-revenue-${calculatorUniqueId}`}
+                  value={metricRevenuePerMonthMinPracticalDiff}
+                  onChange={onChangeMetricRevenuePerMonthMinPracticalDiff}
+                  label='Extra revenue / month'
+                  placeholder='1500'
+                  variant='outlined'
+                  fullWidth
+                  required
+                  type='number'
+                  helperText={`= ${roundToNumPlaces(nanToZero(metricMinimumPracticalDifferenceLift), 4)}% lift`}
```
razvanpapadopol, line 272:
```suggestion
                  helperText={`= ${roundToNumPlaces(nanToZero(metricMinimumPracticalDifferenceLift), 2)}% lift`}
```
2 decimals for the Revenue lift should be more than enough 😁

src/components/experiments/wizard/Metrics.tsx
```diff
@@ -400,69 +463,110 @@ const Metrics = ({
...
+                                name={`experiment.metricAssignments[${index}].minDifference`}
+                                id={`experiment.metricAssignments[${index}].minDifference`}
+                                metricParameterType={indexedMetrics[metricAssignment.metricId].parameterType}
+                              />
+                              <ToggleButton
+                                value='check'
+                                selected={activeMinDiffCalculatorList[index]}
+                                onChange={handleMinDiffCalculatorToggle.bind(null, index)}
+                                className={classes.minDifferenceCalculatorToggle}
+                                title='Min Difference Calculator'
```
razvanpapadopol, line 534:
```suggestion
                                title='Minimum Difference Calculator'
```

razvanpapadopol (approved):
The new feature and its implementation look great, @jessie-ross 🚀 
Thanks for the added tweaks, cleanup, and tests!

I've noticed just an issue with the precision of Revenue hints and left also a couple of minor suggestions. They can be all committed straight away before merging.

Here are also some ideas for potential follow-ups:
- improve the UX of calculator input when typing/pasting values. Currently, the default `0` feels a bit in the way
- For some reason, I don't find the `ArrowDropDown` icon descriptive enough. Maybe we try the `CalculateIcon` once we migrate to the latest MUI version.
- Add tests to the `MinDiffCalculator` component itself

src/components/experiments/wizard/MinDiffCalculator.tsx
jessie-ross: Ah nice catch! I was wondering why I was getting some weirdness after rounding!

src/components/experiments/wizard/Metrics.tsx
razvanpapadopol: Oops, the test needs an update because of this:
https://github.com/Automattic/abacus/pull/746/files#diff-73dbfe1b22a94de480885abb5ea9134f65519605ec205da95f80cb3ad69ca57cR94

PR #751, opened 2022-07-27
Health Report: add platform exception for Assignment distribution indicators

razvanpapadopol:
Related to https://github.com/Automattic/experimentation-platform/issues/751#issuecomment-1195985251
Fixes https://github.com/Automattic/abacus/issues/786

* Stop displaying Assignment distribution health check indicators for `Mlsales` platform.
* Don't take into consideration these indicators for the top-right health check overview.
* Include `Mlsales` platform to the whitelist so minimum runtime is ignored when displaying a deployment recommendation.

#### Demo (using 20827-explat-experiment)

* Before
Screen Shot 2022-10-01 at 00 47 22

* After
Screen Shot 2022-10-01 at 00 51 44

## Follow up idea:
- [ ] Extract existing Platform related customizations to a single config object

## How has this been tested?

- Manual testing with production data (use experiment `20720`):
     * With these changes applied, health indicators "Assignment distribution" and "Assignment distribution without crossovers and spammers" should be hidden from the Health Report of experiments running on `email` platform (to be replaced with `mlsales`) 
    * Also, the two indicators won't be taken into consideration when displaying the general health status at top-right of Experiment Results screen.


aaronyan (commented):
LGTM! Thanks for the making the issue changes and including additional improvements like runtime whitelist.

src/lib/recommendations.ts
```diff
@@ -156,7 +156,7 @@ const maxSafeKruschke = {
 }
 
 const minSafeRuntimeInDays = 7
-export const runtimeWhitelistedPlatforms = [Platform.Email, Platform.Pipe]
```
jessie-ross, line 159: I think this should be in `analyses.ts` as that is where we calculate the health indicators, maybe `minSafeRuntime` too.

jessie-ross (approved):
LGTM! (See comment.)

PR #757, opened 2022-08-08
Add four new supported platforms

aaronyan:
See 751-gh-Automattic/experimentation-platform. Adds four new supported platform to Abacus:

- `mlsales`
- `tumblr`
- `woocommerceandroid`
- `woocommerceios`

## How has this been tested?


bperson (approved):

src/lib/schemas.ts
```diff
@@ -331,13 +331,17 @@ export interface Variation extends yup.InferType<typeof variationSchema> {}
...
-  DayOne = 'dayone',
+  Dayone = 'dayone',
   Email = 'email',
   Jetpack = 'jetpack',
   Lohp = 'lohp',
+  Mlsales = 'mlsales',
   Pipe = 'pipe',
+  Tumblr = 'tumblr',
   Wccom = 'wccom',
-  WooCommerce = 'woocommerce',
```
jessie-ross, line 340: Hmm, any reason we are removing the CamelCase?

aaronyan: No strong reason. I was thinking it would help simplify the enums by only having the first letter capitalized. 

I wasn't sure if we should do `WooCommerceAndroid` or `WooCommerceandroid` since we already have `Wpandroid`. If we do `WooCommerceAndroid`, then should it be `WPAndroid`? 😅

jessie-ross: Since these are Typescript enums it is very easy to change these, I think what matters most here is readability.

I'll maybe one day change these, no need to change them now 🙂

aaronyan: Sounds good!

Issue #742, opened 2022-06-29
Change default tabs for non-staging experiments


aaronyan:
When clicking an experiment from the experiment directory, we always default to the Overview tab. What do y'all think about defaulting to a Overview tab for `staging` experiments and Results tab for all other statuses?
- `staging` - Overview tab
-  `running` - Results tab
- `completed` - Results tab
- `disabled` - Results tab

jessie-ross:
Yep! I love this one! I had this as a fleeting thought a while ago.

In fact, I was thinking we could even integrate the two, maybe with an summary/detailed switchable overview.

razvanpapadopol:
Yes, I don't see a reason why we shouldn't do this. I was also working to design a solution to show the description in the experiment directory (second point of https://github.com/Automattic/abacus/issues/500) that would help when identifying an experiment before clicking.

aaronyan:
Love all the ideas for improving usability around the overview and results. I'm open to iteration!

PR #766, opened 2022-08-18
Use Results as the default tab for non-staging experiments

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/742

When landing from the experiment directory or by using a `/experiments/{experimentId}` link, the default tab (`view`) should be:
 - 'Overview' for staging experiments
 - 'Results' for all other statuses
 
*This change is also somewhat related to 604-gh-Automattic/experimentation-platform (see comments there).
 
**Note:**
Before closing the original issue, we should extract thoughts from https://github.com/Automattic/abacus/issues/742 to new issues.
  
## How has this been tested?


aaronyan (approved):
LGTM! Works on my local machine too as expected.

aaronyan (approved):
LGTM! Works on my local machine too as expected.

src/components/experiments/single-view/ExperimentPageView.tsx
```diff
@@ -139,6 +139,11 @@ export default function ExperimentPageView({
 
   const experimentIdSlug = createIdSlug(experimentId, experiment?.name || '')
 
+  // When landing from experiment directory or by using a '/experiments/{experimentId}' link, the initial view is:
+  // - 'Overview' for staging experiments
+  // - 'Results' for all other statuses
+  const view = viewFromPath || (!!experiment && (canEditInWizard ? ExperimentView.Overview : ExperimentView.Results))
```
jessie-ross, line 145: Can we change this to redirect to the view if the view doesn't exist, I think that's a cleaner solution and will mean the URL will always reference the same page:
```
if (!view) {
    history.replace(`${location.pathname}/${(!!experiment && (canEditInWizard ? ExperimentView.Overview : ExperimentView.Results)}`)
}
```
Probably in a useEffect.

jessie-ross (changes requested):

src/components/experiments/single-view/ExperimentPageView.tsx
razvanpapadopol: Thanks for the suggestion, @jessie-ross ! Implemented in https://github.com/Automattic/abacus/pull/766/commits/a3abee5a9c85d4852f2fcf0a557910be667cd19e together with a bit of logic cleanup and it works really well 🙌🏼

jessie-ross (approved):
Thanks for adding the redirect in!

Issue #653, opened 2022-02-04
Change health report colors


aaronyan:
With the introduction of new recommendations (690-gh-Automattic/experimentation-platform), I think we should reserve green and yellow to mean "actionable" / "inactionable". What do y'all think about changing the health report colors? Making the fields with potential issues more orange will also match the orange in the top-level report.

- Nominal - grey (#dadada4f)
- Potential issue - more orange (#fedb9d)

## Current
<img width="1250" alt="Screen Shot 2022-02-04 at 2 15 25 PM" src="https://user-images.githubusercontent.com/4505888/152610570-d9ff5f31-b6c9-407e-bfcc-32314b8324c5.png">

<img width="370" alt="Screen Shot 2022-02-04 at 2 24 15 PM" src="https://user-images.githubusercontent.com/4505888/152611282-80424098-8efa-4503-8f65-137f8688c309.png">

## Proposal
<img width="1244" alt="Screen Shot 2022-02-04 at 2 27 57 PM" src="https://user-images.githubusercontent.com/4505888/152611630-d038f46e-306b-4069-a969-744fab1266f7.png">

jessie-ross:
This is much better!

PR #767, opened 2022-08-19
Limit metric value of type 'count' to 2 decimals

razvanpapadopol:
At the moment, we display Observed Data conversion values with 3 decimals (mentioned in p10gg3-aFD-p2) because `count` formatter function is expecting an integer (not explicitly, but it can be implied from the lack of processing).

The current change will update the `count` formatter to apply to apply rounding using the global `metricValueFormatPrecision` (two decimals). 
This will leave unaffected other `count` values, that are integers.

**Before**
Screen Shot 2022-08-23 at 11 47 53

**After**
Screen Shot 2022-08-23 at 11 46 25

Alternatively, the same could've been achieved by removing the decimals when the metric type is `Conversion` here: 

https://github.com/Automattic/abacus/blob/5dccd876c913aebf6756ba3689faefd9cb04ca9a/src/components/experiments/single-view/results/MetricAssignmentResults.tsx#L624-L627


aaronyan (commented):
Thanks for trying to be proactive about this @razvanpapadopol. My guess is this was inspired by this [comment](759-gh-Automattic/experimentation-platform#issuecomment-1205614483)?

The decimals are kind of useful right now to communicate to experimenters that these are posterior means. If we prefer not to have decimals in the observed data, then I think introducing the observed values back into Abacus is the move (as Jessie [suggested](759-gh-Automattic/experimentation-platform#issuecomment-1206003149)).

Even if we reduce the prior to be `0.001`, the rounded posterior mean could return a value that is not equal to the actual observed results. So it is probably better to simply return the observed results instead of rounding, changing prior, etc. 

To move forward with how we will deal with observed values, I think we should add them back in. What do y'all think?

razvanpapadopol:
> Thanks for trying to be proactive about this @razvanpapadopol. My guess is this was inspired by this [comment](759-gh-Automattic/experimentation-platform#issuecomment-1205614483)?
> 
> The decimals are kind of useful right now to communicate to experimenters that these are posterior means.

No, it was just an UI attempt to reduce the confusion when seeing for the first time in the app a "Conversion" value with 3 decimals. Combined with the comma-separated thousands, it's a bit hard to read as it can be seen also in the last screenshot included in your comment [here](pbmo2S-1fH-p2#comment-2311).

Now I've noticed this was mentioned as well previously and I agree it makes sense to have them.

> Conversion counts can now have decimals.

WDYT of aligning, for now, with the other common values (conversion rate, differences, min diff) by rounding to 2 decimals?

Screen Shot 2022-08-20 at 02 46 26

aaronyan:
> WDYT of aligning, for now, with the other common values (conversion rate, differences, min diff) by rounding to 2 decimals?

Great idea, that sounds good to me.

aaronyan (approved):
LGTM! I'm a fan of the rounding to two decimals and the addition of the `*` character! :)

jessie-ross (approved):
LGTM! Can we also make the asterisk red?

razvanpapadopol:
> LGTM! Can we also make the asterisk red?

Done! I've first tried to use the "potential issue orange" (`paletter.warning.main`) mentioned in https://github.com/Automattic/abacus/issues/653 but I think we need a darker color in order to be visible on the gray background here so red looks good.

Issue #710, opened 2022-05-20
Add absolute impact credible intervals


aaronyan:
What do we think about adding **absolute impact** credible intervals to the experiment results page like "# of users / month" and "$ in revenue / month"? This was originally suggested in this comment (pbmo2S-UZ-p2#comment-2110) and then analyzed with this method in this comment (pbxNRc-1qR-p2#comment-3353). 

In addition to the first two stats, let's consider adding a third stat "absolute impact" (we can iterate on the name!):
- Absolute change: [ -0.02pp, +0.07pp ]
- Relative change: [ -1.63%, +5.04% ]
- Absolute impact: [ -188, +659 ] / month or [ -4,036, +14,149 ] / year

Design questions:
- [ ] Where do we add this "absolute impact"? Just another column in the analysis table? Replace "absolute change"?
- [ ] Do we allow toggle-able date ranges (like per month vs per year)?

aaronyan:
> Where do we add this "absolute impact"? Just another column in the analysis table? Replace "absolute change"?

I like @razvanpapadopol's proposal to replace absolute change with this absolute impact credible interval. I'd even recommend naming it to a simple 'Impact' to hint at "impact analysis".

I think we can also swap the order of the columns: name, relative change, impact, analysis. I think that order and progression of information helps tell a story:

1. What is the name of metric being analyzed?
2. What are the broad results (% lift)?
3. What kind of impact does that have (# of users, $)?
4. What's the analysis?

jessie-ross:
Ah I think this should be another row in the intervals table rather than altering the metric assignments table, reason being that impact figures are a bit more fluffy and require a bit more understanding for interpretation than absolute change.

aaronyan:
Mmm for sure. I can see how the estimated effects of impact can be more fluffy, and I'm down to add it as another row in the intervals table for the details of a metric!

For the metrics overview, I think the impact could be really nice as a column to help bring context to the relative change. Something like a relative change of +200% but has an impact of +10 more users is nice to know from the start imo!

PR #772, opened 2022-08-24
Experiment results: replace Relative change with Estimated impact

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/710

### Summary of changes:
* Adds impact value above the relative change, as described in https://github.com/Automattic/abacus/issues/780
* Adds an inline Select for impact interval in the table header of the _Estimated impact_ column.
  * The interval selector defaults to "Year"
* Adds 'conversion_impact' and 'revenue_impact' types of metric formatting. 
  * **Note:** These will be cleaned up in https://github.com/Automattic/abacus/pull/774, as [detailed below](https://github.com/Automattic/abacus/pull/772#discussion_r1000750009).
  * `impactNumberFormatter` function transforms millions (eg: `2.51M`) and thousands (eg: `1.25M`) as seen on P2 updates like pbxNRc-1qR-p2#comment-3353
  * Good to test with 20571-explat-experiment and `pipe_*` experiments.

### Demo:
Screen Shot 2022-10-20 at 21 51 56

<details>
  <summary>Previous Demo</summary>

* See how the left side of the interval is saying much more than the 'Absolute change' does
Screen Shot 2022-08-24 at 19 21 47

* Another example with more context and a transformation for millions
Screen Shot 2022-08-24 at 19 20 18

* In very rare cases (conversion rate values with 6+ digits), the table header might be displayed on 2 lines
Screen Shot 2022-08-24 at 19 19 25
</details>

### To do:
- [x] Discuss if we should show the impact column only after a minimum runtime or volume of analyzed participants
- [x] Add a ready-to-share description phrase to the Summary section of metric details
- [x] Test coverage

### Follow up:
- [ ] Update related FG pages
- [ ] Clean-up `MetricValue` and formatters https://github.com/Automattic/abacus/pull/774

### How has this been tested?


aaronyan (commented):
This is looking really great so far, @razvanpapadopol. Nice work and great attention to detail with the thousands (K) and millions (M) and the dropdown impact time period!

Being able to report absolute impact as a function of monthly conversions / USD is so cool. You're right, the 20571-explat-experiment is a great example!

> In very rare cases (conversion rate values with 6+ digits), the table header might be displayed on 2 lines

If everyone agrees, I am totally cool with replacing 'Absolute change' column with 'Impact'. I don't think we need both!

> Discuss if we should show the impact column only after a minimum runtime or volume of analyzed participants

I think it is okay to show the impact right away, since high uncertainty will be reflected as a wide interval. I'm curious to hear y'alls opinion!
> Add a ready-to-share description phrase to the Summary section of metric details

I like that a lot! I think we could simply replace the 'absolute change' values with 'absolute impact' values.

jessie-ross:
Nice work Razvan!! Can't wait for this to go out to experimenters!

> If everyone agrees, I am totally cool with replacing 'Absolute change' column with 'Impact'. I don't think we need both!

I think we should keep both, as impact makes some assumptions that absolute change doesn't make.
On that note I think we need to downgrade the certainty of Impact, I think we should rename it "Guessed Impact".

> I think it is okay to show the impact right away, since high uncertainty will be reflected as a wide interval. I'm curious to hear y'alls opinion!

👍 

> I like that a lot! I think we could simply replace the 'absolute change' values with 'absolute impact' values.

I think lift figures are still important for experimenters and marketing teams.

src/components/general/MetricValue.tsx
```diff
@@ -71,16 +88,25 @@ export const metricValueFormatData: Record<string, MetricValueFormat> = {
     transform: identity,
     formatter: usdFormatter,
   },
+  revenue_impact: {
```
jessie-ross, line 91: Revenue impact and revenue are the same unit: the count of dollars. 
We should join the `usdFormatter` and the `impactNumberFormatter` together.

aaronyan:
> I think we should keep both, as impact makes some assumptions that absolute change doesn't make.

I don't think I've seen anyone use 'absolute change' in their analysis, and I think showing both feels redundant. I'm cool with keeping it for now in this PR, and we can always iterate in the future.

> On that note I think we need to downgrade the certainty of Impact, I think we should rename it "Guessed Impact".

I love that! How about 'estimated impact'?

> > > Add a ready-to-share description phrase to the Summary section of metric details
> >
> > I like that a lot! I think we could simply replace the 'absolute change' values with 'absolute impact' values.
>
> I think lift figures are still important for experimenters and marketing teams.

@jessie-ross, are you thinking to add a lift sentence to summary?

Screen Shot 2022-08-25 at 10 21 20 AM

I think @razvanpapadopol was hoping to add a sentence to the summary section of a metric details? I was thinking we could replace 'absolute change' with 'impact', so the sentence would read:

"The estimated monthly impact in ARPU of -71k to +12k USD is not statistically different from zero because the interval includes zero. Uncertainty is too high because the absolute change overlaps with the monthly minimum practical difference between -55k to +55k USD."

jessie-ross:
> are you thinking to add a lift sentence to summary?

Sorry I was confusing relative and absolute change here, I thinking we were replacing the lift with impact, instead of absolute change with impact.

I think absolute figures are very important, they are what Experimenters should be getting to know more than lift (which  they can't get to know as it changes depending on the absolute figures.)

razvanpapadopol:
> downgrade the certainty of Impact, I think we should rename it "Guessed Impact".

Yes, definitely! It's not about guaranteed or even projected impact but rather a synthetic extrapolation of the results.

> I don't think I've seen anyone use 'absolute change' in their analysis, and I think showing both feels redundant. I'm cool with keeping it for now in this PR, and we can always iterate in the future.

Yes, let's keep them for now as the UI is able to fit everything and we'll just keep an eye on how experimenters are using the values when sharing the following experiment results.

> I think absolute figures are very important, they are what Experimenters should be getting to know more than lift (which they can't get to know as it changes depending on the absolute figures.)

I agree we should try to keep the absolute diff visible as they are also connected to min diff and together are determining the analysis conclusion for a metric.

src/lib/analyses.ts
```diff
@@ -523,3 +523,13 @@ export function getExperimentHealthIndicators(experiment: ExperimentFull): Healt
 export function ratioToDifferenceRatio(ratio: number): number {
   return ratio - 1
 }
+
+export enum ImpactInterval {
```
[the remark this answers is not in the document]
razvanpapadopol: Done in https://github.com/Automattic/abacus/pull/772/commits/7640acd85eb5c47122bb0ce7bd67bdd2ec045fbb

src/components/general/MetricValue.tsx
```diff
@@ -57,6 +67,13 @@ export const metricValueFormatData: Record<string, MetricValueFormat> = {
     transform: (x: number): number => x * 100,
     formatter: standardNumberFormatter,
   },
+  conversion_impact: {
+    unit: '',
```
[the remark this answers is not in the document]
razvanpapadopol: I can see it's a bit difficult to navigate the fine line between metric type and units. Then, the formatting is another layer 😁 Let's discuss more on this (and some other comments below) in https://github.com/Automattic/abacus/pull/774

src/lib/analyses.ts
```diff
@@ -524,12 +524,11 @@ export function ratioToDifferenceRatio(ratio: number): number {
...
 
-export enum ImpactInterval {
-  Month = 'month',
-  Year = 'year',
-}
-
-export const ImpactIntervalInMonths: Record<ImpactInterval, number> = {
-  [ImpactInterval.Month]: 1,
-  [ImpactInterval.Year]: 12,
+export function getImpactMultiplier(
```
jessie-ross, line 527: - This should be called `estimateTotalParticipantsInPeriod` (which is what it does) rather than introducing the new terminology of `impactMultiplier`.
- I think it should work in days to make it general purpose.
- Type should be `(analysis, experiment, periodInDays) => number`
- We would need to look at the variation allocation percentages.
- Needs function documentation.

src/components/experiments/single-view/results/ImpactIntervalSelector.tsx
```diff
@@ -0,0 +1,40 @@
+import { MenuItem, Select } from '@material-ui/core'
+import _ from 'lodash'
+import React from 'react'
+
+export enum ImpactInterval {
+  Month = 'month',
+  Year = 'year',
+}
+
+export const ImpactIntervalInMonths: Record<ImpactInterval, number> = {
```
jessie-ross, line 10: I don't like how these live here - it is tying the analysis code too hard to the UI, really this is just a time interval selector where we specify labels for certain intervals.

We should be able to pass in `(months, onSetMonths: (months: number) => void)` instead.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -385,6 +388,11 @@ export default function MetricAssignmentResults({
   const isVariationSelected = (variation: Variation) =>
     [baseVariationId, changeVariationId].includes(variation.variationId)
 
+  const impactMultiplier = Analyses.getImpactMultiplier(
+    experiment,
+    latestAnalysis.participantStats['total'],
```
jessie-ross, line 393: I don't think this takes into account experiments with less than 100% allocation to variations.

src/components/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -439,23 +440,14 @@ export default function ExperimentResults({
...
-          Impact{' '}
+          Estimated impact{' '}
           <span className={classes.impactIntervalSelectWrapper}>
             (per
-            <Select
-              id='impact-interval-selector'
-              labelId='impact-interval-selector-label'
-              value={impactInterval}
-              onChange={(e) => setImpactInterval(e.target.value as Analyses.ImpactInterval)}
+            <ImpactIntervalSelector
```
jessie-ross, line 446: I think the page is simpler with just one selector at the top level, rather than also selectors per metricAssignment.

I think it results in better UX:
- Users know what they are comparing across metrics.
- No way to get lost in what has been set where.

We can also inline the select and the analysis functions if that's the case.

jessie-ross (changes requested):

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
razvanpapadopol: I thought that even if the allocation is, let's say 20%, that population is still representative of the entire targeted audience. If that's not correct, I still consider useful to use the `total` number of participants and maybe include in the [estimated impact disclaimer](https://github.com/Automattic/abacus/pull/772/files#diff-371c70a25deef9b4efcdb7ca1f7ac64bae51c82513c32e5cd392054f986b6c69R484-R485) something like `[...] conditions were unchanged for one {impactInterval}, across all participants`.

aaronyan: > we should be looking at our best guess of population rather than samples/representations of the population

I think it makes sense to have the estimates be based on the estimated total population (rather than the samples). It would be as as simple as dividing by total allocated percentage:`totalParticipants / totalAllocatedPercentage` where `totalAllocatedPercentage` can be between `0` and `1`.

If we feel strongly one way or another, we could make the population another toggle and put it next to the [top-level interval toggle](https://github.com/Automattic/abacus/pull/772#discussion_r957437210) (targeted, whole). If `totalAllocatedPercentage`, we could default to `whole` and disable the toggle.

jessie-ross: > If we feel strongly one way or another, we could make the population another toggle and put it next to the https://github.com/Automattic/abacus/pull/772#discussion_r957437210 (targeted, whole). If totalAllocatedPercentage, we could default to whole and disable the toggle.

Uhhh... I feel strongly that we shouldn't provide this as an option - more complexity for the code and UX.

src/components/general/MetricValue.tsx
razvanpapadopol: Yes, the unit type (USD or cash) is the same but they require different formatting. 

In a [follow-up PR](https://github.com/Automattic/abacus/pull/774/files#diff-6e88ab8578a50fa634687bfa0a325b8cec511bd1f473b152e112f8787814ac6f), I propose handling the formatting explicitly because what we use now for `usdFormatter` is actually more generic than that and can be called `localizeNumberToFixedDecimals`, while for the impact values we'd likely use an abbreviation.

See also https://github.com/Automattic/abacus/pull/772#discussion_r957194197

src/components/general/MetricValue.tsx
```diff
@@ -71,16 +88,25 @@ export const metricValueFormatData: Record<string, MetricValueFormat> = {
...
   metricParameterType,
   isDifference,
+  isImpact,
 }: {
   metricParameterType: MetricParameterType
   isDifference: boolean
+  isImpact: boolean
 }): MetricValueFormat {
-  return metricValueFormatData[`${metricParameterType}${isDifference ? '_difference' : ''}`]
+  return metricValueFormatData[`${metricParameterType}${isDifference ? '_difference' : ''}${isImpact ? '_impact' : ''}`]
```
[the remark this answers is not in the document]
razvanpapadopol: See my comments above. I think a new `formatter` prop on `MetricValue` [will remove the need](https://github.com/Automattic/abacus/pull/774/files#diff-7ec1a52a765aa5060aa0307214720f40ae0435b60abe68f3e1455b1fe64d3e3dL76-L85) for both `_difference` and `_impact` while enabling further extensibility (eg: importing a 3rd party formatting function as long as it's type `NumberToString`).

src/components/general/MetricValue.tsx
```diff
@@ -57,6 +67,13 @@ export const metricValueFormatData: Record<string, MetricValueFormat> = {
     transform: (x: number): number => x * 100,
     formatter: standardNumberFormatter,
   },
+  conversion_impact: {
+    unit: '',
+    prefix: '',
+    postfix: '',
```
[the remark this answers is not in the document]
razvanpapadopol: Added the postfix in https://github.com/Automattic/abacus/pull/772/commits/6ac08c5d15938d2f980accaea61ab15f69de8e74

src/lib/analyses.ts
razvanpapadopol: Thanks for the hints!

> We would need to look at the variation allocation percentages

This was addressed in https://github.com/Automattic/abacus/pull/772/commits/72c832a25ca562d931fb8b53f5ce13ea6fe9c14c as discussed in https://github.com/Automattic/abacus/pull/772#discussion_r957505000 and below.

Everything else was included in  https://github.com/Automattic/abacus/pull/772/commits/a7b40e790af88fd508d803a9ac502d5d30c9a83e

src/components/experiments/single-view/results/ImpactIntervalSelector.tsx
razvanpapadopol: Good point! Cleaned up the unnecessary type and map and started to pass `months: number` in https://github.com/Automattic/abacus/commit/a7b40e790af88fd508d803a9ac502d5d30c9a83e

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -510,6 +521,39 @@ export default function MetricAssignmentResults({
...
+                  <MetricValue
+                    metricParameterType={metric.parameterType}
+                    value={latestEstimates.diffs[variationDiffKey].top_95 * estimatedTotalParticipantsForImpact}
+                    displayPositiveSign
+                    isImpact
+                  />
+                  . <strong>This is not a statistical forecast</strong> but a theoretical cumulative effect on the
+                  targeted audience, as if the experiment conditions were unchanged for one{' '}
+                  {impactIntervalInMonths === 1 ? 'month' : 'year'}.
+                </Typography>
```
razvanpapadopol, line 556: This should cover aproximately what was summarised in pbmo2S-1Cf-p2#comment-2550 and the other convos we had.
* the relative change, together with the estimated impact in the same sentence
* a hint on how we calculate the estimated impact by including the experiment runtime 
* visible disclaimer that this is not a statistical forecast

We may also include here a link to a detailed explanation in FG

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -510,6 +521,39 @@ export default function MetricAssignmentResults({
                   />
                   .
                 </Typography>
+                <Typography variant='body1' gutterBottom>
+                  Given the relative change (lift) between{' '}
+                  <MetricValue
+                    metricParameterType={MetricParameterType.Conversion}
```
razvanpapadopol, line 527: We shouldn't need to pass the `Conversion` metric type to display relative change. This applies also to "Estimated impact" column in `ExperimentResults` component and should be fixed with the refactor in https://github.com/Automattic/abacus/pull/774

src/components/experiments/single-view/results/ExperimentResults.tsx
razvanpapadopol: As it is right now, the selector is displayed only at top level, in the header of Metric Assignment Results table. Selecting the interval there applies to every metric and also to every Metric Detail text.

razvanpapadopol:
Hey @aaronyan and @jessie-ross! I think this is now ready for another round of reviews. I've updated also the PR description and Demo and here's a quick summary of changes:
* Merged the impact interval and "Relative change (lift)" into an "Estimated Impact" column as described in [#780](https://github.com/Automattic/abacus/issues/780), with vertical centering. Other style changes are going into a separate PR for easier reviewing.
* Updated estimated impact interval tooltip content and the [sentence in the Metric Details - Summary](https://github.com/Automattic/abacus/pull/772#discussion_r1001041327).
* [Total allocated percentage](https://github.com/Automattic/abacus/pull/772#discussion_r957434057) is now taken into consideration.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: As an aside, this might actually be a different calculation to the impact from pipe, as they will want just their treatment sample to show the incremental money made.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: I'm fine with this for this PR, but I think we should include the number of participants over the time period as it is an important factor of the calculation.

jessie-ross (approved):
Woo! Thanks for all your work on this!

It's looking good!

aaronyan (approved):
Yay, amazing job @razvanpapadopol! Tests well and looks great so far. I'm excited to get this in the hands of experimenters. Thanks for iterating and being so thoughtful in the implementation.

I think we can improve the decimal rounding on the estimated impact in a future PR. Some quick thoughts for now:
- No postfix, round to whole number. 451 stays the same
- Has a postfix (K, M) round to a single decimal. 4.51K -> 4.5K.

Relative percentages can also probably use some rounding love: 
- Below 5% round to two decimals. 3.94% stays the same
- Below 30% round to one decimal. 20.24% -> 20.2%
- Otherwise round to whole percents. 45.12% -> 45%

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
razvanpapadopol: Done in https://github.com/Automattic/abacus/pull/774/files#diff-371c70a25deef9b4efcdb7ca1f7ac64bae51c82513c32e5cd392054f986b6c69L527

PR #773, opened 2022-08-29
Require a primary metric for new experiments

jessie-ross:
**This PR adds a test to the schema to require a primary metric for metric assignments.**

Screen Shot 2022-08-29 at 4 45 37 pm

Resolves 685-gh-Automattic/experimentation-platform

## How has this been tested?

- Automated tests
- Manually tested


aaronyan (commented):
Thanks so much for finding and fixing this, @jessie-ross!

razvanpapadopol (approved):
Great little fix! Since the scenario described in 685-gh-Automattic/experimentation-platform#issuecomment-1229949828 happens quite rarely and, more importantly, the primary metric can't be changed mid-experiment, I can't think of any suitable UX improvement. This approach is perfect 👌🏼 🚢

PR #774, opened 2022-08-29
Refactor metric value and formatters

razvanpapadopol:
Follow up for https://github.com/Automattic/abacus/pull/772

This set of changes is based on the thoughts shared in https://github.com/Automattic/abacus/pull/772#pullrequestreview-1084866953

**To do:**
- [x] Fix the need to pass `Conversion` metric type in order to display relative change https://github.com/Automattic/abacus/pull/772#pullrequestreview-1149866244
- [ ] Use `MetricValue` component as a value formatting component even when no `metricParameterType` is available

**Follow-up:**
- [ ] Update the value abreviation logic https://github.com/Automattic/abacus/pull/772#pullrequestreview-1151618649

## How has this been tested?


src/components/general/MetricValue.tsx
```diff
@@ -5,58 +5,62 @@ import React from 'react'
...
   postfix: React.ReactNode
   transform: (n: number) => number
-  formatter: (n: number) => string
+  formatter: NumberToString
 }
 
-const standardNumberFormatter = (n: number): string => `${_.round(n, metricValueFormatPrecision)}`
-const usdFormatter = (n: number): string =>
-  n.toLocaleString(undefined, { maximumFractionDigits: 2, minimumFractionDigits: 2 })
+export enum MetricUnitType {
```
jessie-ross, line 28: I like this 🙂 

Perhaps we can call it `UnitType` and abstract it from Metrics all together in the future?

I think this is the major part of the change - have units be separate from MetricTypes - this goes hand in hand with the metric generalisation we are discussing [here](https://github.com/Automattic/experimentation-platform/pull/782#discussion_r997921317.)

src/components/general/MetricValue.tsx
```diff
@@ -120,24 +98,24 @@ export function getMetricValueFormatData({
...
   metricParameterType,
-  isDifference = false,
+  displayAs,
   displayUnit = true,
   displayPositiveSign = false,
-  isImpact = false,
+  formatter,
 }: {
   value: number
   metricParameterType: MetricParameterType
```
jessie-ross, line 107: Can we remove this and make it the callers responsibility to provide a `displayAs` argument? Maybe we can change the `displayAs` argument to just `unit`?

jessie-ross (commented):
What do you think about changing `displayAs` to `unit` as per the comments?

src/components/general/MetricValue.tsx
razvanpapadopol: Totally! Having just `unit` prop instead of `metricParameterType` and `displayAs` cleaned up nicely the component interface, one step further after removing `isDifference` and `isImpact`. Especially in the `MetricAssignmentResults` component, the results are super visible  🎉 

Note that after this change, callers can use `getUnitType` helper function where the second argument (a specified unit type) is equivalent to the mentioned `displayAs`. We need this in order to always show `UnitType.Usd` unit when a dynamic `MetricParameterType` is `Revenue`.

src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
```diff
@@ -265,7 +265,7 @@ const ExperimentsTable = ({ experiments }: { experiments: ExperimentSummary[] })
               valueGetter: (params: { data: { analyses: Analysis[] } }) =>
                 params.data.analyses[0]?.participantStats.total || 0,
               cellRendererFramework: ({ value: participants }: { value: number }) => {
-                return <MetricValue value={participants} metricParameterType={MetricParameterType.Count} />
+                return localizeNumber(participants)
```
jessie-ross, line 268: I think we should keep one component for displaying values, what do you think?

src/components/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -697,7 +691,7 @@ export default function ExperimentResults({
                     <>
                       <div className={classes.summaryStatsPart}>
                         <Typography variant='h3' className={classes.summaryStatsStat} color='primary'>
-                          {totalParticipants.toLocaleString('en', { useGrouping: true })}
+                          {localizeNumber(totalParticipants)}
```
jessie-ross, line 694: Maybe we should use MetricValue here

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -506,49 +505,42 @@ export default function MetricAssignmentResults({
...
                     value={latestEstimates.diffs[variationDiffKey].bottom_95 * estimatedTotalParticipantsForImpact}
                     displayPositiveSign
                     displayUnit={false}
-                    isImpact
                   />{' '}
                   to{' '}
                   <MetricValue
-                    metricParameterType={metric.parameterType}
+                    unit={UnitType.Count}
+                    formatter={abreviateNumber}
```
jessie-ross, line 541: What if we just abbreviate all counts?

src/components/general/MetricValue.tsx
```diff
@@ -5,58 +5,62 @@ import React from 'react'
...
+export const metricValueFormatData: Record<UnitType, MetricValueFormat> = {
+  [UnitType.Proportion]: {
     unit: '%',
     prefix: '',
     postfix: '%',
     transform: (x: number): number => x * 100,
     formatter: standardNumberFormatter,
   },
-  conversion_difference: {
+  [UnitType.Difference]: {
```
jessie-ross, line 63: This should be called PercentagePoints, it's a little frustrating though because we provide percentagePoints/100 "ratioPoints".

Actually it should be called "ratioPoints" - the unit we are providing.

src/lib/schemas.ts
```diff
@@ -159,8 +159,6 @@ export interface MetricRevenueParams extends yup.InferType<typeof metricRevenueP
 export enum MetricParameterType {
   Conversion = 'conversion',
   Revenue = 'revenue',
-  // Used in UI only:
-  Count = 'count',
```
jessie-ross, line 163: 🎉

jessie-ross (approved):
Some minor comments but this looks good.

src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
razvanpapadopol: If we consider the current `MetricValue` component something like `FormattedValue`, then yes, it makes sense to use the component instead of the utility function. This change was from the time when `metricParameterType` was a prop on `MetricValue`.

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
razvanpapadopol: Abbreviation is something that may or may not be needed depending on the UI region. For example, if we use the unit `Count` for participants (mentioned in https://github.com/Automattic/abacus/pull/774#discussion_r1009917797) we'll lose the nice alignment if we abbreviate.

Screen Shot 2022-11-03 at 00 01 36

razvanpapadopol:
Thanks for the review, @jessie-ross !
We'll merge the refactor and update the value abbreviation logic in a separate PR.

PR #776, opened 2022-08-31
Retain experiment search / filter / sort in URL #545

razvanpapadopol:
* stringify filterState and sortState and add to URL params
* handle invalid JSON format when parsing the params
* keep applying the default sort when no param is passed

## How has this been tested?

### DIFF
--- a/src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
+++ b/src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
@@ -4,10 +4,23 @@ import 'ag-grid-community/dist/styles/ag-theme-alpine.css'
 
 import { Button, createStyles, fade, InputBase, Link, makeStyles, Theme, Typography, useTheme } from '@material-ui/core'
 import { Search as SearchIcon } from '@material-ui/icons'
-import { ColumnApi, GridApi, GridReadyEvent } from 'ag-grid-community'
+import {
+  DateFilterModel,
+  FilterChangedEvent,
+  FirstDataRenderedEvent,
+  GridApi,
+  GridReadyEvent,
+  GridSizeChangedEvent,
+  NumberFilterModel,
+  SortChangedEvent,
+  SortModelItem,
+  TextFilterModel,
+} from 'ag-grid-community'
 import { AgGridReact } from 'ag-grid-react'
 import clsx from 'clsx'
-import React, { useEffect, useRef, useState } from 'react'
+import _ from 'lodash'
+import { useSnackbar } from 'notistack'
+import React, { useRef, useState } from 'react'
 import { Link as RouterLink, useHistory, useLocation } from 'react-router-dom'
 
 import DatetimeText from 'src/components/general/DatetimeText'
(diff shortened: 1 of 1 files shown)


src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
```diff
@@ -86,82 +99,118 @@ const useStyles = makeStyles((theme: Theme) =>
...
+    const { api } = event
+    gridApiRef.current = api
   }
 
-  const onGridResize = () => {
+  const [searchState, setSearchState] = useState<string>(searchFromUrl)
+  const onSearchChange = (searchQuery: string) => {
+    setSearchState(searchQuery)
+    searchQuery ? queryParams.set('search', searchQuery) : queryParams.delete('search')
+    history.replace(`${pathname}?${queryParams.toString()}`)
```
jessie-ross, line 141: Let's not use `queryParams` directly as it introduces mutable state which is quite complicated to reason about especially with React. It also means we have duplication of state - `queryParams` and the individual `search`, `sort`, and `filtersFromUrl`.

We should have a `replaceUrlSearchState(state, stateSchema)` (as well as a `pushUrlSearchState` to match).

That way we can deal with all this immutably like so:
```
replaceUrlSearchState({
    ...gridState,
    search: searchQuery,        
}, gridSearchStateSchema)
```

src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
```diff
@@ -86,82 +99,118 @@ const useStyles = makeStyles((theme: Theme) =>
...
+    gridApiRef.current.setQuickFilter(searchQuery)
+  }
 
-    gridApiRef.current.sizeColumnsToFit()
+  const onFilterChange = (event: FilterChangedEvent) => {
+    const filterState = event.api.getFilterModel()
+    Object.keys(filtersFromUrl).forEach(
+      (filter) => !Object.keys(filterState).includes(filter) && queryParams.delete(filter),
+    )
+    Object.keys(filterState).forEach((key) => key && queryParams.set(key, JSON.stringify(filterState[key])))
```
jessie-ross, line 155: If we have a util function that is given a schema, we can know the type at runtime, and if the type is not simple (string, number) we can convert it to JSON.

You can access the types by using the `.fields` property on the schema, and check their type by using `.type` on the fields, e.g. `gridStateSchema.fields['search'].type === 'string'`.

This allows us to separate all encoding/decoding of the data from the rest of the code :-)

It also gives us a very useful tool that allows simple storage of state of any kind in URLs.

I think we should actually make the hook like so:
`const [ urlSearchState, replaceUrlSearchState, pushUrlSearchState ] = useUrlSearchState(schema)`

Giving us an abstraction that behaves identically to `useState`, making it simple and familiar to all future developers.

jessie-ross (changes requested):
Thanks for adding this!

I think we need to add an abstraction around url search state:
```javascript
const [ urlSearchState, replaceUrlSearchState, pushUrlSearchState ] = useUrlSearchState(schema)
```

I've outlined it in the comments below, but the aim is:
- Centralise encoding/decoding.
- Add validation and typing.
- Avoid mutable state.
- Utilise a more "React" based idiom.
- Make search state easily reusable - we can eventually add search state into results pages, metrics etc.

src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
```diff
@@ -86,82 +99,118 @@ const useStyles = makeStyles((theme: Theme) =>
...
 
-  const gridApiRef = useRef<GridApi | null>(null)
-  const gridColumnApiRef = useRef<ColumnApi | null>(null)
+  const history = useHistory()
+  const { pathname, search } = useLocation()
+  const queryParams = new URLSearchParams(search)
+  const {
+    search: searchFromUrl = '',
+    sort: sortFromUrl,
+    ...filtersFromUrl
```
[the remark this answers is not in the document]
razvanpapadopol: Makes sense. Implemented in https://github.com/Automattic/abacus/pull/776/commits/cb755687ed69aa3be60c69cb71d4cea692e31879 which also cleaned things a little bit.

src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
razvanpapadopol: Moved `queryParams` inside the custom hook for now. I agree it may be further cleaned up.

src/components/experiments/multi-view/ExperimentsTableAgGrid.tsx
razvanpapadopol: Thanks for the directions shared here! I've followed them to extract the URL processing to a custom hook in https://github.com/Automattic/abacus/pull/776/commits/ae8f5e0b1ab1bf90ef90c6d03fddd307a5dc856f
Adapting it to use `yup` (I haven't managed to make it work with the new version) may be the next step and probably a necessary one if the hook starts to be reused.

PR #777, opened 2022-09-01
Namespace ExPlat

jessie-ross:
**This PR namespaces ExPlat to allow pipe to join us in the future!**
- Allows pipe to start going straight away.
- Probably should to refactor the Auth code too, but it isn't too bad.
- Need to either namespace explat routes and share the routing table, or switch the routing table depending on host.

See pbmo2S-1Br-p2 and 1356-gh-Automattic/pipe

## How has this been tested?


razvanpapadopol (approved):
Hey, @jessie-ross 👋🏼 

The namespace refactor looks really good and should be good to go after a rebase and fixing tests.

I do love especially the split into `/general` and `/explat` components 🏅 
Do you see us having also a `lib/general` in the future? I'm thinking at any metrics from pipe that may be used also in ExPlat. But maybe that's not going to be done on the client side 🤔

jessie-ross:
I'm going to redo this PR after #687 to prepare for upcoming MLX work

jessie-ross:
I've realised I am not too happy with this, but this is an improvement and makes it easy for pipe or ml_sales to get involved without messing up our structure.

What might be better is something more like this:
```
src/
src/explat
src/explat/api
src/explat/components
src/explat/lib
src/pipe
src/pipe/api
src/pipe/components
src/pipe/lib
src/platform/api
src/platform/components
src/platfom/lib
```

We could make this move in the future - and it will be easier with the structure of this PR.

Issue #779, opened 2022-09-09
Add multiple lifts table to experiment results


aaronyan:
Proposed initially here: pbmo2S-1E7-p2. Multiple credible intervals (50%, 95%, 99%) has been helpful for experimenters when interpreting experiment results. We currently only show absolute differences; we should add multiple lifts table to experiment results with a show/hide option.

PR #781, opened 2022-09-21
Add a basic credible intervals table

jessie-ross:
**This PR adds the most basic credible interval table.**
- As basic as possible.
- Works for A/B/n.
- Main purpose is to provide a better alternative to using the graph.
- We can improve on it after this iteration.

Screen Shot 2022-09-22 at 1 36 46 am

Resolves https://github.com/Automattic/abacus/issues/779

cc @donlair 

## How has this been tested?


aaronyan:
Whooa very cool! Thanks for adding this credible intervals table, @jessie-ross. What are your thoughts for adding both `treatment - control` and `control - treatment`?

jessie-ross:
> What are your thoughts for adding both treatment - control and control - treatment?

It was mostly done to simplify development and get this out faster.

jessie-ross:
> It was mostly done to simplify development and get this out faster.

I don't view this as a very important feature so I don't think it needs to be too polished, and I'd like us to not put too much effort into it.

aaronyan (commented):
> It was mostly done to simplify development and get this out faster.
> I don't view this as a very important feature so I don't think it needs to be too polished, and I'd like us to not put too much effort into it.

Thanks for the context! Sounds good to me, I'm down to focus more on speed than polish with this. I think it would be slightly cleaner to have only `treatment-control` values, since the rest of the tool uses this standard. LGTM otherwise!

donlair:
Thanks for adding that @jessie-ross

razvanpapadopol (approved):
The changes look good for the first iteration. Great work, @jessie-ross!

I can see the value in displaying all the combinations in case of an A/B/n test but for regular A/B tests I agree with @aaronyan that `[treatment - control]` is enough. I think it wouldn't take much effort to do that simplification using the `isMultivariation` flag but it's fine to do it in a follow-up.

Issue #756, opened 2022-08-04
Minimum Difference Calculator - UI improvements


razvanpapadopol:
- [ ] There is room for improvement to the UX of calculator inputs when typing/pasting values. Currently, the default `0` feels a bit in the way.
- [ ] For some reason, I don't find the `ArrowDropDown` icon descriptive enough before expanding the calculator. Maybe we try the `CalculateIcon` once we migrate to the latest MUI version. Or we can also explore adding a tiny label next to it.
- [ ] Add tests to the `MinDiffCalculator` component itself.

_Originally posted by @razvanpapadopol in https://github.com/Automattic/abacus/pull/746#pullrequestreview-1061578961_

razvanpapadopol:
We're pausing the work on these improvements for now and the ones that were implemented in https://github.com/Automattic/abacus/pull/797 are checked off in the issue description.

PR #782, opened 2022-09-22
Add Min Diff Calculator to an existing experiment

razvanpapadopol:
This PR is integrating the min diff calculator into the Metric assignment dialog for an existing experiment.

* Sample size is persisted as long as the Metric assignment dialog is open
* If no metric is selected:
  * Toggle button is disabled
  * Calculator is hidden

Also, in this PR we introduce a new component called `MetricAssignmentForm`. This can now be used in other parts of the application where we might need to enable metric assignment (eg: on the experiment results tab).

#### Demo:

Screen Shot 2022-09-29 at 18 19 13

## How has this been tested?

  * Open an existing experiment
  * When trying to assign a new metric, use the button next to the Min-diff input to expand the calculator
  * Fill the required fields of the form and set the calculated min diff as the new value for the main input


src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
```diff
@@ -176,6 +191,16 @@ function MetricAssignmentsPanel({
...
   }
 
+  const [isActiveMinDiffCalculator, setIsActiveMinDiffCalculator] = useState<boolean>(false)
+  const handleMinDiffCalculatorToggle = () => {
+    setIsActiveMinDiffCalculator(!isActiveMinDiffCalculator)
+  }
+
+  const [samplesPerMonth, setSamplesPerMonth] = useState(0)
+  const handleMinPracticalDiffUpdate = (newMinDiff: number, formikProps: MetricFormikProps) => {
+    newMinDiff && formikProps.setFieldValue('metricAssignment.minDifference', newMinDiff / 100)
```
jessie-ross, line 201: It is a bit disappointing that the number coming out of the calculator needs to be divided by 100, ideally all the min diff values should be the same and we only change them for presentation.

src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
```diff
@@ -176,6 +191,16 @@ function MetricAssignmentsPanel({
...
+  }
+
```
jessie-ross, line 203: I think we should move the above new code to around line 275, within the `(formikProps) =>` function. 

That will simplify some things:
- We won't need to pass in/type formikProps
- We don't need a function to determine the parameter type, we can just us `const parameterType = `
- It keeps the add-metric functionality all in one place.

src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
```diff
@@ -338,14 +365,31 @@ function MetricAssignmentsPanel({
...
+                    </ToggleButton>
+                  </div>
+                  <div className={clsx(classes.row, !isActiveMinDiffCalculator && classes.minDiffCalculatorCollapsed)}>
+                    <MinDiffCalculator
+                      setMinPracticalDiff={(newMinDiff) => handleMinPracticalDiffUpdate(newMinDiff, formikProps)}
+                      {...{
+                        samplesPerMonth,
+                        setSamplesPerMonth,
+                        experiment: experimentToFormData(experiment),
+                        isConversion: getParameterType(formikProps) === MetricParameterType.Conversion,
```
jessie-ross, line 390: I think the calculator should take a `MetricParameterType` rather than `isConversion`, we'll need it when we add new metric types like for pipe.

jessie-ross (changes requested):
Looking good, main change requested is moving the code within the `<Formik>` closure, I think that will simplify the code and make it simple to extract the form into its own component (something we should probably do as well.)

src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
razvanpapadopol: I agree this needs to be fixed in the calculator itself and maybe also explore if it has something to do with MetricValue component. For now, I've added this point to https://github.com/Automattic/abacus/issues/756

src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
razvanpapadopol: Done the simplification in https://github.com/Automattic/abacus/pull/782/commits/cf615e896ac712597692dd2dd5b9b706a55e13d7 and then the full extraction of `MetricAssignmentForm` component in https://github.com/Automattic/abacus/pull/782/commits/de93c165c52b303150e0491b362ee4b5b994d08c

jessie-ross: Thanks for going ahead and extracting the component!

jessie-ross (commented):
I haven't gotten round to taking a good look at this yet, left some very small comments, I'll check back tomorrow.

jessie-ross (approved):
LGTM!

src/components/experiments/single-view/overview/MetricAssignmentsPanel.tsx
razvanpapadopol: Good point! I've changed the prop in https://github.com/Automattic/abacus/pull/782/commits/8d066f5a5e29202ac63843a645ae4ca2c4120bfe and, for now, the `isConversion` logic happens within the calculator. We can further improve it when needed.

PR #783, opened 2022-09-26
Simplify multiple CI table for A/B experiments

razvanpapadopol:
This is a follow-up based on https://github.com/Automattic/abacus/pull/781#pullrequestreview-1117701623

For A/B experiments we can simplify the table and use only the classic `treatment-control` and `(treatment-control)/control` values.

* **Before**
Screen Shot 2022-09-27 at 01 19 34

* **After**
Screen Shot 2022-09-27 at 01 17 31

## How has this been tested?


aaronyan (commented):
LGTM! Thanks for the quick fix. Works well for two variation experiments and it still shows all combinations for 3+ variations (as expected)!

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -399,6 +399,10 @@ export default function MetricAssignmentResults({
   const isVariationSelected = (variation: Variation) =>
     [baseVariationId, changeVariationId].includes(variation.variationId)
 
+  const variationsTuples = isMultivariation
+    ? cartesianProduct(experiment.variations, experiment.variations)
+    : [variations.slice().reverse()]
```
jessie-ross, line 404: Yeah this is a good solution.

```suggestion
    // Need to reverse the A/B variations for the grid (e.g. Treatment - Control):
    : [[variations[1], variations[0]]]
```
I think this would make what we are doing clearer.

jessie-ross (approved):
See the comment, but LGTM

src/components/experiments/single-view/results/MetricAssignmentResults.tsx
razvanpapadopol: Yes, 100%! Otherwise, it would require a bit of code reading in order to get even why we're using the locally defined `variations` Array.

Issue #778, opened 2022-09-09
Add color to visualized credible interval


aaronyan:
Separating this subtask from https://github.com/Automattic/abacus/issues/660 into its own issue. Color would be used to help with the metric interpretation instead of using color to signal deployment action. Here's a mockup!

- > min diff -> green
- < min diff -> blue
- between min diff -> yellow

<img width="1680" alt="186020501-6d55891b-29be-4979-b5c1-a88d77a237a6" src="https://user-images.githubusercontent.com/4505888/189364683-c32ea030-ae05-49b7-8fc2-0d165eb95b5a.png">

aaronyan:
Hmm this might be a bit too much color. 😅 I'm proposing now to make the whole bar a certain color depending on the **recommendation**.

<img width="1680" alt="Screen Shot 2022-09-15 at 2 56 16 PM" src="https://user-images.githubusercontent.com/4505888/190515856-5d76f37f-24d3-4cb1-bf59-86beecb45943.png">

- The color is based on the recommendation, not analysis
- When the recommendation is "not enough certainty", then we color the bar grey
- Example: Analysis says "control ahead" but the recommendation is "not enough certainty", then we color the bar grey

aaronyan:
The multi-colored bars might be a bit distracting/ confusing. Here's what it would look like mocked-up versus the single color bar!

<img width="1392" alt="Screen Shot 2022-09-15 at 4 14 22 PM" src="https://user-images.githubusercontent.com/4505888/190524573-108f6a14-67a9-4c7b-91e9-80a51c770a5f.png">

<img width="1392" alt="Screen Shot 2022-09-15 at 4 14 29 PM" src="https://user-images.githubusercontent.com/4505888/190524580-d46eaade-2273-4589-9d23-88dc2aa178c7.png">

jessie-ross:
I'm not particularly a fan of this use of colour, it's too easy to ignore the greyed out graphs, and since the decision on whether to colour them or not isn't smart enough an experimenter may accidentally ignore important information that we have greyed out.

I really like the new column arrangment, It is a great way to fit abs/relative/lift all at once :)

razvanpapadopol:
> an experimenter may accidentally ignore important information that we have greyed out

Yup, once we introduce color, grey becomes tricky. An option would be to use only the 2 colors from the original proposal. 

[...]

PR #787, opened 2022-10-07
Experiment Results: add colors to CI diagrams

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/778

#### Summary of changes
* Use green for results where we have enough data: `{variant} winning`, `no difference`.
* Use yellow for cases where we need more data: `inconclusive`.
* Use green for `{variant} ahead` and `{variant} barely ahead` analysis decisions when data is _strong enough_ and yellow when it's not.
* `strongEnoughForDeployment` is renamed to `strongEnoughData` and used also to determine the CI diagram color.
* Abacus custom `theme.palette` includes now these two colors.

#### Demo
Screen Shot 2022-10-08 at 02 00 40

#### To do
- [ ] Fix CredibleIntervalVisualization stories

## How has this been tested?


src/components/experiments/single-view/results/CredibleIntervalVisualization.tsx
```diff
@@ -60,6 +63,11 @@ export default function CredibleIntervalVisualization({
     }
   }
 
+  const isEnoughData =
```
jessie-ross, line 66: Shouldn't we be using [`isDataStrongEnough`](https://github.com/Automattic/abacus/blob/829fc876139645d941c3b08c2b859ca49e31a948/src/lib/recommendations.ts#L161-L202) for this logic?  

There are a few ways to do it such as  changing the input of this component to match the function call, or wrapping this in another component that provides a boolean of `isDataStrongEnough` for this component, or preferably for me, providing [the recommendation data](https://github.com/Automattic/abacus/blob/829fc876139645d941c3b08c2b859ca49e31a948/src/lib/recommendations.ts#L131) to this function.

jessie-ross (changes requested):

src/components/experiments/single-view/results/CredibleIntervalVisualization.tsx
razvanpapadopol: Oh, no. They are different as this one here is only "reading" the CI interval and min diff for a metric, disregarding the platform, runtime, etc. I think I should rename one or the other. `isDataStrongEnough` may as well be called `isDataStrongEnoughForDeploymentRecommendation` 😬 

> providing [the recommendation data](https://github.com/Automattic/abacus/blob/829fc876139645d941c3b08c2b859ca49e31a948/src/lib/recommendations.ts#L131) to this function.

This may work as it gives us access to `Decision` which can be used to map to colors exactly how we want:
> Use green for results where we have enough data: {variant} winning, no difference
> Use yellow for cases where we need more data*: inconclusive, {variant} ahead, {variant} barely ahead

aaronyan: I think we'll want the colors to use the same logic as adding a recommendation. If a metric has enough data for a recommendation, then we color the CI green. Otherwise, yellow.

razvanpapadopol: I see your point to have consistent logic to determine if a recommendation is based on strong data or not. I've pushed the changes and this is an example (using 20806-explat-experiment) where a previously yellow interval is now green:

Screen Shot 2022-10-13 at 18 01 49

If decision is one of `[VariantBarelyAhead, VariantAhead]`, should we display some explanation, maybe linking to a FG section such as PCYsg-Fqg-p2#reasons-for-not-enough-certainty ?

aaronyan: I love the idea of adding an explanation and linking the FG section!

src/components/experiments/single-view/results/CredibleIntervalVisualization.tsx
```diff
@@ -60,6 +67,8 @@ export default function CredibleIntervalVisualization({
     }
   }
 
+  const hasEnoughData = recommendation.strongEnoughData
```
jessie-ross, line 70: Nit: This variable is unnecessary

jessie-ross (approved):
LGTM

src/components/experiments/single-view/results/CredibleIntervalVisualization.tsx
razvanpapadopol: Great! I've opened the follow-up issue https://github.com/Automattic/abacus/issues/791 so we'll handle this extra bit in a separate PR. I'm releasing now the colored diagrams as they are already very informative by themselves 🚀

PR #788, opened 2022-10-10
Health Report: add platform exceptions for runtime indicator

razvanpapadopol:
Inspired by https://github.com/Automattic/abacus/pull/751#discussion_r991055790

Use the `runtimeWhitelistedPlatforms` list not only for deployment recommendations but also to reduce the noise coming from health indicators.

## How has this been tested?


jessie-ross:
I think we should add a custom static indicator to these that says "Health indicators", indication "Disabled", set to warning level, the recommendation can say "Health indicators are disabled for this platform, you are on your own :-)"

jessie-ross:
I like the idea of disabled indicators, I think we should change the assignment indicators use isDisabled as well if we go along this path. I think we'd also need to:
- Change the displayed recommendation to include: "This indicator is disabled for this experiment."
- When calculating the total severity for an experiment we need to exclude disabled indicators.

PR #789, opened 2022-10-17
Update issue templates

aaronyan:
Add issue template to better comply with the new way we will work in collaboration with the MLX group.


jessie-ross (approved):
LGTM :)

Issue #780, opened 2022-09-16
Add new analysis columns


aaronyan:
After adding absolute impact to Abacus (https://github.com/Automattic/abacus/pull/772), I think we can change the way we show the results of the experiment:

- [ ] Display baseline interval below metric name p10gg3-aFD-p2#comment-33110.
- [ ] Rename "absolute change" to "estimated difference"
- [x] Rename "relative change (lift)" to "estimated impact". Done in https://github.com/Automattic/abacus/pull/772
- [ ] Reduce the font size and font color to a grey `#828282` of absolute change and relative change
- [x] Add absolute impact to "estimated impact" with medium weight. Done in https://github.com/Automattic/abacus/pull/772
- [ ] Adjust analysis text to be medium weight
- [ ] Align absolute change and relative change to be on the same horizontal ruler
- [ ] The gap between absolute impact and relative change should be vertically aligned with metric name and analysis text

<img width="1194" alt="Screen Shot 2022-09-16 at 3 18 32 PM" src="https://user-images.githubusercontent.com/4505888/190815505-bf891f65-2fd8-4dfa-aba8-973976b90063.png">

razvanpapadopol:
A couple of these changes were already implemented with the release of estimated impact feature and the rest are all included in https://github.com/Automattic/abacus/pull/795 🚀 .

The only minor observation is the grey `#828282` being possibly a bit too light. Here's a comparison with `#616161` which is the `grey.A700` variable we're already using from the [default MUI color palette](https://mui.com/material-ui/customization/color/#color-palette).

<img width="686" alt="Screen Shot 2022-11-09 at 22 41 43" src="https://user-images.githubusercontent.com/14192054/200937937-c5cef1e0-028a-4af9-a87c-ccb1f0c6a8b1.png">
<img width="694" alt="Screen Shot 2022-11-09 at 22 39 46" src="https://user-images.githubusercontent.com/14192054/200937939-bdcf47fe-9be5-4d59-b3fd-921738b35423.png">

PR #790, opened 2022-10-21
Estimated impact based on relative change

razvanpapadopol:
Because there are still cases of [negative lower relative diff & positive absolute diff](https://github.com/Automattic/experimentation-platform/issues/749) (eg: 20807-explat-experiment) and we want to [display the estimated impact together with relative change interval](https://github.com/Automattic/abacus/issues/780), I've explored here the possibility to calculate the impact using the relative diff and observed mean of the base variation.

#### Demo
* **Before**
Screen Shot 2022-10-21 at 02 51 49

* **After**
Screen Shot 2022-10-21 at 02 52 00

## How has this been tested?


razvanpapadopol:
> possibility to calculate the impact using the relative diff

Ok, I know this is not the way we want the experimenters to calculate it (pbmo2S-1Cf-p2#comment-2565) but how we present it right now, the estimated (guessed) impact is a theoretical measure of scale, not a precision tool.

However, I'd be open also to keeping this as it is and go ahead with the [strange results link](https://github.com/Automattic/experimentation-platform/issues/749#issuecomment-1227152834) 😁

jessie-ross:
We should be using the absolute change here, as the difference calculation involves more steps and rounding/scaling issues.

But maybe we should increase the sample size to fix this discrepancy after all...

aaronyan:
+1 to increasing the sample size so we get way less discrepancies.

> We should be using the absolute change here, as the difference calculation involves more steps and rounding/scaling issues.

I'd be curious how far apart the absolute estimates would be using absolute vs relative.

If there is a discrepancy though, I prefer there being a difference between absolute and estimated impact (left vs right) instead of stacked numbers having a discrepancy. I'm a fan of Razvan's proposed changes, but happy to discuss more!

razvanpapadopol:
> I'd be curious how far apart the absolute estimates would be using absolute vs relative.

Oh, the discrepancy may be pretty big, as seen in the example that I've used here for demo 😬 
But even if this experiment had a large volume of participants, I think it is still kind of an extreme case because of a super low ACPU.

Yeah, let's discuss more as we may find a solution that is both precise and consistent. For now, I'm leaning to keep using absolute diff to calculate impact, especially in the context of showing the Baseline interval p10gg3-aFD-p2#comment-33110. I think we'll finally show the full story from left to right, giving a lot of context to those diff pp that may seem quite abstract by themselves.

razvanpapadopol:
When testing out a revenue metric with negative values, I've noticed that the direction of the relative change interval is actually inversed compared to the absolute change CI. Opened https://github.com/Automattic/abacus/issues/793 to further discuss how we can fix the inconsistency that is more obvious now with the impact displayed on top of relative diff.

**production:**
Screen Shot 2022-10-25 at 19 51 37

**with this PR (impact derived from relative dff):**
Screen Shot 2022-10-25 at 19 51 56

razvanpapadopol:
We can close this now that we have:
1. a way to reduce the discrepancy by increasing the posterior sample size: https://github.com/Automattic/experimentation-platform/pull/791
2. plans to add educational references: https://github.com/Automattic/abacus/issues/810

Issue #769, opened 2022-08-22
Create an individual Metric page and routes


razvanpapadopol:
Follow up for https://github.com/Automattic/abacus/issues/768, 
As mentioned in pbmo2S-1Am-p2#comment-254, we consider adding individual routes for each metric.

Format: `metrics/{metricId}` or `metrics/{metricId-metric-name}` as the experiment routes. Even if the latter may change once we archive metrics, it's probably doable to still do the matching.
**LE:** the matching works after updating the metric name (archiving) as it's based on the `metricId`, same as for the individual experiment.

- [x] add the new Metric page and support for the new routes #796
- [x] links from Experiment Overview & Results #829
- [x] links from Experiment Wizard #822
- [x] [shortcode link](https://github.com/Automattic/tampermonkey-highlighter) for metrics: `{metricId}-explat-metric` https://github.com/Automattic/tampermonkey-highlighter/pull/165

razvanpapadopol:
This issue is now completed will all the items merged & released. The only immediate follow-up is the [PR adding a metric details modal](https://github.com/Automattic/abacus/pull/826) to improve the UX in the experiment wizard.

PR #796, opened 2022-11-03
Metrics: add individual routes for metric details

razvanpapadopol:
Fixes part of https://github.com/Automattic/abacus/issues/769

This PR is adding a new Metric page and support for `/metrics/{metricId-metric-name}` individual routes.

**Summary of changes:**
* Add new Metric page.
* Add a `MetricDetails` component used for both the accordion view of `/metrics` and the new Metric page.
* Update `Routes` to handle `/metrics/:metricId`.
* Use a soft browser history manipulation on the existing `/metrics` page so the URL get the `/:metricIdSlug` but the accordion view is maintained.
  * The `metricIdSlug` for the latest expanded metric will be saved in URL.
  * When collapsing any metric detail panel, the `metricIdSlug` is removed.
* Add a `PageTitleWithSlug` component to be reused between Experiment and Metric pages.

**Demo**

Screenshot 2022-12-20 at 02 46 49

## How has this been tested?


jessie-ross:
This is looking good so far :)

I'd add the metric name in the url as well - like we do for experiments.

razvanpapadopol:
> This is looking good so far :)
> 
> I'd add the metric name in the url as well - like we do for experiments.

Thanks for the hint, @jessie-ross ! 
Going into that direction, I've extracted also the page title component with a loadable slug from experiments.

aaronyan (approved):
LGTM! Tested locally and works well. Thanks for adding @jessie-ross's idea of adding the metric name to the URL. 

> I'd add the metric name in the url as well - like we do for experiments.

I tried adjusting the name of a metric (something we can do in /metrics but not in /experiments) to see what happens to a URL with the old name, and it works as intended (redirects correctly)!

PR #664, opened 2022-03-15
Fix days value in analyses test

razvanpapadopol:
* Fix `getExperimentHealthIndicators` unit test
* Derive days from hours, as we do in the app

This is a follow-up to https://github.com/Automattic/abacus/pull/656

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)
- Manual testing with mock data (locally or on Vercel)
- Additional manual testing instructions (please specify):

PR #799, opened 2022-11-10
Fix date sensitive unit tests

razvanpapadopol:
This PR is **fixing the failing tests and preventing snapshot updating noise*** seen in all other PRs that are now open because the 1 hour +/- difference `new Date()` returns around the DST compared to what we expect in code (we always divide 1 day in 24h). 

Last time I tried to fix it in https://github.com/Automattic/abacus/pull/664 but I think it introduced unnecessary complexity.

* Only tests are affected by this change and no snapshots should be updated
* Use fixed dates instead of "now" in unit tests

***For example:**

Screen Shot 2022-11-10 at 22 50 40

## How has this been tested?


src/components/experiments/single-view/results/ExperimentResults.test.tsx
```diff
@@ -20,8 +19,9 @@ beforeEach(() => {
 })
 
 const experiment = Fixtures.createExperimentFull({
-  startDatetime: subDays(new Date(), 10),
-  status: Status.Running,
+  startDatetime: new Date('2021-04-01T00:00:00Z'),
+  endDatetime: new Date('2021-04-11T00:00:00Z'),
+  status: Status.Completed,
```
aaronyan, line 24: How come the status in these tests changes from `running` to `completed`?

aaronyan: Ah I see, thanks for the explanation!

aaronyan (approved):
LGTM! Thanks for making this daylight savings fix!

Issue #801, opened 2022-11-17
Documentation links lead to internal (and private) sites


bgbg:
The site https://abacus-a8c.vercel.app/ is supposed to be a public demonstration of Abacus. 
[This page](https://abacus-a8c.vercel.app/experiments/3-experiment-name-123/wizard-edit) contains several links that are supposed to provide useful information and guidance. However, these links lead to the private field guide.
 
<img width="1058" alt="image" src="https://user-images.githubusercontent.com/506547/202417131-da0814bf-0d7a-414c-8e26-58d7af8b742b.png">

razvanpapadopol:
Hi, @bgbg 👋🏼 
Thanks for reporting this issue! 
Together with https://github.com/Automattic/abacus/issues/803 that covers more code clean-up, we'll definitely add clarity to the public demonstration side of Abacus.

bgbg:
@razvanpapadopol thanks

PR #744, opened 2022-07-09
WIP: package updates

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/687

Changes:
* Update React to 18.2
* Updates react-scripts to 5.0.1
  * [shell-quote to 1.7.3](https://github.com/Automattic/abacus/pull/754)
  * [ejs to 3.1.8](https://github.com/Automattic/abacus/pull/755)
* Updates serve to 14.1.2 
  * [minimatch to 3.1.2](https://github.com/Automattic/abacus/pull/807)
* Updates jest to 29.3 + [check watch plugins](https://github.com/Automattic/abacus/pull/802#discussion_r1034194425)
* [WIP]

Notes:
* One of the issues migrating to React 18 and its concurrent mode are the unit tests firing `act()` related warnings as detailed [here](https://github.com/WordPress/gutenberg/pull/45235#issuecomment-1300359931).

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)
- Manual testing with mock data (locally or on Vercel)
- Additional manual testing instructions (please specify):

PR #802, opened 2022-11-17
Improve handling for documentation and other private links

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/801 
Fixes https://github.com/Automattic/abacus/issues/803

* Add <PrivateLink /> component that wraps the Material UI `<Link />`.
* Limit rendering the private links depending on the environment (only in production) when the app is built using `mockConfig`.
* Switch to using shortlinks for private resources.

## How has this been tested?


aaronyan (approved):
LGTM! Thanks for adding the new `PrivateLink` component, adding new tests and adjusting the existing ones.

package.json
```diff
@@ -97,6 +97,7 @@
     "isomorphic-fetch": "^3.0.0",
     "jest": "26.6.0",
     "jest-puppeteer": "^6.1.0",
+    "jest-watch-typeahead": "^0.6.5",
```
razvanpapadopol, line 100: After the [node update](https://github.com/Automattic/abacus/pull/806), a jest watch plugin seems to be needed in order to update snapshots. I'll check this again when [updating packages](https://github.com/Automattic/abacus/pull/744).

jessie-ross:
There is nothing wrong with the code here, and this isn't a detrimental change so feel free to merge - using the short-links is good in fact.

But I think the `PrivateLink` functionality is a bit of a security micro-optimisation - the links are still present in the code and if we want to stop scraping we could use a robots.txt.

package.json
jessie-ross: :(

razvanpapadopol:
> There is nothing wrong with the code here, and this isn't a detrimental change so feel free to merge - using the short-links is good in fact.

Thanks for having a look! 

> But I think the `PrivateLink` functionality is a bit of a security micro-optimisation - the links are still present in the code and if we want to stop scraping we could use a robots.txt.

Oh, this is not for security but to have a clean open source / Demo version of Abacus (https://github.com/Automattic/abacus/issues/801). Anyway, I don't think that usual crawlers will index the links if they are not accessible from the UI 🤷🏼

Issue #759, opened 2022-08-09
Add button to report a bug


aaronyan:
Should we add a button to Abacus that allows experimenters to quickly reach us on P2, Slack, or GitHub issue? @razvanpapadopol and I discussed this during our last 1-1!

jessie-ross:
+10000

This is a great idea! Very easy to implement - we can link it to the new issue button in Abacus and if necessary we can move the issue over to experiment-platform.

aaronyan:
Awesome, new issue button in Abacus sounds great to me too! I think part of resolving this issue is to create a template for new issues on Abacus. Editing the top to add these tasks!

razvanpapadopol:
> link it to the new issue button in Abacus and if necessary we can move the issue over to experiment-platform.

While it makes sense that most of the reported issues will be UI related, I wouldn't be comfortable sharing production bugs of an internal platform in a public repo. 

So let's do the other way around and have the bugs reported as an issue in the experiment-platform repo with the `experimenter-interface` label pre-applied and we'll take it from there.

razvanpapadopol:
Did a bit of exploration on this a while ago and concluded for now with https://github.com/Automattic/abacus/pull/805#issuecomment-1361448528.

PR #805, opened 2022-11-23
Add a bug report button

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/759

Screenshot 2022-12-21 at 16 55 48

## How has this been tested?


razvanpapadopol:
Even if the button looks nice, the right-alignment doesn't really work with the experiment wizard that is not full-width and on longer pages is not so visible, being in the footer. So right now I'm thinking it would be better to integrate it into the new header redesign (https://github.com/Automattic/abacus/issues/738).

PR #813, opened 2022-12-08
Wizard: temporary remove LOHP platform

razvanpapadopol:
Related to https://github.com/Automattic/experimentation-platform/issues/814#issuecomment-1336930489

* Comment out the `Platform.Lohp` and its description.

## How has this been tested?


jessie-ross:
This breaks Abacus as we validate all incoming data - we should exclude it from the UI but in general we should only make backwards compatible changes to the schema.

PR #816, opened 2022-12-09
Revert "Wizard: temporary remove LOHP platform"

jessie-ross:
Reverts Automattic/abacus#813


razvanpapadopol (approved):
Good catch! ❤️ 
So the integration tests were passing because they run on the assumption that changes to the schema are backward compatible. Yeah, it makes sense.

PR #818, opened 2022-12-16
Fix impacts: use analysis for enddatetime

jessie-ross:
**This PR has experiment run-time for impacts be calculated from the analysis datetime instead of the current time.**
- Fix impact to use the latest analysis datetime
- Add analysis run hours

Fixes [#822](https://github.com/Automattic/experimentation-platform/issues/822)

## How has this been tested?


razvanpapadopol (approved):
Good catch, @jessie-ross! Thanks for the fix!
The new unit tests are also looking good 🚀

src/lib/analyses.ts
```diff
@@ -540,6 +543,40 @@ export function getTotalEligiblePopulation(analysis: Analysis, experiment: Exper
   return analysis.participantStats['total'] / (getTotalAllocatedPercentage(experiment) / 100)
 }
 
+function utcStartOfDay(date: Date): Date {
+  // (I wish there was an easier way to do this...)
```
razvanpapadopol, line 547: +1
Date handling is a much bigger rabbit hole than one could anticipate!

PR #819, opened 2022-12-19
Explore a fix for flaky unit tests

razvanpapadopol:
**Note:** this PR still doesn't have a fix for the underlying issue.

Possibly after https://github.com/Automattic/abacus/pull/818, some snapshots seem to require updating without actual code changes. 
I've noticed this while working on https://github.com/Automattic/abacus/pull/796:

Screenshot 2022-12-19 at 16 06 28

From what I could track already, it may be because of how different we're handling dates in `src/test-helpers/fixtures.ts` for experiments / analyses, combined with the new way of using also the analysis date when calculating impact.

## How has this been tested?


jessie-ross (approved):

razvanpapadopol:
Thanks for having a look, @jessie-ross ! 
I'll merge this so we unblock the rest of the PRs and continue investigating separately.

PR #822, opened 2022-12-28
Add metric details links to experiment wizard

razvanpapadopol:
Fixes part of https://github.com/Automattic/abacus/issues/769
Follow up to https://github.com/Automattic/abacus/pull/796

Adds links to Metric details in Experiment Wizard.
* Use `OpenInNew` link button
* Open metric details in a new tab
  * **See https://github.com/Automattic/abacus/pull/826 for an alternative solution using a modal**
* Display full metric name
Screenshot 2022-12-28 at 18 25 43

## How has this been tested?


jessie-ross:
I think I prefer this solution, it is simpler code wise and more flexible for experimenters - they can open many at once.

jessie-ross (approved):
The code LGTM :)

razvanpapadopol:
Thanks for reviewing this, @jessie-ross !

> I think I prefer this solution, it is simpler code wise and more flexible for experimenters - they can open many at once.

Yes, this simple solution has solid advantages so I'll go ahead and merge and also implement it for the experiment overview & results. Sharing a metric is also easier once you have access to the metric page. 

On the other hand, the approach of https://github.com/Automattic/abacus/pull/826 involves less context switching which I find sometimes useful, especially for this experiment wizard case. Let's continue to discuss there if we can combine some of the benefits.

Issue #768, opened 2022-08-22
Retain metric search in URL


aaronyan:
Similar to https://github.com/Automattic/abacus/issues/545, let's also retain the search for metrics. This will be a fairly quick way to allow metrics be "linkable". It's okay that a search might return multiple metrics for now. We can think of more solutions later like adding a `/metrics/{metric_id}` endpoint in the future. This idea was originally proposed in this p2 post: pbmo2S-1Am-p2

razvanpapadopol:
Some progress for this feature has been made in https://github.com/Automattic/abacus/pull/824 but since we're planning to change the Metrics table from `material-table` to `ag-grid` (https://github.com/Automattic/experimentation-platform/issues/967), it would be simpler to implement both together.

PR #824, opened 2022-12-30
WIP: Metrics: Retain metric search in browser history

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/768

Similar to https://github.com/Automattic/abacus/pull/707, this PR is adding support to retain metrics search in URL.

To do:
- [ ] Pre-populate the search field with the value of `search` URL param.
- [ ] Add unit tests.

## How has this been tested?


src/components/explat/metrics/multi-view/MetricsTable.tsx
```diff
@@ -124,11 +133,13 @@ const MetricsTable = ({
       options={{
         ...defaultTableOptions,
         actionsColumnIndex: 3,
+        searchText: searchQuery,
```
razvanpapadopol, line 136: When adding this option to the "Basic Search Example" in the [MTable docs](https://material-table.com/#/docs/features/search), the input value is updated as expected. However, in Abacus only the internal state is updated (results are filtered) and not the input.

razvanpapadopol:
Closing this for now as it's not fully supported by the current implementation using `material-table`. 
See https://github.com/Automattic/abacus/issues/768#issuecomment-1603368110

PR #826, opened 2023-01-05
Wizard: improve UX by opening metrics details in a modal

razvanpapadopol:
This was originally an alternative to adding external links https://github.com/Automattic/abacus/pull/822

Now, with external links [successfully added to the Experiment Overview and Experiment Results views](https://github.com/Automattic/abacus/pull/829), opening the metrics details in a modal is an improvement for the wizard experience, to minimize distractions.

* Clicking on the "Open in new" / "Expand" icon will open the Metrics Details as a modal, without any loading.
* Clicking the "Close" (X) button or on the darkened area will close the modal.
* The page URL will be kept unchanged.

**Possible improvement**: add a "Copy" button next to the metric name.

### Demo
Screenshot 2023-01-05 at 01 54 19

## How has this been tested?


aaronyan (approved):
Looking great! Nice idea on using the modal as an alternative to https://github.com/Automattic/abacus/pull/822.

Ran this on my local machine, and it feels great! 🙌

jessie-ross:
I guess this is fine if you and @aaronyan are pro it, I'd personally not pursue a small improvement like this because of the added complexity.

I think we generally need a bar at which we decide not to make improvements because the benefits are too small, a "min practical difference" for our changes.

aaronyan:
> I think we generally need a bar at which we decide not to make improvements because the benefits are too small, a "min practical difference" for our changes.

Agreed! I think this improvement is worth it, but we will operate on what our min practical difference is for our changes based on our team priorities (pbmo2S-1R8-p2).

razvanpapadopol:
> because of the added complexity.

Yeah, I kind of agree and this is why I've tried to hide most of it (separate modal component). We can go even a step further and use a component for the metric name. https://github.com/Automattic/abacus/issues/833

PR #830, opened 2023-01-16
Add pipe metric schema

jessie-ross:
**This PR adds pipe metrics to ExPlat.**

See https://github.com/Automattic/experimentation-platform/issues/781

This PR just adds the schema in, a future PR will add an abstraction between the metric source and it's type.

## How has this been tested?


razvanpapadopol (approved):
The pipe metric type and schema look good!

PR #838, opened 2023-01-31
Add metric type abstraction

jessie-ross:
**This PR implements a futureproof unit abstraction by creating the UnitInfo interface.**
- UnitInfo interface is now passed around instead of unitTypes, this allows for adding extra information and simplifying code.
- Adds getUnitInfo, with the ability to get a derived unit.
- Adds the `ts-pattern` library for pattern matching and making getUnitInfo simpler.
    - Minimal library with no dependencies.
    - Does one basic thing.
    - Based on a [tc39 ECMAScript proposal](https://github.com/tc39/proposal-pattern-matching).

Part of progress on https://github.com/Automattic/experimentation-platform/issues/781

## How has this been tested?


jessie-ross:
Future commits will be passing metric down into any components that need to work with metrics, rather than metricTypes - this gives us a lot more room to grow and is more fitting with the overall architecture of the app where we pass around the central structures defined in schema rather than bits and pieces.

razvanpapadopol (approved):
Great work adding these abstractions, @jessie-ross !
There have been a few iterations but it's nice to see where we are with metric types in terms of flexibility and the declarative approach.

PR #840, opened 2023-02-03
Generalise Metric Management

jessie-ross:
**This PR generalises metric management to arbitrary metrics types.**
- Adds a metric source (the parameterType) to name map
- Removes unused map

Part of https://github.com/Automattic/experimentation-platform/issues/781

## How has this been tested?


src/components/explat/metrics/__snapshots__/MetricFormFields.test.tsx.snap
```diff
@@ -282,7 +334,7 @@ exports[`renders as expected for conversion metric 1`] = `
           <span
             class="MuiTypography-root MuiFormControlLabel-label MuiTypography-body1"
           >
-            Cash Sales
+            Pipe Model
```
razvanpapadopol, line 337: This is because of the changed order of the metric parameter types in the map [metricParameterTypeName](https://github.com/Automattic/abacus/pull/840/files#diff-3b3d20fd4d93130c7d476561821d690afcbb233373e9ec5d757ec8dd9f25a96aR18). Before, `Conversion` was the first option but I guess it's also correct this way.

razvanpapadopol (approved):
Metric management update looks good and tests well 🚀 
Thanks for also cleaning up the unused map! 

Noted just the different order of the options but since it's only accessible in debug mode, I wouldn't worry about it.

PR #841, opened 2023-02-03
Pass metric to MetricDifferenceField

jessie-ross:
**This PR passes metrics down to MetricDifferenceField.**

Part of progress on https://github.com/Automattic/experimentation-platform/issues/781

## How has this been tested?


src/components/general/MetricDifferenceField.tsx
```diff
@@ -31,67 +32,79 @@ const useStyles = makeStyles(() =>
...
-              USD
-            </InputAdornment>
-          ),
-        }}
-      />
-    )
-  } else {
-    throw new Error('Unknown or missing MetricParameterType')
+  switch (props.metric ? getUnitInfo(props.metric, [UnitDerivationType.AbsoluteDifference]).unitType : 'unknown') {
+    case 'unknown':
```
razvanpapadopol, line 42: From what I understand, this would be the loading state, right?
WDYT of moving it to an early return `if (!props.metric) {}` so we reserve the `switch` block for actual values and we don't need the ternary above.

razvanpapadopol (approved):
LGTM! 
Added just a note and it's completely optional.

src/components/general/MetricDifferenceField.tsx
jessie-ross: Yep good idea!

PR #842, opened 2023-02-03
Pass metric down to MinDiffCalculator

jessie-ross:
**This PR passes metrics down to MinDiffCalculator.**
- We'll need to refactor MinDiffCalc at a later stage...

Part of progress on Automattic/experimentation-platform#781

## How has this been tested?


razvanpapadopol (approved):
LGTM!
 
> We'll need to refactor MinDiffCalc at a later stage

I agree and we'll have a good opportunity to refactor when changing some of its functionality in the near future ( pbmo2S-1Up-p2 ). Eg: lift (gain) passed as a %, sample size input removed, etc.

PR #843, opened 2023-02-03
Generalise MetricAssignmentResults

jessie-ross:
**This PR generalises MetricAssignmentResults so that we aren't using metricParameterType.**

This is the last in this series of PRs and resolves the front-end portion of https://github.com/Automattic/experimentation-platform/issues/781

## How has this been tested?

### DIFF
--- a/src/components/explat/experiments/single-view/results/MetricAssignmentResults.tsx
+++ b/src/components/explat/experiments/single-view/results/MetricAssignmentResults.tsx
@@ -24,7 +24,7 @@ import MetricValueInterval from 'src/components/general/MetricValueInterval'
 import PrivateLink from 'src/components/general/PrivateLink'
 import * as Analyses from 'src/lib/explat/analyses'
 import { getChosenVariation, getExperimentRunHours, isOneTimeExperiment } from 'src/lib/explat/experiments'
-import { getUnitInfo, metricParameterTypeName, UnitDerivationType, UnitInfo, UnitType } from 'src/lib/explat/metrics'
+import { getUnitInfo, UnitDerivationType, UnitInfo, UnitType } from 'src/lib/explat/metrics'
 import * as Recommendations from 'src/lib/explat/recommendations'
 import {
   Analysis,
@@ -250,6 +250,7 @@ export default function MetricAssignmentResults({
 
   let estimateTransform: (estimate: number | null) => number | null = identity
   let unitName = ''
+  let countName = ''
   let metricValuePlotTitle = ''
   let metricValueAbsoluteDifferencePlotTitle = ''
   // Some of the information here should eventually be moved to UnitInfo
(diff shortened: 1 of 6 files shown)


src/components/explat/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -557,10 +579,8 @@ export default function MetricAssignmentResults({
           <TableHead>
             <TableRow>
               <TableCell>Variant</TableCell>
-              <TableCell align='right'>
-                {metric.parameterType === MetricParameterType.Revenue
-                  ? 'Average cash per user (ACPU) interval'
-                  : 'Conversion rate interval'}
+              <TableCell align='right' className={classes.unitName}>
+                {unitName}
```
razvanpapadopol, line 583: I was looking through the snapshot diff and noticed this missing. Good to have snapshots even if they add a lot of noise with the classname changes 😬 

```suggestion
                {unitName} interval
```

src/components/explat/experiments/single-view/results/MetricAssignmentResults.tsx
```diff
@@ -682,13 +698,11 @@ export default function MetricAssignmentResults({
                   <TableCell>Variant</TableCell>
                   <TableCell align='right'>Users</TableCell>
                   <TableCell align='right'>
-                    {metric.parameterType === MetricParameterType.Revenue ? 'Cash Sales' : 'Conversions'}
+                    {metricParameterTypeName[metric.parameterType]}
```
razvanpapadopol, line 701: I think we really want the plural of `Conversion` here since we display counts in the table body below.

src/components/explat/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: Ah good spot, I thought had this taken care of

jessie-ross: I think it actually looks better without:
21036-explat-experiment

src/components/explat/experiments/single-view/results/MetricAssignmentResults.tsx
jessie-ross: Nice spot!

src/components/explat/experiments/single-view/results/MetricAssignmentResults.tsx
razvanpapadopol: Now that I see it in the UI, I think you're right 😄 
Thanks for the improvement!

razvanpapadopol (approved):
LGTM! Thanks for adding proper support for multiple metric types to the Metric Assignment Results while also cleaning up.

PR #846, opened 2023-02-09
Fix metric data tests

jessie-ross:
**This PR fixes data-validation tests for metrics to allow the new pipe metrics.**

## How has this been tested?


razvanpapadopol (approved):
Looks good! 
Tested locally with production data and I can confirm there are no more UI errors.

PR #847, opened 2023-02-10
Generalise stringifyMetricParams

jessie-ross:
**This PR generalises stringifyMetricParams.**

## How has this been tested?


razvanpapadopol (approved):
Thanks! 

I guess we could also update [this test](https://github.com/Automattic/abacus/blob/fix/stringify-metric-params/src/components/explat/metrics/multi-view/MetricsTable.test.tsx#L53).
Or even use the utility function there.

PR #851, opened 2023-03-01
Metrics: display tags in the metrics table

razvanpapadopol:
Fixes https://github.com/Automattic/experimentation-platform/issues/868

- adds `tags` to metric schema
- display tags in a raw format (comma separated) in the `MetricsTable`

## How has this been tested?

### DIFF
--- a/src/lib/explat/schemas.ts
+++ b/src/lib/explat/schemas.ts
@@ -114,6 +114,29 @@ export const extendedNumberSchema = yup
   // eslint-disable-next-line no-template-curly-in-string
   .test('is-number', '${path} is not a number', (value: unknown) => value === undefined || typeof value === 'number')
 
+export enum TagNamespace {
+  ExclusionGroup = 'exclusion_group',
+}
+
+export const tagBareSchema = yup
+  .object({
+    tagId: idSchema.defined(),
+    namespace: nameSchema.defined(),
+    name: nameSchema.defined(),
+    description: yup.string().defined(),
+  })
+  .defined()
+  .camelCase()
+export interface TagBare extends yup.InferType<typeof tagBareSchema> {}
+// For consistency and openness:
+export const tagFullSchema = tagBareSchema
+export interface TagFull extends yup.InferType<typeof tagFullSchema> {}
+export const tagFullNewSchema = tagFullSchema.shape({
+  tagId: idSchema.nullable(),
+})
+export interface TagFullNew extends yup.InferType<typeof tagFullNewSchema> {}
+export const tagFullNewOutboundSchema = tagFullNewSchema.snakeCase()
+
 export const eventSchema = yup
   .object({
     event: yup.string().defined(),
@@ -209,6 +232,7 @@ const noTestMetricSchema = yup
       then: metricPipeParamsSchema.defined(),
       otherwise: yup.mixed().oneOf([null]),
     }),
+    tags: yup.array(tagBareSchema),
   })
   .defined()
   .camelCase()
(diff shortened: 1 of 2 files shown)


src/lib/explat/schemas.ts
```diff
@@ -240,24 +264,29 @@ export const metricNewSchema = metricSchema.shape({
...
+          metricRevenueParamsSchema.snakeCase().cast(currentValue.revenue_params)
+        : undefined,
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      pipe_params: currentValue.pipe_params
+        ? // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+          metricPipeParamsSchema.snakeCase().cast(currentValue.pipe_params)
+        : undefined,
+    }),
+  )
+  .shape({
```
jessie-ross, line 287: I don't think this can be merged until after the backend changes have been added?

src/components/explat/metrics/MetricFormFields.tsx
```diff
@@ -71,6 +100,22 @@ const MetricFormFields = ({ formikProps }: { formikProps: FormikProps<{ metric:
...
+    data: tagOptions,
+    error: tagOptionsError,
+  } = useDataSource(async () => {
+    const tags = await TagsApi.findAll()
+    return tags
+      .filter((tag) => ELIGIBLE_METRIC_TAG_NAMESPACES.includes(tag.namespace))
+      .map((tag) => ({
+        name: tag.name,
+        value: tag.tagId,
+        meta: tag,
```
jessie-ross, line 114: Nit: I would call this data and not meta

src/lib/explat/metrics.ts
```diff
@@ -14,6 +14,8 @@ import {
 export const stringifyMetricParams = (metric: Metric): string =>
   JSON.stringify(metric[metricParameterTypeToParameterField[metric.parameterType]], null, 4)
 
+export const ELIGIBLE_METRIC_TAG_NAMESPACES: string[] = ['internal']
```
jessie-ross, line 17: This constant probably belongs in a `tags.ts` file, and should be an enum called `TagNamespaces`. We should use a metric-tags namespace rather than internal.

src/components/explat/metrics/MetricFormFields.tsx
```diff
@@ -1,16 +1,45 @@
...
-import { metricParameterTypeName } from 'src/lib/explat/metrics'
-import { metricParameterTypeToParameterField } from 'src/lib/explat/schemas'
+import { ELIGIBLE_METRIC_TAG_NAMESPACES, metricParameterTypeName } from 'src/lib/explat/metrics'
+import { AutocompleteItem, metricParameterTypeToParameterField, TagFull } from 'src/lib/explat/schemas'
+import { useDataLoadingError, useDataSource } from 'src/utils/data-loading'
+import { isDebugMode } from 'src/utils/general'
 
 import DebugOutput from '../../general/DebugOutput'
 
+function MetricTagsAutocomplete<Multiple extends boolean>(
```
jessie-ross, line 19: I don't understand why we are creating a new component for this, we should be able to use AbacusAutocomplete as we do elsewhere?

jessie-ross (changes requested):
This overall looks good but see comments for the requested change.

src/lib/explat/metrics.ts
razvanpapadopol: What namespace value to use for metric tags is still not defined so `internal` here is temporary. 

> We should use a metric-tags namespace

I think it can also be more generic as we may use the same `divisison-kpi-tags` also for tagging experiments 🤔 

I've added some placeholder values in D104318-code (in `class Tag_Namespace`) but I'm not sure if we'll use those or bring them into some kind of naming convention. This is going to be included in the [3rd phase of this project](https://github.com/Automattic/experimentation-platform/issues/864).
> Create metric tag name convention e.g. divison/experience like wpcom/ux-new-user-signup or wpcom/kpi-retention (@aaronyan)

src/lib/explat/schemas.ts
razvanpapadopol: Yes, this is correct and I've included the warning at the top of the PR description. The backend changes D104318-code that add the support for `tags` can be deployed first as they are compatible with the production UI.

src/components/explat/metrics/MetricFormFields.tsx
razvanpapadopol: Removed it completely with https://github.com/Automattic/abacus/pull/851#discussion_r1139062116

src/api/explat/MetricsApi.ts
```diff
@@ -28,7 +32,11 @@ async function put(metricId: number, newMetric: MetricNew): Promise<Metric> {
     throw new Error('Invalid metricId.')
   }
   const validatedNewMetric = await metricNewSchema.validate(newMetric, { abortEarly: false })
-  const outboundNewMetric = metricNewOutboundSchema.cast(validatedNewMetric)
+  const metricDataForRequest = {
```
jessie-ross, line 35: Maybe we should extract this into an outbound transform function.

jessie-ross (approved):
See comments, but this is looking good to me!

Thanks for all your work on this!

src/api/explat/MetricsApi.ts
razvanpapadopol: Good point! Done in https://github.com/Automattic/abacus/pull/851/commits/bb7864f777f931b0ecb0f9cda1377bd0505da3fa and also took the chance to fix a typescript error.

aaronyan (approved):
LGTM! Thanks Razvan for adding this.

src/api/explat/MetricsApi.ts
jessie-ross: Thanks!

PR #852, opened 2023-03-14
Metrics: Filter by tag in the metrics table

razvanpapadopol:
Follow up to https://github.com/Automattic/abacus/pull/851
Fixes part of https://github.com/Automattic/experimentation-platform/issues/870

This PR adds the functionality to filter metrics by tag in the Metrics Table.
* Uses `TagAutocomplete` to select a tag.
  - [ ] Should we allow multiple tags to be added in the filter field?
* The available tags should have the `namespace` value included in `ELIGIBLE_TAG_NAMESPACES`.
* Only in debug mode for now.

**Demo:**

Screenshot 2023-04-04 at 01 19 34

Screenshot 2023-04-04 at 01 25 11

## How has this been tested?

  - Follow instructions from https://github.com/Automattic/abacus/pull/851
  - `internal/explat_test_internal_metrics` tag is assigned to a metric


aaronyan:
Whatever we decide in https://github.com/Automattic/experimentation-platform/issues/871, we can implement in the Metrics directory too!

razvanpapadopol:
> Whatever we decide in [Automattic/experimentation-platform#871](https://github.com/Automattic/experimentation-platform/issues/871), we can implement in the Metrics directory too!

I agree we should aim for a high degree of consistency and reusability, also for the metric assignment modal for an existing experiment.

The only big difference is that here you get instant feedback in the merics table for whatever filter you've applied. In the wizard / assignment case you get to see the filtered list of metrics only when starting a new search.

jessie-ross:
Is this PR still relevant? Seems like we are using AbacusAutocomplete now, and we are implementing the filter also in https://github.com/Automattic/abacus/pull/854, which would be used in the metrics table too.

razvanpapadopol:
> Is this PR still relevant? Seems like we are using AbacusAutocomplete now, and we are implementing the filter also in #854, which would be used in the metrics table too.

That's right! Initially, https://github.com/Automattic/abacus/pull/854 was based on this PR but lots of things have change so I guess we'll do vice-versa.

PR #855, opened 2023-04-03
Wizard: filter metrics by tag

razvanpapadopol:
Fixes https://github.com/Automattic/experimentation-platform/issues/869

Renders tags as clickable Chip components to enable filtering in the Metrics step of the experiment wizard.
* Show selected (primary color) state for the active tag.
* Render loading state while fetching all tags.
* Use an "All" Chip as default state and to reset filtering.
* Only one tag can be applied at a time.
* Available only in debug mode.

**Demo:**

Screenshot 2023-04-04 at 02 38 22

## How has this been tested?

  * Apply diff D104318-code and sandbox the wpcom API
  * Open the Metrics step of the Experiment Wizard and filter by metric tag
     * When "All" Chip tag is selected, all metrics should be displayed
     * The "assigned" metrics should remain regardless of the selected tag


src/components/explat/experiments/wizard/Metrics.tsx
```diff
@@ -411,6 +425,30 @@ const Metrics = ({
...
+  const eligibleTagsForFilter = useMemo(
+    () => tags?.filter((tag) => DIVISION_KPI_TAG_NAMESPACES.includes(tag.namespace)) || [],
+    [tags],
+  )
+
+  const indexedMetricsForSearch = useMemo(
+    () =>
+      selectedTagOption
+        ? indexMetrics(
+            // TODO: Try to include the tag.metricIds already available on backend in the tags API response.
```
jessie-ross, line 443: I think this would be an unnecessary optimisation.

src/components/explat/experiments/wizard/Metrics.tsx
```diff
@@ -411,6 +425,30 @@ const Metrics = ({
   const [expandedMetricId, setExpandedMetricId] = useState<number | undefined>()
   const handleMetricDetailsClose = () => setExpandedMetricId(undefined)
 
+  const [selectedTagOption, setSelectedTagOption] = useState<TagFull | null>(null)
```
jessie-ross, line 428: We should go with an array and allow multiple tags to be selected.

jessie-ross (changes requested):
Thanks for these PRs!

Hmm, I think we should be using the multi-chip search selector for tags that we use for targeting:
Screenshot 2023-04-04 at 10 22 29 pm

See also my comment in the PI for overall thoughts on the UX:
https://github.com/Automattic/experimentation-platform/issues/864#issuecomment-1496050924

It might make things simpler if we had a top level selector in the header.

src/components/explat/experiments/wizard/Metrics.tsx
razvanpapadopol: Yeah, I guess we'll need a better case for including that list of ids 😬

src/components/explat/experiments/wizard/Metrics.tsx
razvanpapadopol: I've also considered the multi-tag select but with this straight-forward solution (clickable chips with no search) I don't think it's very useful and it introduces a bit of an UX puzzle: "do we have an AND or an OR relation between selected tags?". I guess it would be "OR" and a likely use case could be "I'm adding a cross-division experiment and want to search for existing "purchase_*" metrics, right?

razvanpapadopol:
Thanks for checking these out, @jessie-ross !! 

> Hmm, I think we should be using the multi-chip search selector for tags that we use for targeting

After exploring a solution using the search selector in https://github.com/Automattic/abacus/pull/855 I've started to enjoy the simplicity of this one. If we manage to have a limited number or curated tags it's perfect for wizard. To filter the tags if they are too many, I guess we can use the tag (or its namespace) with the selected `platform`.

> It might make things simpler if we had a top level selector in the header.

I'm note sure if a top level selector will achieve this level of simplicity and it may cause some confusion for new users. Eg: I've set it to filter metrics directory for some division and then when I create an experiment I don't have a local cue that a filter has been applied. However, I like the idea of persisting last user selection locally ❤️

aaronyan:
I've commented on a proposed unified design [here](https://github.com/Automattic/experimentation-platform/issues/871#issuecomment-1496838448)!

razvanpapadopol:
Closing this in favour of https://github.com/Automattic/abacus/pull/854 based on all the discussions in https://github.com/Automattic/experimentation-platform/issues/871

PR #856, opened 2023-05-01
Remove storybook

jessie-ross:
**This PR removes the Storybook from the repository.**

Storybook was initially added to the repository to help with the development and testing of components. However, it has had limited use and isn't a major benefit for Abacus development, for the best development experience we should be using real data via the documented `REACT_APP_PRODUCTION_CONFIG_IN_DEVELOPMENT=true npm run dev`.

At the same time storybook isn't free, it is entangled with the build process and build dependencies. While developing https://github.com/Automattic/experimentation-platform/issues/406 (manual experiment start/stop), I've had to upgrade Yup which meant I needed to upgrade Typescript which meant I needed to upgrade etc. etc. There is a lot of upgrades needed and storybook is in the way, and it is most worthwhile to just remove it - we aren't a big enough project/don't have enough resources for it to be worthwhile maintaining.

The PR includes the removal of Storybook configuration files, dependencies, and related scripts.

Please review the changes and let me know if there are any concerns or questions.

Part of https://github.com/Automattic/experimentation-platform/issues/406

## How has this been tested?


package-lock.json
jessie-ross, line 1: As you can see from the amount of additions/deletions, storybook is for some reason an incredibly complex and involved set of packages, bringing in an inordinate number of transitive dependencies and is affects a lot of the other packages we have going on.

razvanpapadopol (approved):
Code changes look good and the build and tests seem happy ✅ 
Thanks for the cleanup!

> There is a lot of upgrades needed and storybook is in the way, and it is most worthwhile to just remove it - we aren't a big enough project/don't have enough resources for it to be worthwhile maintaining.

I can see that the costs outweigh the benefits in our case. If something changes in the future, we can add it back at a relatively low cost compared to continuous maintenance.

Issue #614, opened 2021-12-09
Move e2e tests to Teamcity


jessie-ross:
Edit: We've decided to disable the e2e tests for now and will move the mover to Teamcity later (p7H4VZ-2Bs-p2).

It seems like CircleCI OOMs are failing our E2E tests.
(Although I haven't confirmed it is an OOM.)

Two options here:
- Decrease memory usage, I think there is a variable for Create React App that we could tune? Maybe something else?
- Increase CircleCI memory, not sure if this is doable...

aaronyan:
Thanks for trying to fix the CircleCI e2e tests @is4ac! @jessie-ross and I discussed this during our 1-1 today, and we decided to disable these these tests for now, and we'll move them over to Teamcity later (p7H4VZ-2Bs-p2). Editing this issue to reflect this new direction!

jessie-ross:
In the spirit of further simplification I think we should remove these e2e tests:
- Further complicates our dependencies, by a lot, puppeteer is very heavy.
- Provides no added value - it doesn't actually tell us if there is an issue with production, it instead tests a different environment.
- We don't need to ensure 99.999% uptime, it is ok if production fails temporarily.

jessie-ross:
For context, these tests just spin up an mock Abacus, and clicks the navigation links, that's basically it.

jessie-ross:
Since I was already working on our dependencies I've created a removal PR here:
https://github.com/Automattic/abacus/pull/858

PR #858, opened 2023-05-03
Remove E2E testing

jessie-ross:
**This PR removes the unused E2E testing.**

See https://github.com/Automattic/abacus/issues/614:

> In the spirit of further simplification I think we should remove these e2e tests:
>
> - Further complicates our dependencies, by a lot, puppeteer is very heavy.
> - Provides no added value - it doesn't actually tell us if there is an issue with production, it instead tests a different environment.
> - We don't need to ensure 99.999% uptime, it is ok if production fails temporarily.
>
> For context, these tests just spin up an mock Abacus, and clicks the navigation links, that's basically it.
>
> Overkill for a small project like this.

## How has this been tested?


razvanpapadopol (approved):
Yes, the E2E testing here looked like only boilerplate code anyway. Aside from the linting issue, LGTM!

jessie-ross:
Seems like the 0 day impact discrepancy is flaky and depends on the time of day run, I will fix it in this PR.

PR #862, opened 2023-05-10
Upgrade yup to 1.1.1

jessie-ross:
**This PR upgrades yup to 1.1.1**
- Uses the new yup typescript system where we create interfaces and then confirm it matches the type created from yup. This is cleaner and simpler (more maintainable) than the old system as there is better separation between yup and typescript, and adds an extra control for code correctness - these are effectively tests for the types.
- Adds a `transformSchemaFieldNames` function so we now have correct schemas for outbound data, this is another added benefit but was necessary due to the new internal order of transformation within yup.

Part of :
- #406
  - Necessary to fix a bug in the change.
- #687

## How has this been tested?

- [x] Manual testing creation/editing


razvanpapadopol (commented):
Hey @jessie-ross ! Thanks for kicking off an update to yup, our schemas and schema validations!

When trying to add or update metrics, I get the following validation errors even if the entered values are correct:
-  `"metricId": "This field is required"`
- for Conversion metrics: 
``"eventParams": "metric.eventParams must be a `array` type, but the final value was: `\"[]\"`."``
- for Cash Sales metrics I get the `"This field is required"` error for all of the params (`refundDays`, `productSlugs`, `transactionTypes`).

Can you reproduce these or it's some kind of problem on my local install? I should mention that TypeScript isn't complaining and I have the packages correctly installed, including `"yup": "1.1.1"`.

jessie-ross:
@razvanpapadopol thanks for testing and finding the issue with metrics, it looks like Yup's internal casting/transform order has changed a bit and also now we need to explicitly state that we can accept JSON strings for the parameters. I've just updated the PR with a fix. Should be good to go now.

razvanpapadopol:
> @razvanpapadopol thanks for testing and finding the issue with metrics, it looks like Yup's internal casting/transform order has changed a bit and also now we need to explicitly state that we can accept JSON strings for the parameters. I've just updated the PR with a fix. Should be good to go now.

Works fine now and the changes look good! I'm glad it was just that 🎉 

Now I have one more issue 😬 . The experiments directory crashes when fetching the list of experiments in "debug mode" with this error: `DataLoadingError Experiment Error: Invalid originalValue for date`.

PR #865, opened 2023-05-12
Update analysisDatetime < startDatetime check

razvanpapadopol:
Fixes p1683914537625449-slack-C7HH3V5AS

* allow `experiment.startDatetime` to be up to 1 day later than the `analysis.analysisDatetime` date.

## How has this been tested?
21190-explat-experiment shouldn't crash.


aaronyan (approved):
LGTM. Thanks for the quick fix, @razvanpapadopol!

aaronyan:
What do you think about rounding the dates so that the timestamp is gone and then allowing `analysisDate <= startDate`? We should also consider adding some unit tests for this case. What do you think about this, @razvanpapadopol?

razvanpapadopol:
> What do you think about rounding the dates so that the timestamp is gone and then allowing analysisDate <= startDate?

Something like this, right? https://github.com/Automattic/abacus/pull/866
It looks good to me. I assume the analysis time is hardcoded at 00:00:00 because the only scenario that may break that approach is a value like 23:59:59 with `[day - 1]`😅 

>  We should also consider adding some unit tests for this case.

Yes, I agree! I'll keep it in draft until adding test coverage.

PR #866, opened 2023-05-12
Update analysisDatetime < startDatetime check - part 2 

razvanpapadopol:
Use `utcStartOfDay` of the `experiment.startDateTime` when comparing to `analysis.analysisDatetime` so it's allowed to be up to one day after the latter.

Related to https://github.com/Automattic/abacus/pull/865#issuecomment-1546272918

- [x] Add unit tests to check for this scenario

## How has this been tested?


src/lib/explat/analyses.ts
```diff
@@ -563,9 +563,9 @@ export function getAnalysisRunHours(analysis: Analysis, experiment: ExperimentFu
   }
 
   // istanbul ignore next; shouldn't occur
-  // Add 1 day to this check because it was crashing the results page in the first day of the experiment.
-  // See: https://a8c.slack.com/archives/C7HH3V5AS/p1683914537625449
-  if (add(analysis.analysisDatetime, { hours: 24 }) < experiment.startDatetime) {
+  // Use the date, allowing the experiment to start in the first 24h after the 'analysis_datetime' returned by the backend.
+  // This should prevent crashing the results page in the first day of the experiment (https://a8c.slack.com/archives/C7HH3V5AS/p1683914537625449)
+  if (analysis.analysisDatetime.getDate() < experiment.startDatetime.getDate()) {
```
aaronyan, line 568: Should this be `<=`?

razvanpapadopol: As it is right now, it allows the values to be equal (11 for 21190-explat-experiment at the time of writing), which is what we want. It will `throw` if the start date is sometime tomorrow.

razvanpapadopol: I've changed the approach to use `utcStartOfDay` to 'round' the experiment start time to the beginning of the day and we'll compare timestamps as it's the sensible way to compare dates (or datetimes) in JS. The day of month will break on the 1st of next month and day of year will break on 1st of January 😁

aaronyan: Thanks for the info!

jessie-ross (approved):
Thanks for finding and fixing this bug!
This LGTM but also your original fix seems good enough and perhaps leaves things less fragile.

razvanpapadopol:
@aaronyan, I've added the [requested](https://github.com/Automattic/abacus/pull/865#issuecomment-1546272918) test coverage but I'm still not sure if to go ahead with updating the check as it will bring back some of the previous [fragility](https://github.com/Automattic/abacus/pull/866#pullrequestreview-1426313276).

Basically, with this PR, we're expecting the experiment `startDatetime` to be [on the same day](https://github.com/Automattic/abacus/pull/866/files#diff-c1eebfaae4b0e18b462ce4973ba22a613672b8ae7f84b4ee5a8ec83e9b6ca5bdR657) as the `analysisDatetime` while currently, with the original fix, we allow a 24h interval between the two datetimes.

aaronyan (approved):
Thanks @razvanpapadopol! LGTM.

I prefer to be slightly more strict here and break when the analysis datetime is before the start datetime. Thanks for double checking!

PR #867, opened 2023-05-15
Remove experiment only

jessie-ross:
**This PR removes the ability to schedule experiments from the wizard.**
- Removes scheduling.
- Fixes the schema to match.
- Allows changing the end date of an experiment to null in the general-panel-edit-modal.
- Removes unneeded tests.

**This PR should only be merged once a "Complete" button has been added.**

## How has this been tested?


jessie-ross:
@aaronyan, for simplicity I've decided to go without the enddate-popup on manual start - Experimenters can always specify enddates via the General Panel edit experiment modal.

jessie-ross:
We'll need to find a solution for existing experiments. They will continue to be run automatically as before, but after this PR is merged they will no longer be able to be edited.

I propose we update staging experiments so that they have null start/end datetimes, keeping a record of what they are, contacting experimenters that have start dates very soon.

aaronyan:
> Allows changing the end date of an experiment to null in the general-panel-edit-modal.

> for simplicity I've decided to go without the enddate-popup on manual start - Experimenters can always specify enddates via the General Panel edit experiment modal.

I don't think we should allow null end dates. It's important to require an experiment duration popup to prevent experiments from running forever. 

Before merging, I think it is important to add a stacked PR adding the experiment duration popup to set the start and end date on the experiment launch. What additional work is required to support this?

> - Upon manual start, a popup comes up with a box asking for experiment duration in days
> - Default show 14 days, but editable
> - Upon start, the start and end date for the experiment are filled out

aaronyan:
> We'll need to find a solution for existing experiments. They will continue to be run automatically as before, but after this PR is merged they will no longer be able to be edited.
> 
> I propose we update staging experiments so that they have null start/end datetimes, keeping a record of what they are, contacting experimenters that have start dates very soon.

I like this plan!

jessie-ross:
> I don't think we should allow null end dates. It's important to require an experiment duration popup to prevent experiments from running forever.

I don't think an end-date solves the problem of experiments running forever, we already have experiments running forever:
- 20854-explat-experiment
- 20748-explat-experiment

I think a lot of experimenters select far off end-dates, and then they conclude their experiments manually, so we can't see currently how many in the past have been set with far off end dates.

And we get the same problem of automated deploy if someone does forget about it - there will be a sudden change and no-one is around while it happens.

razvanpapadopol:
> experiments running forever

Leaving a 20% / 10% / 1% holdout group for a long term impact measurement on a low traffic user flow is something that we're supporting indeed. But the 14 days suggested runtime can be edited to 365 for example.

What I think @aaronyan is trying to highlight is that we are sometimes running high-impact experiments such as 21190-explat-experiment which may ultimately affect the company's financials if left running for longer than originally planned and I don't see a universal solution to address this (eg: Slack pings after 14 days) especially because we support such a wide range of experiments.

> And we get the same problem of automated deploy if someone does forget about it - there will be a sudden change

It might be a potential problem for brittle E2E tests that were added/updated with the experiment. However, all users getting assigned to `control` which was already a live experience shouldn't be problematic. Also, I don't see any risk or increase load from HEs.

aaronyan:
> we already have experiments running forever

True in sentiment, but they still have an end date so the experiment will auto-turn off at some point.

If someone wants to run an experiment forever, I want them to manually need to continue adjusting the end date. If an experiment ends prematurely, the experimenter responsible will hopefully learn how it works and remember next time to set a proper duration or extend if needed.

> And we get the same problem of automated deploy if someone does forget about it - there will be a sudden change and no-one is around while it happens.

Great point about wanting someone to be around when an experiment ends (the same way we want someone to be around when an experiment starts). Although, I don't see many risks for having an end date auto-transition when no one is around:

- When an experiment auto-transitions to `complete`, there will be two weeks of buffer where the users experience the same variants. 
- When the experiment auto-transitions to `disabled`, I don't see much risk in all of the users transitioning to `control`, since some % of users are already experiencing `control` (as @razvanpapadopol [mentions](https://github.com/Automattic/abacus/pull/867#issuecomment-1550350351)).

I feel pretty strongly about requiring an end date and it sounds like you feel strongly about wanting to manual experiment end. Should we hop on a call and discuss to get aligned on this, @jessie-ross?

jessie-ross:
Since it is a large change, I've added it to a separate child PR:
https://github.com/Automattic/abacus/pull/871

aaronyan (approved):
Thank you @jessie-ross! LGTM

PR #864, opened 2023-05-11
Metrics: add toggle to show/hide archived and debug metrics

razvanpapadopol:
Fixes https://github.com/Automattic/experimentation-platform/issues/912

* Hide `'archived_*'` metrics by default in the Metrics Table. For metric assignments, they are already hidden.
* Add a toggle (checkbox) to the top-right, above the table to allow showing archived metrics (for documentation, reviewing old experiments, etc).
* Add another toggle to allow showing debug (test) metrics (for viewing, testing experimental metrics).
* Add test coverage for filtering out archived and debug metrics and for the show/hide toggles functionality.

Screenshot 2023-05-16 at 02 32 10

## How has this been tested?

- Automated tests that cover added/changed functionality
- Manual testing with production data (locally)
- Additional manual testing instructions (please specify):
  * Checking the "Show archived metrics" checkbox should display also the archived metrics in the Metrics table.
  * Checking the "Show debug metrics" checkbox should display also the debug metrics in the Metrics table.
  * Search state should remain unchanged.

PR #869, opened 2023-05-15
Hotfix: Include debug metrics in the experiment view

### LINKED ISSUE #848: UI crashing because of 'explat_test_'  metric assigned
## Context

UI crashing for a specific experiment - 20938-explat-experiment as mentioned in 
p1676386757691059/1676317539.612699-slack-C7HH3V5AS

When in debug mode, the experiment works fine.

## Current Behavior vs Expected Behavior

Like every other experiment, it should load.

This one stops loading with these error messages as tested:

* Overview:
`Failed to lookup metric with ID XXX for assignment with ID XXXX.`

* Results:
`Error: Could not find matching unit-type for metric and derivations.`

## Possible Solution

Check the latest changes related to introducing pipe params and see why it works in `debug` mode.

LE: Fetching all metrics (add `?debug` parameter) for experiments page (Overview and Results tabs) and filtering out by name excluding `explat_test_` seems like a good fix.
*Removing completely the `?debug` filtering on the backend and using the same client-side filtering also on Metrics page and Experiment Wizard could be a potential simplification.

jessie-ross:
**This PR loads debug metrics for the experiment view.**

This is a hotfix for 21214-explat-experiment, it does mean that debug metrics will be included in the assign metric modal for running experiments, but we can fix that later.

Resolves #848

## How has this been tested?


razvanpapadopol (commented):
The code changes to load all metrics in the experiment view look good to me!

As a potential improvement, we may consider hiding the debug metrics by default and, if `some` are assigned we can add a toggle to show them, similar to the [metrics directory](https://github.com/Automattic/abacus/pull/864#pullrequestreview-1426296794).

> debug metrics will be included in the assign metric modal for running experiments, but we can fix that later.

I've created https://github.com/Automattic/abacus/pull/870 to address this part.

PR #875, opened 2023-05-24
Experiment: add route flexibility

razvanpapadopol:
This is a potential solution for https://github.com/Automattic/experimentation-platform/issues/957 to prevent an app crash when the URL param doesn't exactly match the experiment `/:view`, by rendering the default view and logging the validation error to the console.

## How has this been tested?

  * add `/bad-param` to an experiment URL and the default view should be loaded (eg: for status `running` => `/results` view).


razvanpapadopol:
> I don't see this as a good solution as we are just hiding errors into the console, and moving the issue downstream by making further down code accept null or invalid views.
> 
> Instead I suggest fixing the routing table itself and adding a 404 page: See #876

Thanks for having a look, @jessie-ross !

Indeed, when I've scanned other routes to see if they need any sort of error handling, I couldn't find any. So probably having the experiment `:view` routes [explicitly defined](https://github.com/Automattic/abacus/pull/876/files#r1205540388) is the only needed fix for https://github.com/Automattic/experimentation-platform/issues/957.

razvanpapadopol:
Closing this in favor of https://github.com/Automattic/abacus/pull/876

PR #877, opened 2023-05-25
Experiment wizard: remove documentation step

razvanpapadopol:
Removes the first step of the Experiment Wizard. Since we no longer require the P2 URL, we shouldn't instruct the experimenters to start with the documentation and instead start designing in Abacus.
* Renamed 'Basic Info' step to 'Start'.
* Moved the Documentation field to the end of the new first step.
* Updated copy for the "Submit" step to say what's required at launch. 

Initially, I considered this change as an immediate follow-up to https://github.com/Automattic/experimentation-platform/issues/946 but it should be feasible to ship it right away, before making the announcement. 

Partially fixes https://github.com/Automattic/experimentation-platform/issues/754

#### Demo
Screenshot 2023-06-01 at 04 04 54

<details>
<summary>Previous demo</summary>
Screenshot 2023-05-25 at 23 16 19
</details>

#### To do
- [x] Update the copy where needed if we choose to proceed with this.
- [x] Fix unit tests.

## How has this been tested?

   * The wizard should have only 4 steps now and the first step should include the P2 post URL optional field.
   * In the last step you should learn about the required information at launch (https://github.com/Automattic/abacus/pull/877#discussion_r1206154936)


src/pages/explat/experiments/ExperimentWizard.tsx
```diff
@@ -154,7 +154,7 @@ export default function WizardEdit({
     )
 
   const titleByExperimentWizardMode: Record<ExperimentWizardMode, string> = {
-    [ExperimentWizardMode.Create]: 'Create an Experiment',
+    [ExperimentWizardMode.Create]: 'Create an Experiment Draft',
```
aaronyan, line 157: I like the idea of this, but I'm not sure we want to introduce the term "draft". 

What do you think about sticking with "Create an Experiment" and we mention at the "submit" step that the experiment will be in `staging` and requires two things to launch:

- An end date
- A p2 URL of the experiment documented and reviewed

src/components/explat/experiments/wizard/BasicInfo.tsx
```diff
@@ -124,6 +145,41 @@ const BasicInfo = ({
...
+
+      <div className={classes.row}>
+        <FormControl component='fieldset'>
+          <FormLabel>Document your experiment</FormLabel>
+          <FormHelperText>
+            {' '}
+            We think one of the best ways to prevent a failed experiment is by documenting what you hope to learn.
+          </FormHelperText>
+          <Alert severity='info'>
+            Once an experiment draft is created, the next step is{' '}
```
aaronyan, line 157: In a future PR, we can change this info alert to say that submitting an experiment will prepare copyable text to be used in an experiment p2 draft.

aaronyan (commented):
Thanks for thinking about the UX, @razvanpapadopol! I've added a few alternative ideas for us to consider.

src/components/explat/experiments/wizard/BasicInfo.tsx
razvanpapadopol: If we conclude that the overall approach of this PR looks good, I'd also update the text as it will be merged anyway after adding the feature that prepares P2 draft to copy.

src/components/explat/experiments/wizard/BasicInfo.tsx
```diff
@@ -124,6 +145,43 @@ const BasicInfo = ({
...
+        <FormControl component='fieldset'>
+          <FormLabel>Document your experiment</FormLabel>
+          <FormHelperText>
+            {' '}
+            We think one of the best ways to prevent a failed experiment is by documenting what you hope to learn. This
+            will also help future experiments as your documented design and results may get featured in an upcoming
+            Experiments Learnings Library!
+          </FormHelperText>
+          <Alert severity='info'>
+            Once an experiment is submitted, Abacus will automatically prepare the copyable P2 post content to start{' '}
```
aaronyan, line 159: What do you think about this text?

"After submitting your experiment, Abacus will provide a pre-filled p2 experiment template in the `Setup` tab to get you started on experiment documentation.

aaronyan (approved):
LGTM so far! I've also offered another perspective on how we want to present the URL requirements and expectations.

jessie-ross:
Nice! I love simplifying things.

I think we can go further:

241056766-edec835b-5477-413f-8def-b26d7d20608a

- Removing the form-part header, and the scary helper text.
- Remove the info box - optionally we could move it below the input.
- Adds simple text explaining it isn't required until launch.

jessie-ross:
Ah sorry, I hadn't read Aaron's comments which seem to cover most of what I've mentioned.

aaronyan (approved):
LGTM as long as the tests pass!

PR #878, opened 2023-05-27
Experiment setup: Add copyable P2 post draft

razvanpapadopol:
Fixes https://github.com/Automattic/experimentation-platform/issues/950

#### Notes:
* The copy button may be repositioned, for example to the top-right where the "Edit" and "Assign metric" buttons are on the Overview tab.
* Target audience documentation is very simplistic (Yes/No). Added a comment in code.
* The exposure events are not used at the moment (not present in the P2 template).
* We're re-introducing private links even if they are not clickable. However, shortening is [not making much difference](https://github.com/Automattic/abacus/pull/802#issuecomment-1331746356).

#### Demo:
Screenshot 2023-06-01 at 03 36 23
<br><br>
<details>
<summary>Previous demos</summary>
Screenshot 2023-05-31 at 23 00 01
Screenshot 2023-05-27 at 04 30 09
</details>

#### To do:
- [x] Add / update tests.
- [ ] (post-merge) Update FG mentions from `code-setup` to `setup`.

## How has this been tested?

  - When testing locally, a Chrome flag should be used in order for the _Copy to clipboard_ action to work. See https://github.com/Automattic/abacus/pull/740#discussion_r908715048


src/components/explat/experiments/single-view/ExperimentSetup.tsx
```diff
@@ -0,0 +1,317 @@
...
+          P2 design post
+        </Typography>
+        <br />
+        <Typography variant='body1'>
+          Below you can find the P2 design post prepared for your experiment, ready to be used in an{' '}
+          <PrivateLink href='https://wp.me/bxNRc'>a8cexperiments P2 draft</PrivateLink>.
+        </Typography>
+        <pre className={classes.pre}>
+          <div className={classes.clipboardButtonWrapper}>
+            <ClipboardButton text={p2PostContent} targetName='P2 design post' label='Copy all P2 blocks' />
```
aaronyan, line 283: When testing on my local env, the copy button doesn't seem to work for me. This could be an issue with my local dev env though.

aaronyan: What do you think about changing the label to "Copy experiment design"?

src/components/explat/experiments/single-view/ExperimentSetup.tsx
```diff
@@ -0,0 +1,317 @@
...
+  return (
+    <>
+      <Paper className={classes.p2DesignPaper}>
+        <Typography color='textPrimary' variant='h3'>
+          P2 design post
+        </Typography>
+        <br />
+        <Typography variant='body1'>
+          Below you can find the P2 design post prepared for your experiment, ready to be used in an{' '}
+          <PrivateLink href='https://wp.me/bxNRc'>a8cexperiments P2 draft</PrivateLink>.
```
aaronyan, line 279: This link is not underlined for some reason.

aaronyan (commented):
Great start, @razvanpapadopol! It would be nice to include these in the experiment p2 draft as well:

- Hypothesis: use the experiment description. We can consider even changing the Experiment Wizard field to read "Hypothesis" instead of "Experiment description", since having the experiment hypothesis documented in the Abacus model will probably be most effective.
- Metrics: Remove the bullets on "More success metrics" and add the minimum practical difference values.
- Add bulleted instructions on how to use the copyable experiment template code since it requires going into the "code editor".
- Add a button on the setup page that directs to the a8cexperiments new post as a new tab. We want to have a call to action for the experimenter to document their experiment right after the submitted an experiment in Abacus.

src/components/explat/experiments/single-view/ExperimentSetup.tsx
```diff
@@ -0,0 +1,317 @@
...
+  const targetAudience = experiment.segmentAssignments.length ? 'Custom' : 'All'
+
+  const existingUsers = formatBoolean(experiment.existingUsersAllowed)
+
+  const p2PostContent = `
+  <!-- wp:heading {"level":1,"className":"wp-block-heading","anchor":"experiment-experiment-title"} -->
+  <h1 class="wp-block-heading" id="experiment-experiment-title">Experiment: ${experiment.name}</h1>
+  <!-- /wp:heading -->
+  
+  <!-- wp:a8c/editor-notes {"notes":"\\u003cmark style=\\u0022background-color:#7bdcb5\\u0022 class=\\u0022has-inline-color has-black-color\\u0022\\u003e\\u003cstrong\\u003eOverwhelmed?\\u003c/strong\\u003e Ping us on #a8c-experiments to arrange a run through :-)\\u003c/mark\\u003e"} /-->
```
razvanpapadopol, line 86: Because of how React is automatically escaping Unicode sequences used in the `wp:a8c/editor-notes` content, I had to [manually escape them in code by using double backslashes](https://github.com/Automattic/abacus/pull/878/commits/29db4679e9dae911d2a4e8a8f4c09167e64dc778). I couldn't find this approach documented somewhere because I guess it's quite an edge case.

jessie-ross: It seems like we need to give the block editor strings escaped strings - the `\u003c` for `<` - we can see that by inspecting what we copy originally from the block editor.

Since we want to produce strings with the escape codes in them, we need to make sure we escape the backslashes from javascript backtick-string which is why we need the second backslash.

I think what you've done is the best way to go about it, we should just leave a comment explaining with instructions how to update the template however.

src/components/explat/experiments/single-view/ExperimentSetup.tsx
```diff
@@ -0,0 +1,317 @@
...
+        <br />
+        <Typography variant='body1'>
+          Below you can find the P2 design post prepared for your experiment, ready to be used in an{' '}
+          <PrivateLink href='https://wp.me/bxNRc'>a8cexperiments P2 draft</PrivateLink>.
+        </Typography>
+        <pre className={classes.pre}>
+          <div className={classes.clipboardButtonWrapper}>
+            <ClipboardButton text={p2PostContent} targetName='P2 design post' label='Copy all P2 blocks' />
+          </div>
+          <code>{p2PostContent}</code>
```
jessie-ross, line 285: I think we should hide the code.

src/components/explat/experiments/single-view/ExperimentSetup.tsx
```diff
@@ -0,0 +1,317 @@
...
+  
+  <!-- wp:paragraph -->
+  <p><strong>Crosspost and CC:</strong><br>Recommended: &lt;your team P2&gt;, &lt;stakeholders&gt;, &lt;wpcomhappy.wordpress.com, if applicable&gt;, etc.</p>
+  <!-- /wp:paragraph -->`
+
+  return (
+    <>
+      <Paper className={classes.p2DesignPaper}>
+        <Typography color='textPrimary' variant='h3'>
+          P2 design post
```
jessie-ross, line 274: This should be title case.

jessie-ross (approved):
The code itself looks good to me, see Aaron and my comments for suggested UI changes.

We also need to fix the redirect after creating an experiment - right now it redirects to code-setup which is broken.

src/components/explat/experiments/single-view/ExperimentSetup.tsx
razvanpapadopol: Good point! By manually copying the content, the formatting gets included as well and an additional step is required (what Aaron mentioned here).

> Add bulleted instructions on how to use the copyable experiment template code since it requires going into the "code editor".

This is not an issue when using the "Copy experiment design" button since no formatting is applied and the copied text can be dropped directly in the Visual editor.

src/components/explat/experiments/single-view/ExperimentSetup.tsx
razvanpapadopol: > When testing on my local env, the copy button doesn't seem to work for me.

The functionality is working in production because of the secure context. Locally we can use Chrome flag `--unsafely-treat-insecure-origin-as-secure` to 'whitelist' Abacus local URL. Updated testing instructions with a [link to a previous discussion](https://github.com/Automattic/abacus/pull/740#discussion_r908715048) for more context.

src/components/explat/experiments/single-view/ExperimentSetup.tsx
aaronyan: > This is not an issue when using the "Copy experiment design" button since no formatting is applied and the copied text can be dropped directly in the Visual editor.

Oh neat! Glad to hear that this is how the copy-paste interacts with the p2 visual editor.

razvanpapadopol:
@aaronyan, @jessie-ross, thanks for reviewing this new feature and for the great ideas and improvements! [Nice catch](https://github.com/Automattic/abacus/pull/878#pullrequestreview-1453730355) of the redirect bug as well 😁 

I've pushed now the changes that should cover everything that was discussed, including the move of some text from the Wizard mentioned in https://github.com/Automattic/abacus/pull/877#discussion_r1210974291

aaronyan (approved):
Looks great, Razvan! Thanks for making those adjustments and iterating so quickly.

Here are a few more minor spacing recommendations:
- Let's make the button look like the other buttons (similar vertical padding as the Launch and Complete buttons)
- I think it would be nice to have some space between the clipboard icon and the word "Copy"
- Let's make the vertical spacing above and below the button equal

razvanpapadopol:
> Let's make the button look like the other buttons (similar vertical padding as the Launch and Complete buttons)

They are all using the same `Button` component with the `outline` variant so the padding is the same. I've now switched rendering the icon [according to the docs](https://mui.com/material-ui/react-button/#buttons-with-icons-and-label) so it should align better.


> I think it would be nice to have some space between the clipboard icon and the word "Copy"
Let's make the vertical spacing above and below the button equal

✅

aaronyan (commented):
Looks great, thanks for making those changes!

PR #880, opened 2023-06-02
Experiment P2 draft: remove primary from guardrail metrics

razvanpapadopol:
We shouldn't display the primary metric twice in the pre-filled P2 experiment template. 
This PRs filters it out when rendering the list of 'guardrail metrics'.

## How has this been tested?


aaronyan (approved):
Nice catch, thanks Razvan. LGTM!

PR #881, opened 2023-06-09
Experiments directory: async load analyses to render participants

razvanpapadopol:
* Add the 'analyses' query param to the 'GET /experiments' API call
* Update schema to make analyses optional for ExperimentSummary
* Display Skeleton for 'participants' column while loading analyses
* Table columns should have fixed width

Temporarily, use D113058-code for testing.

## How has this been tested?


src/api/explat/ExperimentsApi.ts
```diff
@@ -126,12 +127,12 @@ async function assignMetric(
  *
  * @throws UnauthorizedError
  */
-async function findAll(): Promise<ExperimentSummary[]> {
+async function findAll(shouldIncludeAnalyses = false): Promise<ExperimentSummary[]> {
+  const query = qs.stringify({ analyses: shouldIncludeAnalyses, debug: isDebugMode() })
```
jessie-ross, line 131: I think these booleans need to be converted to strings.

jessie-ross (approved):
See comment, but LGTM if we can't improve this elsewhere.

aaronyan:
Looks great, thank you so much @razvanpapadopol for making these changes happen so fast.

Is it possible to make the loading bar float on top of the window? Once it finishes loading, I want to avoid the contents below the loading bar from "jumping" upwards.

aaronyan (approved):
🎉 🚀

Issue #884, opened 2023-06-28
Have analysis method be encoded in the URL state


jessie-ross:
From p1687883275481119/1687160385.770289-slack-C7HH3V5AS

We should have the analysis method be stored in the URL state:
e.g. `/experiments/1000-experiment-name/results?analysis-method=assigned`

This would allow results to be shared and linked to the correct method.

razvanpapadopol:
Hey, @jessie-ross ! I was working at something similar just now so managed to quickly come up with this solution: https://github.com/Automattic/abacus/pull/885

PR #885, opened 2023-06-29
Persist selected analysis method in the URL

razvanpapadopol:
Fixes https://github.com/Automattic/abacus/issues/884

This PR adds support for the analysis method to be encoded in the URL state.
* Adds newly selected analysis strategy to URL as a query param 
* Use the `analysis-method` query param to pre-select the analysis method and display a success message as shown below.
* If the value for `analysis-method` query param is not a valid analysis strategy, show an error message as shown below, without crashing the UI.

Screenshot 2023-06-29 at 07 59 31

Screenshot 2023-06-29 at 10 30 18

## How has this been tested?


src/components/explat/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -235,9 +237,35 @@ export default function ExperimentResults({
...
+
+  const [strategy, setStrategy] = useState<AnalysisStrategy>(() => {
+    if (initialStrategyFromUrl) {
+      try {
+        const initialStrategy = yup
+          .string()
+          .oneOf(Object.values(AnalysisStrategy))
+          .defined()
+          .validateSync(initialStrategyFromUrl)
+        enqueueSnackbar(`'${Analyses.AnalysisStrategyToHuman[initialStrategy]}' analysis is selected!`, {
```
jessie-ross, line 255: I think we can be silent if it works.

src/components/explat/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -235,9 +237,35 @@ export default function ExperimentResults({
...
+          variant: 'success',
+        })
+        return initialStrategy
+      } catch (e) {
+        enqueueSnackbar(`Selecting '${initialStrategyFromUrl}' analysis failed!`, { variant: 'error' })
```
jessie-ross, line 260:
```suggestion
        enqueueSnackbar(`Selecting '${initialStrategyFromUrl}' analysis from query parameter failed!`, { variant: 'error' })
```

jessie-ross (approved):
Thanks for the quick turnaround on this!

See comments but looks great! 🚢

src/components/explat/experiments/single-view/results/ExperimentResults.tsx
razvanpapadopol: I've added this to compensate for a missing visual cue that a method other than the recommended one is selected. People are not usually checking the query params when clicking a link or when sharing a link so any hint in the UI should be useful in this case.

Issue #889, opened 2023-07-18
Fix failing build error


jessie-ross:
From p1689045920246949-slack-CNTD4G117

jessie-ross:
Systems request: pMz3w-hYB-p2

jessie-ross:
Resolved by #890

aaronyan:
Thank you so much, @jessie-ross!

PR #890, opened 2023-07-18
Update production node

jessie-ross:
Likely fix for #889

## How has this been tested?


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Use long term support version of Node.js
-FROM node:14.16.1
+FROM node:16.18.1
```
withinboredom, line 2: Might as well go for LTS...

aaronyan: Woah, hi there @withinboredom! @jessie-ross, what would be required to move to [v18.12.0 (LTS)](https://nodejs.org/en/blog/release/v18.12.0)?

jessie-ross: Haha hi!!!

If we move to LTS we'll run into this again ;)

Screenshot 2023-07-19 at 1 50 15 pm

jessie-ross: We could move to 18, maybe as a separate PR though.

jessie-ross: We should move to 18, the maintenance window is closing on 16 soonish

jessie-ross: I'm trying it in this PR

aaronyan (approved):
LGTM thank you!

PR #896, opened 2024-01-05
Fix experiment draft min diff precision

aaronyan:
Address: https://github.com/Automattic/experimentation-platform/issues/971

This fixes the precision issue for min diff values during when copying the experiment design in `/experiments/{experiment_id}/setup`.

## How has this been tested?

Go to an experiment setup page and click `Copy experiment design`. Paste the contents into p2 draft. The new decimals should be fixed for conversion metrics.

I couldn't get this to work when testing locally, but I checked the values are being properly scaled with `console.log()`.


src/components/explat/experiments/single-view/ExperimentSetup.tsx
```diff
@@ -43,20 +43,34 @@ export default function ExperimentSetup({
     experiment.name,
   )}`
 
+  function scaleAndRound(num: number, scale: number) {
```
aaronyan, line 46: This addresses floating-point imprecision. I was seeing `0.002` multiplied by `100` get turned into `0.1999999999`.

samacart (approved):
LGTM! 

The only thought I had was the use of `Math.log10` assumes that `scale` is always a power of 10. I can see your code has hard coded this to `100` so it should be fine. If it wasn't a power of 10 it might lead to some unexpected results.

PR #898, opened 2024-02-09
Fix render error for results when no startDateTime is not present

necronet:
### What's fix ?

There is a rendering issue happening for when a experiment in staging gets disabled. Since there is no `startDateTime` assigned to it, the result page does not get render, instead we see an error like this:

image

These changes aimed to ignore this restriction that is imposed as an error on `Experiments.getExperimentRunHours`, since it throws an error when no `startDateTime` is present and it make sense to do so, so instead the `getExperimentHealthIndicators` is handling the error and setting the hours difference for the experiment as zero.

You can see an example of this render issue [here](https://experiments.a8c.com/experiments/21505-explat-test-trying-to-break-clone/results). 

### Test that it works ? 

These changes include a test case for nullable `startDateTime`, which you can check by running:

`npm run jest -- --silent src/lib/explat/analyses.test.ts`

Additionally to test the same example above you can follow this steps:

- Branch out these changes locally
- Run with production flag `REACT_APP_PRODUCTION_CONFIG_IN_DEVELOPMENT=true npm run dev`
- Look for the local abacus mine is called `a8c-abacus-local` and search for the experiment results `21505-explat-test-trying-to-break-clone` for example:
`http://a8c-abacus-local:3000/experiments/21505-explat-test-trying-to-break-clone/results`

### Showcase

Screenshot 2024-02-09 at 16 09 10

Related issue: https://github.com/Automattic/experimentation-platform/issues/1014


src/lib/explat/analyses.ts
```diff
@@ -478,10 +478,17 @@ export const whitelistedPlatforms = {
  * Get experiment health indicators for a experiment.
  */
 export function getExperimentHealthIndicators(experiment: ExperimentFull): HealthIndicator[] {
+  let experimentRunHours = 0
+  try {
+    experimentRunHours = Experiments.getExperimentRunHours(experiment) / 24
+  } catch (error) {
+    console.error('Error getting experiment run hours', error)
```
aaronyan, line 485: I don't think we should be logging the errors in the console. When an experiment is `disabled` and has no `startDatetime`, `Experiments.getExperimentRunHours(experiment)` will always fail to return a value.

aaronyan: What do you think about adjusting the [condition](https://github.com/Automattic/abacus/blob/ba644813467708eb2e571e7602ca074c35895ef2/src/lib/explat/experiments.ts#L79C7-L79C32) `!experiment.startDatetime` in the method `getExperimentRunHour` instead?

aaronyan (commented):
Thanks for investigating and proposing a fix, @necronet. Shared a few thoughts about where we might want to consider making these checks.

src/lib/explat/analyses.ts
necronet: Yeah I consider the option of adding to the `getExperimentRunHour` method but this is also getting called on [MetricAssigmentResults](https://github.com/Automattic/abacus/blob/master/src/components/explat/experiments/single-view/results/MetricAssignmentResults.tsx#L550), and [recommendation.tsx](https://github.com/Automattic/abacus/blob/master/src/lib/explat/recommendations.ts#L186).

- Would we want to hide the `startDateTime` from all the use cases or only this particular result page for Staging > Disable experiments ? 

- I added `console.error` but I couldn't find any log mechanism that would signal something unexpected happen.

jessie-ross: Returning 0 from `getExperimentRunHour(experiment)` for these types of experiments makes a lot of sense to me - the run time is 0 afterall. It is fine for it to be called on those other pages.

We should also probably not be getting to this code from the Results page, I think we should add a new function `getExperimentStatus(Experiment)` that maps experiments to a more comprehensive status:
```
enum ExperimentStatus  {
    Staging,
    Cancelled,
    Running,
    Completed,
    Finished,
}
```

Where an experiment is Cancelled if `experiment.status == disabled && !experiment.startDatetime`.

We currently don't show results if it is Staging, but we should also not show results if it is cancelled, we could add a `experiments.hasResults(experiment)` function.

necronet: Excellent let me address these changes! I will:

- `getExperimentRunHour` to return 0 for `startDateTime` empty.
- Add the `getExperimentStatus(Experiment)` and ensure that a disabled experiment with no `startDateTime` does not call the `getExperimentRunHour` method gets called
- Modify and add additional test for those cases!

src/components/explat/experiments/single-view/results/ExperimentResults.test.tsx
```diff
@@ -617,3 +617,15 @@ test('renders correctly a one time experiment', async () => {
   expect(screen.queryByRole('row', { name: /over one year/ })).toBeNull()
   expect(screen.queryAllByRole('row', { name: /on the entire targeted audience/ }).length).toBeGreaterThanOrEqual(1)
 })
+
+test('renders correctly with an disabled and start date empty experiment', async () => {
```
aaronyan, line 621:
```suggestion
test('renders correctly with a disabled and empty start date experiment', async () => {
```

aaronyan (approved):
LGTM!

src/lib/explat/experiments.ts
```diff
@@ -77,7 +77,7 @@ export function getExperimentRunHours(experiment: ExperimentFull): number {
 
   // istanbul ignore next; shouldn't occur
```
jessie-ross, line 78: We can remove this line now - this turns off code coverage, but now we're covering this case.

src/components/explat/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -394,10 +394,14 @@ export default function ExperimentResults({
     experiment,
     primaryMetricLatestAnalysesByStrategy,
   )
-  const experimentHealthIndicators = [
-    ...Analyses.getExperimentParticipantHealthIndicators(experiment, experimentParticipantStats),
-    ...Analyses.getExperimentHealthIndicators(experiment),
-  ]
+
+  const experimentHealthIndicators =
+    experiment.status === Status.Disabled && !experiment.startDatetime
```
jessie-ross, line 399:
```suggestion
    experiments.shouldHaveResults(experiment)
```
This needs an explanation, I think the best way is to add a function, an alternative is to add a comment.

If we use the function we should explain that missing a start datetime means the experiment went straight from staging to disabled.

jessie-ross (approved):
Looks good, but see required changes in comments, mainly around documenting the missing start_datetime.

PR #899, opened 2024-02-15
Add duration to results page

samacart:
Related Issue: https://github.com/Automattic/experimentation-platform/issues/1009

## Overview

When an experiment's only health issue is a short run time, we still flag the experiment as having "Potential Issues" in the health summary box on the results page. This could give a false alarm to the user and is better managed in other ways.

An example of the current screen is: 

image

## Solution

To reduce alarming users, but still ensure they are aware of the shorter than preferred run time, the following changes have been made in this PR.

1. Add a new "Duration" box to the summary of the results page. 

This will show the duration of the experiment using the units defined in the experiment. Similar to the current health summary box, the duration box will have a colored bar across the top if the duration is outside of the preferred bounds. These bounds are using the existing code, with one change: the removal of the bound up to 3 days. 

2. Removal of the short duration from the health report summary box calculations. 

To remove the false alarm for experiments with only a short run time, the health indicator for durations of `HealthIndicationCode.Low` has been removed from the calculation. If the only health indicator is a longer than desired duration, this will still be shown as a "Potential Issue".

## Questions

1. Number of decimal places on duration summary

A design consideration. I have added the new duration summary box at a split of 33% for duration and the remainder for health summary. I need to limit this to 3 decimal places to ensure it only takes 2 lines. However, this differs from the health report at the bottom showing 4 decimal places. I also tried changing the split, but anything further will move the health summary to more than 2 lines. 

**Question** - Do we want to show decimals on the duration summary? If so, do we want the number of decimal places to match? 

### Use Case 1: Experiment with no health issues

**Expected Outcome:** Results page displays with no health issues in summary and duration box with no colored bar.
**Actual Outcome:** 🟢  Displays correctly with no warnings

CleanShot 2024-02-15 at 16 29 41@2x

### Use Case 2: Experiment with only health issue being low duration ( < 7 days )

**Expected Outcome:** Results page displays with no health issues in summary and duration box with orange colored bar.
**Actual Outcome:** 🟢  Results page displays correctly. No health issues are shown and the duration is shown with orange bar.

CleanShot 2024-02-15 at 16 31 27@2x

### Use Case 3: Experiment with only health issue being HIGH duration ( > 28 days )

**Expected Outcome:** Results page displays with potential health issues and duration box with orange colored bar.
**Actual Outcome:** 🟢  Results page displays correctly. "Potential Issues" displays for health summary and the duration is shown with orange bar.

CleanShot 2024-02-15 at 16 28 48@2x

### Use Case 4: Experiment with no duration health issues but other health issues

**Expected Outcome:** Results page displays with potential health issues and duration box with no colored bar.
**Actual Outcome:** 🟢  Results page displays correctly. "Potential Issues" displays for health summary and the duration is shown with no colored bar.

CleanShot 2024-02-15 at 17 28 50@2x

## How has this been tested?


src/components/explat/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -805,24 +825,45 @@ export default function ExperimentResults({
...
+                    // @ts-ignore: Component extensions aren't appearing in types.
+                    href='#health-report'
+                  >
+                    <div className={classes.summaryStats}>
+                      <Typography variant='h3' className={clsx(classes.summaryStatsStat)} color='primary'>
+                        Duration
+                      </Typography>
+                      <Typography variant='subtitle1'>
+                        {experimentDurationHealthIndicators[0].value.toFixed(3)}{' '}
+                        {experimentDurationHealthIndicators[0].unit}
```
jessie-ross, line 844: Maybe let's not use the indicator system just to pick out the date, rather use `math.Round(Experiments.getExperimentRunHours(experiment) / 24)` directly.

It's not really made for piecemeal usage, and would make it complicated to refactor later - it is already a complex thing so it would be nice to keep it contained to its intended use-case.

src/components/explat/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -394,18 +406,26 @@ export default function ExperimentResults({
     experiment,
     primaryMetricLatestAnalysesByStrategy,
   )
+  const experimentDurationHealthIndicators = Analyses.getExperimentHealthIndicators(experiment)
```
jessie-ross, line 409: `experimentDurationHealthIndicators != experimentHealthIndicators`
 
There is currently only one item here but there could be more.

src/lib/explat/analyses.ts
```diff
@@ -485,14 +485,6 @@ export function getExperimentHealthIndicators(experiment: ExperimentFull): Healt
       unit: HealthIndicatorUnit.Days,
       link: 'https://wp.me/PCYsg-Fqh/#experiment-run-time',
       indicationBrackets: [
```
jessie-ross, line 487: If we do want to reuse this, we could reuse the data instead of the result maybe, and extract this object out into a variable `experimentRunTimeIndicatorDefinition`.

Then we can get the indication by `getIndicationFromBrackets(experimentRunTimeIndicatorDefinition.indicationBrackets, runTimeDays)`

src/components/explat/experiments/single-view/results/ExperimentResults.tsx
```diff
@@ -394,18 +406,26 @@ export default function ExperimentResults({
...
+    )
     .map(({ indication: { severity } }) => severity)
     .sort(
       (severityA, severityB) =>
         Analyses.healthIndicationSeverityOrder.indexOf(severityB) -
         Analyses.healthIndicationSeverityOrder.indexOf(severityA),
     )[0]
+  const durationHealthIndicationSeverity = experimentDurationHealthIndicators.map(
+    ({ indication: { severity } }) => severity,
+  )[0]
```
jessie-ross, line 428: We shouldn't be assuming the order with `[0]` being experiment duration run time, we can instead use `experimentDurationHealthIndicators.find(x => x.name === "Experiment run time").indication.severity`

jessie-ross (changes requested):
See comment about adding Advisory level to the health stats - I don't really like breaking apart the already complex health indicator system.

But if we still want duration in the top level stats these are my thoughts on the UI:

- Can we fit the duration into the existing stats box, under partipants:
   304960220-493444e9-e36b-42f3-bc8d-5f1839365d34
- Let's round days.
- Maybe we can have the stat second line under the days saying "too short", "good length", "too long".

jessie-ross:
@samacart Easy mistake coming from private side of things but since Abacus is a public repo we can't include sensitive data here, I've gone a head and removed the images from your PR description, and then deleted the revision history (GitHub keeps past changes so those need to be deleted too.)

src/lib/explat/analyses.test.ts
```diff
@@ -536,8 +536,8 @@ describe('getExperimentHealthIndicators', () => {
       Array [
         Object {
           "indication": Object {
-            "code": "very low",
-            "reason": "−∞ < x ≤ 3",
+            "code": "low",
+            "reason": "−∞ < x ≤ 7",
```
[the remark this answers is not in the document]
samacart: Thanks - I did run --updateSnapshot but you are right, I think I need to use a few more `--` 😄

